PR #2, opened 2020-09-29
[framework migration] add terraform for ec2 

wyTrivail:
This is an initial pr to migrate the current testing framework to terraform. 

i have done one test case on ec2/amazon_linux, the plan is to

1. expand to multiple ec2 amis.
2. ecs
3. eks. 

this branch will eventually replace the current master branch.


terraform/ec2/amis.tf
```diff
@@ -0,0 +1,48 @@
...
+    name   = "owner-alias"
+    values = ["amazon"]
+  }
+
+  filter {
+    name   = "name"
+    values = ["amzn2-ami-hvm*"]
+  }
+
+  owners = ["amazon"] # Canonical
```
anuraaga, line 30:
```suggestion
  owners = ["amazon"]
```

terraform/ec2/amis.tf
```diff
@@ -0,0 +1,48 @@
...
+    name   = "owner-alias"
+    values = ["amazon"]
+  }
+
+  filter {
+    name   = "name"
+    values = ["suse-sles-15*"]
+  }
+
+  owners = ["amazon"] # Canonical
```
anuraaga, line 46:
```suggestion
  owners = ["amazon"]
```

terraform/ec2/ec2.tf
```diff
@@ -0,0 +1,97 @@
+provider "aws" {
+  region = var.region
+}
+
+data "aws_s3_bucket_object" "ssh_private_key" {
```
anuraaga, line 5: Is it possible to generate private key with terraform instead of maintaining a bucket?

https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key

It would be ephemereal and deleted when terraform destroy happens automatically

terraform/setup/setup.tf
```diff
@@ -0,0 +1,150 @@
+# in this module, we create the necessary common resources for the integ-tests, this setup module will only need to be executed once.
+# vpc, iam role, security group, the number of those resources could be limited, creating them concurrently for every pr would trigger throttling issue.
```
anuraaga, line 2: +1

terraform/setup/setup.tf
```diff
@@ -0,0 +1,150 @@
...
+
+provider "aws" {
+  region = var.region
+}
+
+# create ssh key pair and upload them to s3
+resource "tls_private_key" "ssh_key" {
```
anuraaga, line 9: Ah I see you generate key here - do you need to store it in a bucket? I think it's ok to generate key ever PR run, unless you know of a throttling limitation of `aws_key_pair`. Less permanent keys is safer :)

terraform/setup/setup.tf
anuraaga, line 2: Since you use default VPC you can clarify that here, I assumed from this comment you were creating a VPC here

terraform/ec2/ec2.tf
```diff
@@ -0,0 +1,97 @@
+provider "aws" {
```
anuraaga, line 1: Looks like we'll save a lot by deleting this folder! https://github.com/aws-observability/aws-otel-collector-test-framework/tree/master/integ-test/src/main/java/com/amazon/aocagent/services

terraform/setup/setup.tf
anuraaga: Though maybe you want to consider just creating a VPC here, so it's easier to manage e.g., private subnets

https://github.com/aws-samples/aws-xray-sdk-with-opentelemetry-sample/blob/both-otel-and-xray/scripts/terraform/network.tf#L75

anuraaga (commented):
Looked through terraform files first, I'll need to take some more time to look at the Java code tomorrow.

terraform/ec2/amis.tf
wyTrivail: thx, will do

terraform/ec2/ec2.tf
wyTrivail: i'm trying to keep the private key in case we need to log into the ec2 instance for debugging. we usually need to log into the instance to debug the collector and the image as the log of integ-test can't tell too much info to us.

terraform/setup/setup.tf
wyTrivail: yeah, just keep the key in s3 for debug purpose. And there might be a limit on it, but not like vpc (5 vpcs in one account), it could be hard to reach.

terraform/setup/setup.tf
wyTrivail: yes, i will probably create a vpc for integ-test when i do ecs and fargate if the default one is not easy to use.

terraform/ec2/ec2.tf
wyTrivail: yeah, it does!! i will provide ecs and eks soon

wyTrivail:
as per discussion today, i've change the validator from fetching s3 to directly calling the web application. right now we validate trace id and span id, will validate the payload in the later pr.

wyTrivail:
merged it for now, we can review it while merging this branch to master

PR #12, opened 2020-10-07
add a go application using x-ray sdk

JohnWu20:
Description:
Add a simple go application to the sample-apps. 

Test: 
I run the app and AOC in my local and get the trace data through the x-ray collector.


sample-apps/gp-xray-sdk-emitter/sample.go
```diff
@@ -0,0 +1,42 @@
+// Copyright The OpenTelemetry Authors
```
wyTrivail, line 1: please follow the requirement of building the sample app.  

1. *Web application/Docker image:  *The data emitter needs to be a web application serving a http endpoint, and needs to be able to built as a Docker image and run for testing as a Docker app. A Dockerfile needs to be provided.
2. *Integrate with SDK Repo workflow: *For each new code commit you make in the SDK, you will need to build a new Docker image. Every new Docker image should have a unique tag (you can use the code commit as the tag). Also maintain a “latest” tag for this image which builds with the released SDK. This will help theAOC (https://github.com/aws-observability/aws-otel-collector) repo to use the latest tag image to test with the latest version of the SDK. The SDK workflow sends a dispatch event with the built image link.
3. *Response pattern:* The response of each URL path need to be followed using the data structure listed below i.e. you just need to return the traceid. [json/application]

{
    "traceId": "xxxx"
}

4. *Unique ID*. [skip it if you just do trace not metric]Every time the validator, which is a component in the testing framework, calls the URL, the validator will generate a test-suite-id, which is unique for each run. The URL parameter, for example: http://x.x.x.x:8080/get-trace?testing-id=12313123241, can be used as an unique identifier for the metric e.g. adding this id as part of the metric name so that the metric could be different for each testing run. Note that no matter which http method you use, it will not be in the body, but just in the url parameter.
5. *Environment Variable: There will be some env vars which will be set while running this web application. The env vars will be but not limited to:*
    1. OTEL_EXPORTER_OTLP_ENDPOINT
    2. OTEL_RESOURCE_ATTRIBUTES
    3. *LISTEN_ADDRESS : *This is a mandatory environment variable. This web application address, for example (0.0.0.0:8080), makes the address controllable by the upper level; prevents port conflict with any other software on the testbed and also helps us limit the web application access within a private subnet.
6. *Support multiple URL path: *There’s no limitation on the URL path. There could be multiple URLs in this web application depending on the number of test cases you want.
7. *Keep “/” accessible with response code 200.:* This “/” will be used for the load balancer health check.


this code doesn't meet 1, 2, 3, 5

wyTrivail (changes requested):

sample-apps/gp-xray-sdk-emitter/sample.go
wyTrivail: 1.  you need to add a Dockerfile
2. reach out xray go sdk to add image building in sdk.  bhautip@
3. return the trace id in the response.
4. leverage OTEL_EXPORTER_OTLP_ENDPOINT as the aoc endpoint
5. leverage LISTEN_ADDRESS as the listen address of your endpoint. 


6. create pr with xrayreceiver.  check the workflow
7.  adding testing suite, follow https://github.com/aws-observability/aws-otel-collector-test-framework/tree/terraform/terraform

PR #32, opened 2020-10-16
TraceValidator

### LINKED ISSUE #36: [OTel-Test-Framework] Add Trace validator to validate trace data
Trace validator to validate trace data. Currently, TraceValidator supports testing aws sdk, http instrumentation for OTel and XRay SDKs.

bhautikpip:
1. Added trace validation logic
2. Added http and aws pre stored json file
3. tested with spring boot sample app


validator/src/main/java/com/amazon/aoc/callers/HttpCaller.java
```diff
@@ -29,7 +30,7 @@ public SampleAppResponse callSampleApp() throws Exception {
 
     AtomicReference<SampleAppResponse> sampleAppResponseAtomicReference = new AtomicReference<>();
     RetryHelper.retry(
-        60,
+        3,
```
wyTrivail: we will still need a bigger retry count, service launching time in different platform varies. 3 times might not be enough in some cases.

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
```diff
@@ -19,97 +19,99 @@
...
+            }
+        }
+    }
+
+    // this method will hit get trace from x-ray service and get retrieved trace
+    private Map<String, Object> getRetrievedTrace(List<String> traceIdList) throws Exception {
+        Map<String, Object> flattenedJsonMapForRetrievedTrace = null;
+
+        // retrieve trace from x-ray service
+        TimeUnit.SECONDS.sleep(15);
```
wyTrivail: sleeping 15 seconds might not be enough in some case, instead, retry it until success could make this validation more stable.

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
```diff
@@ -19,97 +19,99 @@
...
+        List<Trace> retrieveTraceList = xrayService.listTraceByIds(traceIdList);
+
+        try {
+            // flattened JSON object to a map
+            flattenedJsonMapForRetrievedTrace = JsonFlattener.flattenAsMap(retrieveTraceList.get(0).getSegments().get(0).getDocument());
+        } catch (Exception e) {
+            e.printStackTrace();
```
wyTrivail: log?

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
```diff
@@ -19,97 +19,99 @@
...
+    private Map<String, Object> getRetrievedTrace(List<String> traceIdList) throws Exception {
+        Map<String, Object> flattenedJsonMapForRetrievedTrace = null;
+
+        // retrieve trace from x-ray service
+        TimeUnit.SECONDS.sleep(15);
+        List<Trace> retrieveTraceList = xrayService.listTraceByIds(traceIdList);
+
+        try {
+            // flattened JSON object to a map
+            flattenedJsonMapForRetrievedTrace = JsonFlattener.flattenAsMap(retrieveTraceList.get(0).getSegments().get(0).getDocument());
```
wyTrivail: it will be better if we can do null point check here

validator/src/main/resources/expected-data-template/expectedAWSSDKTrace.json
```diff
@@ -0,0 +1,59 @@
...
+      "throttle": false,
+      "subsegments": [
+        {
+          "name": "S3",
+          "fault": false,
+          "error": false,
+          "throttle": false,
+          "http": {
+            "request": {
+              "url": "https://s3.us-west-2.amazonaws.com/",
```
wyTrivail: if we want to test in other region?

validator/src/main/resources/expected-data-template/expectedAWSSDKTrace.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "/aws-sdk-call",
+  "fault": false,
+  "error": false,
+  "throttle": false,
+  "http": {
+    "request": {
+      "url": "http://127.0.0.1:8080/aws-sdk-call",
```
wyTrivail: why the ip is 127.0.0.1? i imagine the ip should be the ip of the service?

validator/src/main/resources/expected-data-template/expectedHTTPTrace.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "/outgoing-http-call",
+  "fault": false,
+  "error": false,
+  "throttle": false,
+  "http": {
+    "request": {
+      "url": "http://127.0.0.1:8080/outgoing-http-call",
```
wyTrivail: same ask on ip :)

validator/build.gradle
```diff
@@ -41,6 +41,12 @@ dependencies {
     // yaml reader
     compile group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-yaml', version: '2.11.1'
 
+    // json reader
+    compile group: 'com.googlecode.json-simple', name: 'json-simple', version: '1.1'
```
wyTrivail: what's the reason to use json-simple instead of jackson? 
jackson is also easy to create map from json. https://mkyong.com/java/how-to-convert-java-map-to-from-json-jackson/ :)

and it's built-in in java, no need to add a dependency.

wyTrivail (changes requested):
just some small comments, thanks!

validator/build.gradle
bhautikpip: Basically challenge was to convert nested structure like segment documents to a flattened map (since for validation we want to access each field) but not sure we can do that using Jackson. I have tried couple of examples but unfortunately no luck.

validator/src/main/java/com/amazon/aoc/callers/HttpCaller.java
bhautikpip: Yes. I agree. Maybe we can probably keep this to 10? I feel 60 times retry is too much especially we will be using this module on Github actions where we don't have any control to terminate it we just have to wait for 60 retries to complete.

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
bhautikpip: Yes initially I was trying retry only but with retry it doesn't block consecutive operation for example, it tried for the first time and `retrieveTraceList` is empty so it tries again but when it retries second time it actually executes program with whatever it got on first retry so I want something blocking in order to use retries. Basically, can we block retry handler until `retrieveTraceList` is populated and once we have the list then only it performs further execution.

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
bhautikpip: good suggestion!

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
bhautikpip: sure.

validator/src/main/resources/expected-data-template/expectedAWSSDKTrace.json
bhautikpip: good point. maybe we can provide env variable to set this value? because url will be depending on which platform we deploy application right? so we won't be having url value on hand.

validator/src/main/resources/expected-data-template/expectedAWSSDKTrace.json
bhautikpip: maybe this also we can set using env variable?

wyTrivail: env var will work, like https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/validator/src/main/java/com/amazon/aoc/validators/MetricValidator.java#L146, we can use mustachehelper to render the context object into it.

wyTrivail: you add some parameter into the context object

validator/src/main/java/com/amazon/aoc/callers/HttpCaller.java
wyTrivail: i'm not sure if 10 would work, because it normally takes 10+ times to get it work for ecs test. probably 30?

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
wyTrivail: sorry i don't quite follow, can you retry on "listTraceByIds" until its size becomes none-zero?

validator/src/main/java/com/amazon/aoc/callers/HttpCaller.java
bhautikpip: Okay we can do 30 then.

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
bhautikpip: yes so the issue is retry helper retries when there is an exception but in the case of `listTraceByIds` we want to keep retrying until list becomes non empty. Here we have to wait for couple seconds because sample app sends trace to x-ray service and then we can retrieve it.

validator/src/main/resources/expected-data-template/expectedAWSSDKTrace.json
bhautikpip: okay will give it a shot.

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
wyTrivail: i see, can you throw a baseexception when the list is empty so that it will retry

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
```diff
@@ -78,35 +80,56 @@ public void validate() throws Exception {
     // this method will hit get trace from x-ray service and get retrieved trace
     private Map<String, Object> getRetrievedTrace(List<String> traceIdList) throws Exception {
         Map<String, Object> flattenedJsonMapForRetrievedTrace = null;
+        List<Trace> retrieveTraceList = null;
+        int MAX_RETRY_COUNT = 3;
+        int count = 1;
+
+        while (retrieveTraceList == null || retrieveTraceList.isEmpty()) {
+            // wait for sample app to send trace data to x-ray
```
wyTrivail: you can throw a new baseexception("the trace id list is empty") then retry handler can capture it

validator/src/main/resources/expected-data-template/expectedAWSSDKTrace.mustache
```diff
@@ -5,7 +5,7 @@
   "throttle": false,
   "http": {
     "request": {
-      "url": "http://127.0.0.1:8080/aws-sdk-call",
+      "url": "{{endpoint}}/aws-sdk-call",
```
wyTrivail: cool

validator/src/main/java/com/amazon/aoc/helpers/RetryHelper.java
```diff
@@ -37,11 +37,16 @@ public static void retry(
       throws Exception {
     while (retryCount-- > 0) {
       try {
-        log.info("still can retry for {} times", retryCount);
+        log.info("retry attempt left : {} ", retryCount);
         retryable.execute();
         return;
       } catch (Exception ex) {
-        log.error("exception during retry, you may ignore it", ex);
+        log.info("retrying after 10 seconds");
```
wyTrivail: i remember the interval is a parameter, not a fixed 10 seconds

validator/src/main/java/com/amazon/aoc/helpers/RetryHelper.java
```diff
@@ -37,11 +37,16 @@ public static void retry(
...
+
+        if (retryCount == 0) {
```
wyTrivail, line 46: can we move this out of the loop? since when the retries exhausted, it exit the loop

validator/src/main/resources/expected-data-template/expectedHTTPTrace.mustache
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "/outgoing-http-call",
+  "fault": false,
+  "error": false,
+  "throttle": false,
+  "http": {
+    "request": {
+      "url": "{{endpoint}}/outgoing-http-call",
```
wyTrivail: can we use "{{httpPath}}" here?

validator/src/main/resources/expected-data-template/expectedAWSSDKTrace.mustache
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "/aws-sdk-call",
+  "fault": false,
+  "error": false,
+  "throttle": false,
+  "http": {
+    "request": {
+      "url": "{{endpoint}}/aws-sdk-call",
```
wyTrivail: can we use {{httpPath}} here?

wyTrivail (approved):

validator/src/main/java/com/amazon/aoc/helpers/RetryHelper.java
bhautikpip: the reason I kept that inside is because we can print the actual exception after retry is exhausted (why retry failed?).

validator/src/main/java/com/amazon/aoc/helpers/RetryHelper.java
bhautikpip: sure will add param here.

PR #37, opened 2020-10-22
support multi span check in validator

wyTrivail:


validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
```diff
@@ -25,108 +25,135 @@
...
-
-        RetryHelper.retry(MAX_RETRY_COUNT, () -> {
-            List<Trace> retrieveTraceList = null;
-            retrieveTraceList = xrayService.listTraceByIds(traceIdList);
-            retrieveTraceListAtomicReference.set(retrieveTraceList);
-
-            if (retrieveTraceList == null || retrieveTraceList.isEmpty()) {
-                throw new BaseException(ExceptionCode.EMPTY_LIST);
-            }
+  }
```
bhautikpip, line 82: Can we add a line here saying validator for this path is done successfully? Currently we don't see any kind of success log it would be good for debugging on Github actions.

bhautikpip:
Suggested very minor change. Looks good:)

bhautikpip (approved):

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
wyTrivail: good catch!! let me add it

PR #40, opened 2020-10-23
ecscontainermetrics: update otconfig, expected metrics, and Validator logic

hossain-rayhan:
This change updates-
1. OT Config for ECS Container Metrics Receiver
2. Expected metrics for ECS Container Metrics Receiver
3. Validation logic for empty dimensions

Note: Once the dimensions/label settings feature is available in the exporter, we will create another PR to update our otconfig and expected metrics file.

### DIFF
--- a/validator/src/main/resources/expected-data-template/ecsContainerExpectedMetric.mustache
+++ b/validator/src/main/resources/expected-data-template/ecsContainerExpectedMetric.mustache
@@ -1,22 +1,32 @@
 -
+  metricName: ecs.task.memory.reserved_{{testingId}}
+  namespace: {{metricNamespace}}
+  dimensions:
+-
   metricName: ecs.task.memory.utilized_{{testingId}}
   namespace: {{metricNamespace}}
   dimensions:
-    -
-      name: ecs.cluster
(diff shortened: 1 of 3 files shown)


wyTrivail:
thanks! could you fix the build?

validator/src/main/resources/expected-data-template/ecsContainerExpectedMetric.mustache
```diff
@@ -1,22 +1,32 @@
+-
+  metricName: ecs.task.memory.reserved_{{testingId}}
+  namespace: {{metricNamespace}}
+  dimensions:
```
wyTrivail, line 4: let's wait until you have the real dimensions, then we can merge

wyTrivail (changes requested):
looks good to me overall, let's wait until you have the real dimensions, then we can merge it, thanks!

validator/src/main/resources/expected-data-template/ecsContainerExpectedMetric.mustache
hossain-rayhan: We are planning release with empty dimensions. Will create a new PR once the label/dimensions settings feature is available.

PR #42, opened 2020-10-26
modified template to perform metric validation on sample apps standard

bhautikpip:
1. Modified metric template to work with sample apps standard
2. tested with spring boot app (trace + metric data generator)


validator/src/main/resources/expected-data-template/defaultExpectedMetric.mustache
```diff
@@ -7,46 +7,7 @@
       value: cloudwatch-otel
     -
       name: apiName
-      value: /span0
+      value: /outgoing-http-call
```
wyTrivail, line 10: is "OTelLib" there? looks like you removed it?

wyTrivail (approved):

PR #75, opened 2020-11-13
Fix the trace unstable validation issue

wyTrivail:


validator/src/main/resources/expected-data-template/spark/sparkAppExpectedAWSSDKTrace.mustache
```diff
@@ -16,35 +16,7 @@
     {
       "fault": false,
       "error": false,
-      "throttle": false,
-      "http": {
-        "request": {
```
bhautikpip, line 21: Why are we removing all of it?

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
```diff
@@ -88,61 +88,56 @@ public void validate() throws Exception {
...
         () -> {
           List<Trace> retrieveTraceList = null;
           retrieveTraceList = xrayService.listTraceByIds(traceIdList);
-          retrieveTraceListAtomicReference.set(retrieveTraceList);
-
           if (retrieveTraceList == null || retrieveTraceList.isEmpty()) {
             throw new BaseException(ExceptionCode.EMPTY_LIST);
           }
+
+          // in case the json format is wrong, retry it.
```
bhautikpip, line 102: What do you mean by json format is wrong?

wyTrivail: check this: https://github.com/aws-observability/aws-otel-collector/runs/1397013128?check_suite_focus=true, the "e2etest-ecs (otlp_trace, EC2)" one, looks like xray returned a empty/invalid json of segment. but i can't reproduce it on my local even with the same trace id, so i suspect xray backend somehow returned a invalid json

validator/src/main/resources/expected-data-template/spark/sparkAppExpectedAWSSDKTrace.mustache
wyTrivail: the final one is 

```
[{
  "name": "aws-otel-integ-test",
  "fault": false,
  "error": false,
  "throttle": false,
  "http": {
    "request": {
      "url": "{{endpoint}}/aws-sdk-call",
      "method": "GET"
    },
    "response": {
      "status": 200
    }
  },
  "subsegments": [
    {
      "fault": false,
      "error": false,
      "throttle": false
    }
  ]
}]
```

the http block under subsegments are unpredictable, some times a put request is appeared

bhautikpip (approved):

PR #96, opened 2020-11-17
Add test cases for Datadog exporter

mx-psi:
**Description:**

Add integration testing for metrics and traces for the Datadog exporter﻿

**Testing:**

Ran it locally with
```
cd terraform/mock; terraform init && terraform apply -var="testcase=../testcases/datadog_exporter_metric_mock"
cd terraform/mock; terraform init && terraform apply -var="testcase=../testcases/datadog_exporter_trace_mock"
```

checked that the changes made sense

### DIFF
--- a/terraform/testcases/datadog_exporter_trace_mock/otconfig.tpl
+++ b/terraform/testcases/datadog_exporter_trace_mock/otconfig.tpl
@@ -1,0 +1,27 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:${grpc_port}
+
+processors:
+  batch:
+    timeout: 10s
+
+exporters:
+  logging:
+    loglevel: debug
+  datadog:
+    api:
+      key: testapikey
+    metrics:
+      endpoint: "https://${mock_endpoint}"
+    traces:
+      endpoint: "https://${mock_endpoint}"
+
+service:
+  pipelines:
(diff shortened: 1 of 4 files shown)


terraform/testcases/datadog_exporter_trace_mock/otconfig.tpl
```diff
@@ -0,0 +1,27 @@
...
+  batch:
+    timeout: 10s
+
+exporters:
+  logging:
+    loglevel: debug
+  datadog:
+    api:
+      key: testapikey
+    metrics:
```
wyTrivail, line 17: could you remove the metric block here since the test case is for trace?

wyTrivail (changes requested):
just a small nit, thanks!

terraform/testcases/datadog_exporter_trace_mock/otconfig.tpl
mx-psi: I should have made a comment about this, my bad: this is necessary because we actually send some stats about traces to that endpoint when running the trace exporter; if we leave it unset we would be hitting Datadog's backend on each test

wyTrivail: Make sense then, thx for explaining.

wyTrivail (approved):

PR #101, opened 2020-11-18
ecs-metrics: update expected template with new dimension names

hossain-rayhan:
Updated some metric dimension names in the expected metric template.


mxiamxia (approved):
Thanks.

PR #102, opened 2020-11-19
Abstract AOC as a service

kohrapha:
This PR abstracts the AOC container pod as a service to allow for future use cases that requires AOC to run as a service instead of a sidecar. We expose 2 ports (TCP/gRPC and UDP) which run in separate services, because k8s does not support a service with mixed protocols currently ([ref](https://github.com/kubernetes/kubernetes/pull/64471)).

### DIFF
--- a/terraform/eks/main.tf
+++ b/terraform/eks/main.tf
@@ -109,7 +109,7 @@ locals {
   eks_pod_config = yamldecode(data.template_file.eksconfig.rendered)["sample_app"]
 }
 
-# deploy aoc and sample app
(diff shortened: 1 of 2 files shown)


wyTrivail (approved):
Thanks, could you run the xrayreceiver testcase to test the udp port before i merge it?

kohrapha:
@wyTrivail I ran the xrayreceiver testcase and verified it succeeds with this workload

wyTrivail (approved):

PR #103, opened 2020-11-19
Add test case for New Relic exporter

umaannamalai:
**Description:**
Add integration testing for traces for the New Relic exporter

**Testing:**
Verified output/ changes after running the tests locally using the following commands:
```
cd aws-otel-test-framework/terraform/mock
terraform init
terraform apply -var="testcase=../testcases/newrelic_exporter_trace_mock" 
terraform destroy
```


wyTrivail (approved):
thanks

PR #104, opened 2020-11-19
Add test case for Dynatrace metric exporter

dyladan:
Description:

Add integration testing for metrics for the Dynatrace exporter﻿

Testing:

Ran tests locally using `terraform apply -var="testcase=../testcases/dynatrace_exporter_metric_mock"`

/cc @arminru


wyTrivail (approved):
thanks!

PR #128, opened 2020-11-25
Add canary tests

shaochengwang:
**Description:**
Add canary tests

**Tests done:**
1. Tested each test case locally in `shaochengwang/aws-otel-test-framework` e.g.
```
cd terraform/canary && terraform init && terraform apply -auto-approve -lock=false -var-file="../testcases/otlp_trace/parameters.tfvars" -var="aoc_version=latest" -var="testcase=../testcases/otlp_trace" -var="testing_ami=canary_windows"
```
2. Tested with GitHub workflow in ` shaochengwang/aws-otel-collector`


terraform/canary/common.auto.tf
```diff
@@ -0,0 +1,48 @@
+variable "region" {
```
wyTrivail: this file is a symlink to the common.tf, you don't need to really add content for it. just create a symlink instead

terraform/canary/main.tf
```diff
@@ -0,0 +1,53 @@
...
+# launch ec2
+module "ec2_setup" {
+  source = "../ec2"
+
+  ami_family = var.ami_family
+  amis = var.amis
+  testing_ami = var.testing_ami
+  aoc_version = var.aoc_version
+  region = var.region
+  canary = var.canary
```
wyTrivail: i guess canary will always equal to true here?

terraform/ec2/variables.tf
```diff
@@ -98,3 +98,7 @@ variable "install_package_local_path" {
 variable "testing_type" {
   default = ""
 }
+
+variable "canary" {
```
wyTrivail, line 98: maybe give a comment here to explain the meaning of the "canary" var?

validator/src/main/java/com/amazon/aoc/App.java
```diff
@@ -86,14 +95,43 @@ public Integer call() throws Exception {
...
     List<ValidationConfig> validationConfigList =
         new ConfigLoadHelper().loadConfigFromFile(configPath);
 
     // run validation
+    int maxValidationCycles = 1;
     ValidatorFactory validatorFactory = new ValidatorFactory(context);
-    for (ValidationConfig validationConfigItem : validationConfigList) {
-      validatorFactory.launchValidator(validationConfigItem).validate();
+    if (this.isCanary) {
+      maxValidationCycles = 30;
```
wyTrivail, line 128: any possibility we can write this part of logic in a separate method?

validator/src/main/java/com/amazon/aoc/App.java
```diff
@@ -86,14 +95,43 @@ public Integer call() throws Exception {
...
+    }
+    for (int cycle = 0; cycle < maxValidationCycles; cycle++) {
+      for (ValidationConfig validationConfigItem : validationConfigList) {
+        try {
+          validatorFactory.launchValidator(validationConfigItem).validate();
+        } catch (Exception e) {
+          if (this.isCanary) {
+            //emit metric
+            cloudWatchService.putMetricData("Otel/Canary", "Success", 0.0);
```
wyTrivail: any dimension?

validator/src/main/java/com/amazon/aoc/App.java
wyTrivail, line 128: how long does it take to run a round of canary, there's a limit on github that a job can only be running for no more than 6 hours.

validator/src/main/java/com/amazon/aoc/validators/MetricValidator.java
```diff
@@ -90,8 +90,10 @@ public void validate() throws Exception {
           log.info("check if all the expected metrics are found");
           compareMetricLists(expectedMetricList, metricList);
 
-          log.info("check if there're unexpected additional metric getting fetched");
-          compareMetricLists(metricList, expectedMetricList);
+          if (!context.isCanary()) {
```
wyTrivail: i guess you don't need to skip this check now, we changed the metric template.

wyTrivail (changes requested):
thanks!

validator/src/main/java/com/amazon/aoc/App.java
shaochengwang: Previously, I included `testingId` as one dimension, but then I found we cannot set alarm if the each run generates a different metric. I think canary tag is included in the namespace. What other information do we need for dimension?

validator/src/main/java/com/amazon/aoc/App.java
shaochengwang: 30, 31 minutes

wyTrivail (approved):
lgtm

PR #129, opened 2020-11-25
Metrics Validator for Prometheus Pipeline

JasonXZLiu:

Related PRs:
* PR for sample app: [Link](https://github.com/aws-observability/aws-otel-test-framework/pull/131)
* PR for terraform/testcase configuration: [Link](https://github.com/aws-observability/aws-otel-test-framework/pull/130)

PR 2/3:

In this PR, we create a new validator for Prometheus metrics. As we want to verify dynamic metric (generated by the sample app), we pull expected metrics from the sample app. We then pull the actual metrics from the Cortex instance (provided a Cortex instance endpoint).


validator/build.gradle
```diff
@@ -55,6 +59,10 @@ dependencies {
     implementation 'com.amazonaws:aws-java-sdk-cloudwatch'
     implementation 'com.amazonaws:aws-java-sdk-xray'
 
+    // https://mvnrepository.com/artifact/com.github.awslabs/aws-request-signing-apache-interceptor
+    implementation 'com.github.awslabs:aws-request-signing-apache-interceptor:b3772780da'
+    implementation files('libs/AWSPrometheusJavaClient-1.11.x.jar')
```
wyTrivail: when will this lib to be available on maven?

validator/build.gradle
```diff
@@ -55,6 +59,10 @@ dependencies {
     implementation 'com.amazonaws:aws-java-sdk-cloudwatch'
     implementation 'com.amazonaws:aws-java-sdk-xray'
 
+    // https://mvnrepository.com/artifact/com.github.awslabs/aws-request-signing-apache-interceptor
+    implementation 'com.github.awslabs:aws-request-signing-apache-interceptor:b3772780da'
```
wyTrivail, line 63: what's "b3772780da"?

validator/src/main/java/com/amazon/aoc/clients/ExpectedResultClient.java
```diff
@@ -0,0 +1,51 @@
...
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+import java.io.IOException;
+import java.util.List;
+
+public class ExpectedResultClient {
```
wyTrivail: the name of "ExpectedResultClient" made me think it's a generic client for all the types of metrics including cw and prometheus but looks like it's just for prometheus?

validator/src/main/java/com/amazon/aoc/clients/ExpectedResultClient.java
```diff
@@ -0,0 +1,51 @@
...
+   * list expected metrics from the sample app.
+   */
+  public List<PrometheusMetric> getExpectedResults() throws IOException {
+    Request request = new Request.Builder().url(endpoint).build();
+    return execute(request);
+  }
+
+  private List<PrometheusMetric> execute(Request request) throws IOException {
+    OkHttpClient client = new OkHttpClient();
+    Response response = client.newCall(request).execute();
```
wyTrivail: do you wanna retry on it in case any network issue would make the endpoint unstable?

validator/src/main/java/com/amazon/aoc/clients/ExpectedResultClient.java
```diff
@@ -0,0 +1,51 @@
...
+    if (response.code() >= 300) {
+      throw new ExpectedResultClientException(response.body().string());
+    }
+
+    String body = response.body().string();
+
+    return new ObjectMapper().readValue(body, new TypeReference<List<PrometheusMetric>>() {});
+  }
+
+  static class ExpectedResultClientException extends RuntimeException {
```
wyTrivail: instead of defining a new exception from runtimeexception, do you want to define a new exception code in the baseexception?

validator/src/main/java/com/amazon/aoc/clients/PrometheusClient.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.http.impl.client.HttpClients;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.List;
+import java.util.Objects;
+
+@Log4j2
+public class PrometheusClient {
```
wyTrivail: i'm assuming this is to call the sample app? any possibility we can make it generic? name it as "PullModeSampleAppClient"?

validator/src/main/java/com/amazon/aoc/clients/PrometheusClient.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    String body = IOUtils.toString(response.getEntity().getContent());
+    PrometheusQueryResult result =
+            new ObjectMapper().readValue(body, PrometheusQueryResult.class);
+
+    return result.getData().getResult();
+  }
+
+  static class PrometheusClientException extends RuntimeException {
```
wyTrivail: same as above, try baseexception?

validator/src/main/java/com/amazon/aoc/models/prometheus/PrometheusMetric.java
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * Strict comparator for comparing two Prometheus metrics by the metric name, labels and values.
+   */
+  public static class StrictMetricComparator implements Comparator<PrometheusMetric> {
+    public StrictMetricComparator() {
+    }
+
+    @Override
+    public int compare(PrometheusMetric o1, PrometheusMetric o2) {
```
wyTrivail: it will be great if we could do it in a helper class, so that other metric validator could use it?

validator/src/main/java/com/amazon/aoc/validators/PrometheusMetricValidator.java
```diff
@@ -0,0 +1,164 @@
...
+  @Override
+  public void validate() throws Exception {
+    log.info("Start prometheus metric validating");
+    // get expected metrics
+    final List<PrometheusMetric> expectedMetricList = this.getExpectedMetricList(context);
+
+    // get metric from cortex
+    PrometheusService prometheusService = new PrometheusService(context);
+    RetryHelper.retry(
+        MAX_RETRY_COUNT,
```
wyTrivail, line 59: just retry once?

validator/src/main/java/com/amazon/aoc/validators/PrometheusMetricValidator.java
```diff
@@ -0,0 +1,164 @@
...
+                        metric, metricSet));
+      }
+    }
+  }
+
+  private List<PrometheusMetric> getExpectedMetricList(Context context) throws Exception {
+    if (!validationConfig.getExpectedResultPath().isEmpty()) {
+      // add delay to allow sample app lb to start
+      log.info("sleeping to allow sample app lb to start");
+      TimeUnit.SECONDS.sleep(90);
```
wyTrivail: using retry would be better than purely waiting.

validator/src/main/java/com/amazon/aoc/validators/ValidatorFactory.java
```diff
@@ -45,8 +45,13 @@ public IValidator launchValidator(ValidationConfig validationConfig) throws Exce
...
         expectedData = validationConfig.getExpectedTraceTemplate();
         break;
-      case "metric":
-        validator = new MetricValidator();
+      case "cw-metric":
+        validator = new CWMetricValidator();
+        expectedData = validationConfig.getExpectedMetricTemplate();
+        break;
+      case "prom-metric-scraping":
+      case "prom-metric-sd":
```
wyTrivail: what's the difference between "prom-metric-scraping" and "prom-metric-sd" from validator perspective?

wyTrivail (changes requested):
thanks! some small nits.

validator/build.gradle
JasonXZLiu: I forgot to remove this jar (we don't actually use it in the code and it's not actually in this PR).

validator/src/main/java/com/amazon/aoc/clients/PrometheusClient.java
JasonXZLiu: Oh, this one is to call our Cortex Instance. The ExpectedResultClient is actually the one that calls the sample app. I'll rename both clients to make it more clear.

validator/build.gradle
JasonXZLiu: It's the version of the library in maven: https://mvnrepository.com/artifact/com.github.awslabs/aws-request-signing-apache-interceptor?repo=mulesoft-public.

validator/src/main/java/com/amazon/aoc/validators/ValidatorFactory.java
JasonXZLiu: There is no difference between the `prom-metric-scraping` and `prom-metric-sd` testcases when instantiating the validators. It only instantiates different comparators later on in the actual validation. 

I'll add a field `shouldValidateMetricValue` that can be configured in the Validator Config instead.

validator/src/main/java/com/amazon/aoc/models/prometheus/PrometheusMetric.java
JasonXZLiu: I believe the other metric validator can use it as it is a a static class, but I can move it to it's own helper.

wyTrivail (approved):

PR #130, opened 2020-11-25
Add Testcases for Prometheus Pipeline

JasonXZLiu:

Related PRs:
* PR for sample app: [Link](https://github.com/aws-observability/aws-otel-test-framework/pull/131)
* PR for validator: [Link](https://github.com/aws-observability/aws-otel-test-framework/pull/129)

PR 3/3:

In this PR, we add the parameter and AOC configuration files to run the metric scraping and service discovery testcases.

### DIFF
--- a/terraform/basic_components/main.tf
+++ b/terraform/basic_components/main.tf
@@ -73,6 +73,10 @@ data "template_file" "otconfig" {
     testing_id = var.testing_id
     grpc_port = module.common.grpc_port
     udp_port = module.common.udp_port
+    cortex_instance_endpoint = module.common.cortex_instance_endpoint
+    sample_app_listen_address_ip = module.common.sample_app_listen_address_ip
+    sample_app_listen_address_port = module.common.sample_app_listen_address_port
+    sample_app_lb_port = module.common.sample_app_lb_port
 
     mock_endpoint = var.mocked_endpoint
   }
--- a/terraform/basic_components/main.tf
+++ b/terraform/basic_components/main.tf
@@ -73,7 +73,7 @@ data "template_file" "otconfig" {
     testing_id = var.testing_id
     grpc_port = module.common.grpc_port
     udp_port = module.common.udp_port
-    cortex_instance_endpoint = module.common.cortex_instance_endpoint
+    cortex_instance_endpoint = var.cortex_instance_endpoint
     sample_app_listen_address_ip = module.common.sample_app_listen_address_ip
     sample_app_listen_address_port = module.common.sample_app_listen_address_port
     sample_app_lb_port = module.common.sample_app_lb_port
--- a/terraform/basic_components/main.tf
+++ b/terraform/basic_components/main.tf
@@ -74,10 +74,7 @@ data "template_file" "otconfig" {
     grpc_port = module.common.grpc_port
     udp_port = module.common.udp_port
     cortex_instance_endpoint = var.cortex_instance_endpoint
-    sample_app_listen_address_ip = module.common.sample_app_listen_address_ip
-    sample_app_listen_address_port = module.common.sample_app_listen_address_port
-    sample_app_lb_port = module.common.sample_app_lb_port
-
+    
     mock_endpoint = var.mocked_endpoint
   }
 }
--- a/terraform/eks/main.tf
+++ b/terraform/eks/main.tf
@@ -109,8 +109,39 @@ locals {
   eks_pod_config = yamldecode(data.template_file.eksconfig.rendered)["sample_app"]
 }
 
+resource "kubernetes_service_account" "aoc-role" {
+  metadata {
+    name = "aoc-role"
+    namespace = kubernetes_namespace.aoc_ns.metadata[0].name
+  }
+
+  automount_service_account_token = true
+}
+
+resource "kubernetes_cluster_role_binding" "aoc-role-binding" {
+  metadata {
+    name = "aoc-role-binding"
+  }
+  role_ref {
+    api_group = "rbac.authorization.k8s.io"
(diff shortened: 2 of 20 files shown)


terraform/basic_components/main.tf
```diff
@@ -73,6 +73,10 @@ data "template_file" "otconfig" {
     testing_id = var.testing_id
     grpc_port = module.common.grpc_port
     udp_port = module.common.udp_port
+    cortex_instance_endpoint = module.common.cortex_instance_endpoint
```
wyTrivail: it will be a fixed value? if we put it in the output of common module, we can't use var to change its value while we use var to change the value in github workflow. can we configure it in variables?

terraform/eks/variables.tf
```diff
@@ -17,4 +17,7 @@ variable "eks_cluster_name" {
   default = "aoc-test-eks-ec2"
 }
 
+variable "sample_app_mode" {
```
wyTrivail: i think we can define this var in the common.tf so that all the platforms can leverage this flag even though they don't have the logic yet.

terraform/testcases/prometheus_sd/otconfig.tpl
```diff
@@ -0,0 +1,28 @@
+receivers:
```
wyTrivail, line 1: will we be able to use this config in ec2?

terraform/testcases/prometheus_static/otconfig.tpl
```diff
@@ -0,0 +1,23 @@
+receivers:
```
wyTrivail, line 1: will we be able to use it in ec2?

terraform/eks/main.tf
```diff
@@ -109,8 +109,39 @@ locals {
...
+  }
+}
+
+##########################################
+# Push mode deployments
+##########################################
+
 # deploy aoc and mocked server
 resource "kubernetes_deployment" "aoc_deployment" {
+  count = var.sample_app_mode == "pull" ? 0 : 1
```
wyTrivail: can we try to run one testcase which is using push mode in case this change break it?

wyTrivail: use == "push" would be more straightforward

terraform/eks/main.tf
```diff
@@ -151,6 +184,13 @@ resource "kubernetes_deployment" "aoc_deployment" {
           }
         }
 
+        volume {
+          name = kubernetes_service_account.aoc-role.default_secret_name
```
wyTrivail: do we need it in push mode?

terraform/eks/main.tf
```diff
@@ -196,13 +242,15 @@ resource "kubernetes_deployment" "aoc_deployment" {
 
 # create service upon the mocked server
 resource "kubernetes_service" "mocked_server_service" {
+  count = var.sample_app_mode == "pull" ? 0 : 1
```
wyTrivail: i would suggest we keep mock server no matter if it's a pull mode, so that we can also switch to use mock server to validate in the future.

terraform/eks/main.tf
```diff
@@ -344,13 +398,249 @@ resource "kubernetes_deployment" "sample_app_deployment" {
...
+
+    port {
+      port = 80
+      target_port = 8080
+    }
+  }
+}
+
+# deploy sample app
+resource "kubernetes_deployment" "pull_mode_sample_app_deployment" {
```
wyTrivail: since this file is going to be too long, can we put all the things about pull mode into a separate file named pull_mode_deployment.tf something?

wyTrivail (changes requested):
thanks! some small nits

terraform/testcases/prometheus_sd/otconfig.tpl
JasonXZLiu: No, this config is specific to kubernetes (as its doing service discovery on the Kubernetes objects).

terraform/testcases/prometheus_static/otconfig.tpl
JasonXZLiu: It should be able to without many changes. Right now it isn't, it only works for EKS, as we need to add some environment variables (`SAMPLE_APP_HOST` and `SAMPLE_APP_PORT`) to the EC2 instance so the Collector can point to the right host.

terraform/eks/main.tf
JasonXZLiu: Nope, we can take it out.

terraform/eks/main.tf
JasonXZLiu: I tried running the otlp_mock test case, and it looks like it's passing: 

```
module.validator.null_resource.validator (local-exec): validator_1  | 20:29:56.842 [main] INFO  com.amazon.aoc.validators.MockedServerValidator - mocked server validation passed
module.validator.null_resource.validator (local-exec): eks_validator_1 exited with code 0
module.validator.null_resource.validator (local-exec): Aborting on container exit...
module.validator.null_resource.validator: Creation complete after 1m37s [id=7503739792628168714]
```

terraform/testcases/prometheus_sd/otconfig.tpl
amanbrar1999: Do we need to get it working on EC2 in order to merge this PR? I have made the changes to get it working with EC2 and was going to submit the change in a PR for soaking/performance tests. I can move those changes to this PR if preffered

wyTrivail: no, we can tackle that in the next pr

wyTrivail (approved):
thanks

PR #131, opened 2020-11-25
Add prometheus sample app

amanbrar1999:

Related PRs:

PR for validator: [Link](https://github.com/aws-observability/aws-otel-test-framework/pull/129)
PR for terraform/testcase configuration: [Link](https://github.com/aws-observability/aws-otel-test-framework/pull/130)

PR 1/3:

In this PR, we introduce a new sample app that produces metrics in Prometheus exposition format at the `/metrics` endpoint, so that the Prometheus pipeline within AOC can scrape and process it.

We also have chosen to do dynamic metrics validation as opposed to the static validation done by the other validators, hence we have also included a `/expected_metrics` endpoint for the validator to query and use as the expected values.

Note that the go.sum file bloats the size of this PR.


wyTrivail (approved):
thanks!

PR #138, opened 2020-11-25
Add test case for SignalFx metric exporter

tigrannajaryan:
Add integration testing for metrics for SignalFx exporter


wyTrivail (approved):
thanks!

PR #139, opened 2020-11-25
Add test case for SignalFxCorrelation trace exporter

tigrannajaryan:
Add integration testing for traces for SignalFxCorrelation exporter


wyTrivail (approved):
thanks!

PR #143, opened 2020-11-26
Fix configuration of SignalFx metric exporter

tigrannajaryan:
The configuration was missing the required options. Presumably this is the reason why
the build of https://github.com/aws-observability/aws-otel-collector/pull/181 failed.


wyTrivail (approved):
thanks

PR #144, opened 2020-11-27
Make ec2 test compatible with prometheus

amanbrar1999:
Note that this change depends on the 2 previous PRs that introduce Prometheus logic into this testing framework: https://github.com/aws-observability/aws-otel-test-framework/pull/130, https://github.com/aws-observability/aws-otel-test-framework/pull/129

Since Prometheus receiving is a pull based system, there is a dependency that AOC has on the sample app. We have chosen to include this dependency via environment variable. This does not disrupt any other functionality.


terraform/ec2/main.tf
```diff
@@ -302,6 +303,7 @@ data "template_file" "cwagent_config" {
     data_rate = "${var.soaking_data_mode}-${var.soaking_data_rate}"
     instance_type = aws_instance.aoc.instance_type
     testing_ami = var.testing_ami
+    testing_type = var.testing_type
```
wyTrivail: you can remove this line now, we don't need it anymore

terraform/ec2/main.tf
```diff
@@ -222,6 +222,7 @@ resource "null_resource" "start_collector" {
   provisioner "remote-exec" {
     inline = [
       local.ami_family["install_command"],
+      local.selected_ami["family"] == "amazon_linux" ? "sudo chmod 777 /opt/aws/aws-otel-collector/etc/.env && sudo echo 'SAMPLE_APP_HOST=${aws_instance.sidecar.public_ip}' >> /opt/aws/aws-otel-collector/etc/.env && sudo echo 'SAMPLE_APP_PORT=${module.common.sample_app_lb_port}' >> /opt/aws/aws-otel-collector/etc/.env" : "echo \"not amazon linux\"",
```
wyTrivail: can you configure it as a parameters in the ami family somewhere? like [here](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/ec2_setup/amis.tf#L3), i'm not sure if you want to support windows because the soaking test will also cover windows

wyTrivail (changes requested):
just some small nits, thanks!

terraform/ec2/main.tf
amanbrar1999: I tried that before, but it does not allow the use of the variable that I require (aws_instance.sidecar.public_ip)

amanbrar1999: As for windows support I can add that in, I was under the assumption that only amazon linux is used for perf/soaking tests

terraform/ec2/amis.tf
```diff
@@ -44,6 +46,7 @@ variable "ami_family" {
       otconfig_destination = "C:\\ot-default.yml"
       download_command_pattern = "powershell -command \"Invoke-WebRequest -Uri %s -OutFile C:\\aws-otel-collector.msi\""
       install_command = "msiexec /i C:\\aws-otel-collector.msi"
+      set_env_var_command = "SET SAMPLE_APP_HOST=%s && SET SAMPLE_APP_PORT=%s"
```
wyTrivail: did you test soaking with windows?

wyTrivail (approved):

terraform/ec2/amis.tf
amanbrar1999: Will do so first thing tomorrow morning

terraform/ec2/variables.tf
```diff
@@ -100,6 +100,5 @@ variable "canary" {
 }
 
 variable "cortex_instance_endpoint" {
```
wyTrivail, line 102: would this parameter to be in common.tf?

PR #145, opened 2020-11-27
added Prometheus mock test

amanbrar1999:
This is the mock test that will be run on github workflows. 

This does not depend on any of the other Prometheus-related PRs. 

Note that we do not need to make a call to the Prometheus sample app in order for it to generate data hence why some changes were required in the validator.

### DIFF
--- a/validator/src/main/resources/validations/default-mocked-server-prometheus-validation.yml
+++ b/validator/src/main/resources/validations/default-mocked-server-prometheus-validation.yml
@@ -1,0 +1,4 @@
+-
+  validationType: "mocked-server"
+  # no call to prometheus sample app needed to produce metrics
+  callingType: "none" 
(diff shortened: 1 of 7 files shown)


validator/src/main/resources/validations/default-mocked-server-prometheus-validation.yml
```diff
@@ -0,0 +1,4 @@
+-
+  validationType: "mocked-server"
+  # no call to prometheus sample app needed to produce metrics
```
wyTrivail, line 3: i guess this comment should be "no call to prometheus sample app needed to get expected metrics" ?

wyTrivail (changes requested):
a small nit :)

wyTrivail (approved):

validator/src/main/resources/validations/default-mocked-server-prometheus-validation.yml
amanbrar1999: my understanding was that the caller is used to tell the sample app to produce metrics to send to AOC, rather than get expected metrics

PR #146, opened 2020-11-27
Add prometheus soaking and perf tests

amanbrar1999:
Note that the code in this PR functions correctly on its own, however it overlaps with the prometheus EC2 compatibility PR (https://github.com/aws-observability/aws-otel-test-framework/pull/144) and the prometheus mock PR (https://github.com/aws-observability/aws-otel-test-framework/pull/145), so should probably be merged after them.

With this PR, performance and soaking tests work for the prometheus mock test. I added logic to make the soaking data emitter a variable so that the Prometheus sample app can be used, and also made the soaking docker compose file a variable so that a custom one can be used by the Prometheus test cases, or any other future test cases that require it to be custom.

When the rest of the Prometheus code is merged, I will add the soaking docker compose file for prometheus_static if necessary.

The commands I used to successfully run these tests are as follows:

Soaking: `terraform init && terraform apply -var="testcase=../testcases/prometheus_mock" -var="install_package_source=local" -var-file="../testcases/prometheus_mock/parameters.tfvars"`

Performance: `terraform init && terraform apply -var="data_rate=100" -var="testcase=../testcases/prometheus_mock" -var="install_package_source=local" -var-file="../testcases/prometheus_mock/parameters.tfvars"`


terraform/ec2/main.tf
```diff
@@ -224,6 +224,7 @@ resource "null_resource" "start_collector" {
   provisioner "remote-exec" {
     inline = [
       local.ami_family["install_command"],
+      local.selected_ami["family"] == "amazon_linux" ? "sudo chmod 777 /opt/aws/aws-otel-collector/etc/.env && sudo echo 'SAMPLE_APP_HOST=${aws_instance.sidecar.public_ip}' >> /opt/aws/aws-otel-collector/etc/.env && sudo echo 'SAMPLE_APP_PORT=${module.common.sample_app_lb_port}' >> /opt/aws/aws-otel-collector/etc/.env" : "echo \"not amazon linux\"",
```
wyTrivail: same as the comment in the other pr :)

terraform/soaking/variables.tf
```diff
@@ -56,4 +56,8 @@ variable "soaking_data_rate" {
 
 variable "soaking_data_type" {
   default = "otlp"
-}
\ No newline at end of file
+}
+
+variable "data_emitter" {
```
wyTrivail: can you help to change the name of "data_emitter" to "sample_app_image"? in case any confusion.

terraform/templates/defaults/soaking_docker_compose_pull_mode.tpl
```diff
@@ -0,0 +1,23 @@
...
+      resources:
+        limits:
+          memory: 1G
+  ot-metric-emitter:
+    image: ${sample_app_image}
+    command: []
+    ports:
+      - ${sample_app_external_port}:${sample_app_listen_address_port}
+    environment:
+      METRICS_LOAD: ${rate}
```
wyTrivail, line 18: cool

terraform/ec2_setup/main.tf
```diff
@@ -27,11 +37,11 @@ module "ec2_setup" {
   aoc_version = var.aoc_version
   region = var.region
   testcase = var.testcase
-  sample_app_image = var.soaking_data_emitter_image
+  sample_app_image = var.data_emitter != "" ? "${data.aws_ecr_repository.sample_apps.repository_url}:${var.data_emitter}-latest" : var.soaking_data_emitter_image
```
wyTrivail: can you help to change the soaking_data_emitter_image to soaking_sample_app_image?

wyTrivail (changes requested):
some small nits, thx!

terraform/soaking/variables.tf
amanbrar1999: this was a variable I added that was meant to be what "sample_app" is to "sample_app_image" for data emitter images. The idea was for data_emitter to be from the imagebuild sample apps. Setting this to "sample_app_image" will overlap with a variable with a different meaning so I don't think that is the right thing to do. I think I can remove this variable and just use the variable "sample_app" because it has no other uses in ec2_setup, soaking, and performance

amanbrar1999: On second thought I realised sample_app is in the common.tf and has a default of "spark", so it cannot be reused here, I will instead name it "soaking_sample_app"

terraform/ec2_setup/main.tf
amanbrar1999: will do 👍

wyTrivail: maybe we can use sample_app instead of data_emitter? check if sample_app is set first and then use soaking_data_emitter_image if not?

wyTrivail (approved):

wyTrivail (approved):

PR #150, opened 2020-11-30
Add soaking and perf to prometheus static

amanbrar1999:
This PR is branched off of https://github.com/aws-observability/aws-otel-test-framework/pull/146 and heavily based on https://github.com/aws-observability/aws-otel-test-framework/pull/144. This should wait until both of those are merged. 

This simply adds configs needed to run performance and soaking tests using the prometheus static config, whereas previously only the mock case was supported

I have also moved `cortex_instance_endpoint` to be defined in the `common.tf` file since it is to be shared across multiple platforms, it can be specified there and can be overwritten if need be in individual `parameters.tfvars` files. 

I have also removed logging exporter from the prometheus static case as it was mainly for debugging, and may also effect the performance tests.


terraform/ec2/main.tf
```diff
@@ -224,6 +224,7 @@ resource "null_resource" "start_collector" {
   provisioner "remote-exec" {
     inline = [
       local.ami_family["install_command"],
+      local.selected_ami["family"] == "amazon_linux" ? "sudo chmod 777 /opt/aws/aws-otel-collector/etc/.env && sudo echo 'SAMPLE_APP_HOST=${aws_instance.sidecar.public_ip}' >> /opt/aws/aws-otel-collector/etc/.env && sudo echo 'SAMPLE_APP_PORT=${module.common.sample_app_lb_port}' >> /opt/aws/aws-otel-collector/etc/.env" : "echo \"not amazon linux\"",
```
wyTrivail: the same comment, could you move it into the ami family def

wyTrivail (changes requested):
this pr seems similar to https://github.com/aws-observability/aws-otel-test-framework/pull/146/files?

wyTrivail:
can this pr be closed?

amanbrar1999:
When the other PRs are merged, this will boil down to a very small change within the prometheus_static directory. The overlap is not 100% though so this needs to be left open

wyTrivail (approved):
thanks!

PR #152, opened 2020-12-01
made cluster role binding and service account unique, and added delay…

amanbrar1999:
… to pull mode aoc deployment

Hopefully this fixes the issue with the other tests. 

I also added in a delay for the pull mode AOC deployment since it depends on the sample app, and previously some scrapes failed because sample app was not ready but AOC was.


wyTrivail (approved):
Thanks

PR #154, opened 2020-12-01
Update default trace template + Move Springboot

NathanielRN:
### Description

Makes the template for sample app spans more general to acommodate the discrepancies between the SDKs. These simpler template should help enforce certain fields to show up in AWS X-Ray and allow others to be included by the language SDK without causing the test to fail.

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/fileconfigs/ExpectedTrace.java
+++ b/validator/src/main/java/com/amazon/aoc/fileconfigs/ExpectedTrace.java
@@ -26,7 +26,11 @@ public enum ExpectedTrace implements FileConfig {
   XRAY_SDK_HTTP_EXPECTED_TRACE("/expected-data-template/xraySDKexpectedHTTPTrace.mustache"),
   SPARK_SDK_HTTP_EXPECTED_TRACE("/expected-data-template/spark/sparkAppExpectedHTTPTrace.mustache"),
   SPARK_SDK_AWSSDK_EXPECTED_TRACE(
-          "/expected-data-template/spark/sparkAppExpectedAWSSDKTrace.mustache")
+          "/expected-data-template/spark/sparkAppExpectedAWSSDKTrace.mustache"),
+  SPRINGBOOT_SDK_HTTP_EXPECTED_TRACE(
+    "/expected-data-template/springboot/springbootAppExpectedHTTPTrace.mustache"),
+  SPRINGBOOT_SDK_AWSSDK_EXPECTED_TRACE(
+    "/expected-data-template/springboot/springbootAppExpectedAWSSDKTrace.mustache")
   ;
 
   private String path;
--- a/validator/src/main/resources/expected-data-template/otelSDKexpectedAWSSDKTrace.mustache
+++ b/validator/src/main/resources/expected-data-template/otelSDKexpectedAWSSDKTrace.mustache
@@ -1,8 +1,6 @@
 [{
-  "name": "/aws-sdk-call",
   "fault": false,
   "error": false,
-  "throttle": false,
   "http": {
     "request": {
       "url": "{{endpoint}}/aws-sdk-call",
@@ -12,47 +10,27 @@
       "status": 200
     }
   },
+  "metadata": {
+    "default": {
+      "otel.resource.telemetry.sdk.name": "opentelemetry"
+    }
+  },
   "subsegments": [
     {
+      "name": "S3",
       "fault": false,
       "error": false,
-      "throttle": false,
-      "subsegments": [
-        {
-          "name": "S3",
-          "fault": false,
-          "error": false,
-          "throttle": false,
-          "http": {
-            "request": {
-              "url": "https://s3.{{region}}.amazonaws.com/",
-              "method": "GET"
-            },
-            "response": {
-              "status": 200
-            }
-          },
-          "namespace": "aws",
-          "subsegments": [
-            {
-              "name": "s3.{{region}}.amazonaws.com",
-              "fault": false,
-              "error": false,
-              "throttle": false,
-              "http": {
-                "request": {
-                  "url": "https://s3.{{region}}.amazonaws.com/",
-                  "method": "GET"
-                },
-                "response": {
-                  "status": 200
-                }
-              },
-              "namespace": "remote"
-            }
-          ]
+      "http": {
+        "response": {
+          "status": 200
         }
-      ]
+      },
+      "metadata": {
+        "default": {
+          "aws.service": "s3"
+        }
+      },
+      "namespace": "aws"
     }
   ]
 }]
--- a/validator/src/main/resources/expected-data-template/otelSDKexpectedHTTPTrace.mustache
+++ b/validator/src/main/resources/expected-data-template/otelSDKexpectedHTTPTrace.mustache
@@ -1,8 +1,6 @@
 [{
-  "name": "/outgoing-http-call",
   "fault": false,
   "error": false,
-  "throttle": false,
   "http": {
     "request": {
       "url": "{{endpoint}}/outgoing-http-call",
@@ -12,29 +10,26 @@
       "status": 200
     }
   },
+  "metadata": {
+    "default": {
+      "otel.resource.telemetry.sdk.name": "opentelemetry"
+    }
+  },
   "subsegments": [
     {
+      "name": "HTTP GET",
       "fault": false,
       "error": false,
-      "throttle": false,
-      "subsegments": [
-        {
-          "name": "aws.amazon.com",
-          "fault": false,
-          "error": false,
-          "throttle": false,
-          "http": {
-            "request": {
-              "url": "https://aws.amazon.com/",
-              "method": "GET"
-            },
-            "response": {
-              "status": 200
-            }
-          },
-          "namespace": "remote"
+      "http": {
+        "request": {
+          "url": "https://aws.amazon.com/",
+          "method": "GET"
+        },
+        "response": {
+          "status": 200
         }
-      ]
+      },
+      "namespace": "remote"
     }
   ]
 }]
--- a/validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedAWSSDKTrace.mustache
+++ b/validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedAWSSDKTrace.mustache
@@ -1,0 +1,54 @@
+[{
+  "name": "/aws-sdk-call",
+  "fault": false,
+  "error": false,
+  "http": {
+    "request": {
+      "url": "{{endpoint}}/aws-sdk-call",
+      "method": "GET"
+    },
+    "response": {
+      "status": 200
+    }
+  },
+  "subsegments": [
+    {
+      "fault": false,
+      "error": false,
+      "subsegments": [
+        {
+          "name": "S3",
+          "fault": false,
+          "error": false,
+          "http": {
+            "request": {
+              "url": "https://s3.{{region}}.amazonaws.com/",
+              "method": "GET"
+            },
+            "response": {
+              "status": 200
+            }
+          },
+          "namespace": "aws",
+          "subsegments": [
+            {
+              "name": "s3.{{region}}.amazonaws.com",
+              "fault": false,
+              "error": false,
+              "http": {
+                "request": {
+                  "url": "https://s3.{{region}}.amazonaws.com/",
+                  "method": "GET"
+                },
+                "response": {
+                  "status": 200
+                }
+              },
+              "namespace": "remote"
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}]
--- a/validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedHTTPTrace.mustache
+++ b/validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedHTTPTrace.mustache
@@ -1,0 +1,37 @@
+[{
+  "name": "/outgoing-http-call",
+  "fault": false,
+  "error": false,
+  "http": {
+    "request": {
+      "url": "{{endpoint}}/outgoing-http-call",
+      "method": "GET"
+    },
+    "response": {
+      "status": 200
+    }
+  },
+  "subsegments": [
+    {
+      "fault": false,
+      "error": false,
+      "subsegments": [
+        {
+          "name": "aws.amazon.com",
+          "fault": false,
+          "error": false,
+          "http": {
+            "request": {
+              "url": "https://aws.amazon.com/",
+              "method": "GET"
+            },
+            "response": {
+              "status": 200
+            }
+          },
+          "namespace": "remote"
+        }
+      ]
+    }
+  ]
+}]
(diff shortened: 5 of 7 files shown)


validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedAWSSDKTrace.mustache
```diff
@@ -0,0 +1,58 @@
...
+    {
+      "fault": false,
+      "error": false,
+      "throttle": false,
+      "subsegments": [
+        {
+          "name": "S3",
+          "fault": false,
+          "error": false,
+          "throttle": false,
```
bhautikpip: can we remove the throttle from here?

validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedAWSSDKTrace.mustache
```diff
@@ -0,0 +1,58 @@
...
+              "status": 200
+            }
+          },
+          "namespace": "aws",
+          "subsegments": [
+            {
+              "name": "s3.{{region}}.amazonaws.com",
+              "fault": false,
+              "error": false,
+              "throttle": false,
```
bhautikpip: from  here too

validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedHTTPTrace.mustache
```diff
@@ -0,0 +1,40 @@
...
+    },
+    "response": {
+      "status": 200
+    }
+  },
+  "subsegments": [
+    {
+      "fault": false,
+      "error": false,
+      "throttle": false,
```
bhautikpip: from here too

validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedHTTPTrace.mustache
```diff
@@ -0,0 +1,40 @@
...
+      "subsegments": [
+        {
+          "name": "aws.amazon.com",
+          "fault": false,
+          "error": false,
+          "throttle": false,
```
bhautikpip: here as well

validator/src/main/resources/expected-data-template/otelSDKexpectedAWSSDKTrace.mustache
```diff
@@ -1,58 +1,37 @@
 [{
-  "name": "/aws-sdk-call",
+  "name": "127.0.0.1:8080",
```
bhautikpip: Do you think all SDKs will generate this name? We should just use the env variable to set the service name so that top level segment name can be similar within all SDKs

validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedAWSSDKTrace.mustache
NathanielRN: Should've removed them from the start, they've been removed!

validator/src/main/resources/expected-data-template/otelSDKexpectedAWSSDKTrace.mustache
NathanielRN: I'm actually not sure, I think this service name makes sense _if_ the sample app has a backend service that is instrumented to create a span with a `http.host` attribute, but if it doesn't, this will break.

In this specific case, the `{{environment}}` variable has `http://127.0.0.1:8080` so it will fail to match `127.0.0.1:8080` 😕 

I think if we count on there being `http.request.url == {{endpoint}}/aws-sdk-call` then we can count on there being a service name of `127.0.0.1:8080` because that _is_ the service. So in that case, it would be _ok_ to leave this.

Otherwise we might consider removing the `"name"` check if the SDK _does not_ add `http.host` or even if it adds `service.name` instead. What do you think?

wyTrivail:
the template will be different among different framework?

validator/src/main/java/com/amazon/aoc/fileconfigs/ExpectedTrace.java
```diff
@@ -26,7 +26,11 @@
   XRAY_SDK_HTTP_EXPECTED_TRACE("/expected-data-template/xraySDKexpectedHTTPTrace.mustache"),
   SPARK_SDK_HTTP_EXPECTED_TRACE("/expected-data-template/spark/sparkAppExpectedHTTPTrace.mustache"),
   SPARK_SDK_AWSSDK_EXPECTED_TRACE(
-          "/expected-data-template/spark/sparkAppExpectedAWSSDKTrace.mustache")
+          "/expected-data-template/spark/sparkAppExpectedAWSSDKTrace.mustache"),
+  SPRINGBOOT_SDK_HTTP_EXPECTED_TRACE(
```
wyTrivail, line 30: what's the difference between the spark template and the springboot template?

bhautikpip: the segment hierarchy diff

NathanielRN:
@wyTrivail It's not the goal, but right now some SDKs instrument the calls with more/less spans than others & with different attributes. Hopefully soon we would be able to have all the SDKs produce the same spans.

For now this PR tries to make the requirement as broad as possible.

validator/src/main/resources/expected-data-template/otelSDKexpectedAWSSDKTrace.mustache
NathanielRN: We spoke about it and decided that for now, we should remove the `"name"` property until we establish something consistent across the SDKs.

validator/src/main/resources/validations/springboot-otel-metric-validation.yml
```diff
@@ -0,0 +1,7 @@
+-
```
bhautikpip: I don't think we would need this config we can just use the default one we have. let's try to reduce the possible templates.

validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedHTTPTrace.mustache
```diff
@@ -0,0 +1,37 @@
...
+      "fault": false,
+      "error": false,
+      "subsegments": [
+        {
+          "name": "aws.amazon.com",
+          "fault": false,
+          "error": false,
+          "http": {
+            "request": {
+              "url": "https://aws.amazon.com/",
```
wyTrivail, line 25: sometimes "https://aws.amazon.com/" return 500/504 so that error/fault equal to true, how do we deal with it?

wyTrivail: if we just want to test the outgoing-http-call, maybe just calling the sample app itself will make the test more stable

wyTrivail: just got 500 in our canary test. here is the json body we got from xray, you can see status is 500 and error is true

                "subsegments": [
                    {
                        "id": "9d629cbd0d36a450",
                        "name": "aws.amazon.com",
                        "start_time": 1606969615.2203882,
                        "end_time": 1606969615.2429497,
                        "fault": false,
                        "error": true,
                        "http": {
                            "request": {
                                "url": "https://aws.amazon.com",
                                "method": "GET"
                            },
                            "response": {
                                "status": 500,
                                "content_length": 0
                            }
                        },
                        "aws": {
                            "xray": {
                                "auto_instrumentation": false,
                                "sdk_version": "1.1.0",
                                "sdk": "X-Ray for Go for Go"
                            }
                        },
                        "metadata": {
                            "default": {
                                "http.response_content_length": ""
                            }
                        },
                        "namespace": "remote",
                        "subsegments": [
                            {
                                "id": "a35b9f0af395066e",
                                "name": "request",
                                "start_time": 1606969615.2204092,
                                "end_time": 1606969615.2205038,
                                "fault": false,
                                "error": false,
                                "aws": {
                                    "xray": {
                                        "auto_instrumentation": false,
                                        "sdk_version": "1.1.0",
                                        "sdk": "X-Ray for Go for Go"
                                    }
                                }
                            },
                            {
                                "id": "4e70f38632f72310",
                                "name": "response",
                                "start_time": 1606969615.2205071,
                                "end_time": 1606969615.2429152,
                                "fault": false,
                                "error": false,
                                "aws": {
                                    "xray": {
                                        "auto_instrumentation": false,
                                        "sdk_version": "1.1.0",
                                        "sdk": "X-Ray for Go for Go"
                                    }
                                }
                            }
                        ]
                    }
                ]
            },
            "Id": "db8287ae0cb92c4d"
        },

wyTrivail (changes requested):

validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedHTTPTrace.mustache
NathanielRN: That's really interesting that it returns 500, I would expect it to never return 500... Is this avoided if we pick a different endpoint?

NathanielRN: > if we just want to test the outgoing-http-call, maybe just calling the sample app itself will make the test more stable

Do you mean not including the HTTP call in the App route method? I think that's reasonable for a different route, like even the `/` route, but I think the point of the `/outgoing-http-call` is to make a request to an external service so we should probably keep that validation... although we should resolve this 500 response code issue 😕

bhautikpip: I don't think `https://aws.amazon.com` should return 500/504 unless it's down right?

validator/src/main/resources/expected-data-template/otelSDKexpectedHTTPTrace.mustache
```diff
@@ -12,29 +10,26 @@
...
     }
   },
+  "metadata": {
+    "default": {
+      "otel.resource.telemetry.sdk.name": "opentelemetry"
+    }
+  },
   "subsegments": [
     {
+      "name": "HTTP GET",
```
KKelvinLo, line 20: Can we remove this? The name differs between the SDKS for example the segment mapped by JS is the name of the query "aws.amazon.com".

NathanielRN: The [OpenTelemetry Specifications on HTTP](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md) say that HTTP client spans SHOULD follow this conventation:

> Therefore, HTTP client spans SHOULD be using conservative, low cardinality names formed from the available parameters of an HTTP request, such as "HTTP {METHOD_NAME}"

Only if it is a HTTP server span should it look like the route `/outgoing-http-call`

> In case of HTTP servers, these endpoints are often mapped by the server frameworks to more concise HTTP routes, e.g. /api/users/{user_id}, which are recommended as the low cardinality span names.

`aws.amazon.com` doesn't fit into either of those situations, do you think an issue in the JS SDK would be a better solution?

validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedHTTPTrace.mustache
wyTrivail: make sense, then maybe don't validate the error, fault and statuscode key?

NathanielRN: Do you know why it returns a 500 status code? I don't expect endpoints to ever return 500+ status codes, maybe we should pick a different endpoint?

bhautikpip: I think we should because that is very essential to test. That will tell us about any kind of throttle or fault  span issues. We must keep those in validation.

validator/src/main/resources/validations/springboot-otel-metric-validation.yml
NathanielRN: Sounds good, I will remove the Springboot templates

validator/src/main/resources/expected-data-template/springboot/springbootAppExpectedHTTPTrace.mustache
wyTrivail: Then we need to figure out a way, we saw 500 from time to time

validator/src/main/resources/expected-data-template/otelSDKexpectedHTTPTrace.mustache
```diff
@@ -12,29 +10,26 @@
...
-          "fault": false,
-          "error": false,
-          "throttle": false,
-          "http": {
-            "request": {
-              "url": "https://aws.amazon.com/",
-              "method": "GET"
-            },
-            "response": {
-              "status": 200
```
NathanielRN, line 32: @wyTrivail The test framework was already validating status code 200 before, this PR does not change that, so I think it's out of scope for the changes here.

wyTrivail: i suspect this template would make the test unstable even though it's already there. in collector test, we use the spark one, as you can see there we removed lots of items...

wyTrivail (approved):
i'm approving this pr for now. If you find the test unstable in the future, you should be able to modify the template.

but i want to call out aws.amazon.com does return 500/504 from time to time in our testing, which also doesn't mean it's down. A website is down when its availability down to almost 0.

validator/src/main/resources/expected-data-template/otelSDKexpectedHTTPTrace.mustache
KKelvinLo: For JS one the current expected template will not work due to some minor issues such as status code/names. I think the question is should we simplify it like the Spark one so that it can work for JS as well and be more stable?

bhautikpip: I think we can remove all the names for now but I believe status code should be `200`

wyTrivail: i think it might take some time for us to find the balance between complexity and stability.  i'm okay with the current the template since we can improve it later. we don't need make it perfect at the beginning. 

in additional, we should always try our best to add unit tests to validate the data structure instead of relying on end to end test, which is consuming multi resources and hard to debug. 

for the js sdk one, maybe you can try to use the spark template, let me know if there's any field js sdk can't support.

NathanielRN: To keep the validation on `HTTP GET` for the span name (which I think is useful), we could merge this PR for now and then create a new one for JavaScript later. We should create an issue in the JavaScript repo that includes @KKelvinLo 's concerns and link it to that PR to remove the custom JavaScript template when the fixes are included.

KKelvinLo: @wyTrivail I think that's good justification, we can always improve it later and this does look good for now.

KKelvinLo (approved):
lgtm, can file PRs to address any future changes

PR #163, opened 2020-12-03
Sleep 30s before starting docker compose on sample app

wyTrivail:
in case the dockerd service is not totally set up. 

saw some errors while running ec2 test

```
ERROR: for tmp_sample_app_1  Cannot start service sample_app: unable to remount dir as readonly: mount tmpfs:/var/lib/docker/containers/4098d40e33e4f505bc8717f182237fe5505f33104511aaf39b9f755d577e425a/mounts/secrets, flags: 0x21, data: uid=0,gid=0: device or resource busy
```


mxiamxia (approved):
Thanks

PR #165, opened 2020-12-03
use double to sort spans

wyTrivail:
instead of string, to fix the failure in the workflow:

https://github.com/aws-observability/aws-otel-collector/runs/1492729910?check_suite_focus=true


bhautikpip (approved):
Looks good :) Thank you for finding this.

PR #166, opened 2020-12-03
Add data mode to performance.json

kohrapha:
This PR adds the `dataMode` attribute to `performance.json` so that we can split up the Performance Report tables by metrics and traces.

I've tested this locally and a sample output is:
```json
{
    "testcase": "otlp_metric",
    "instanceType": "m5.2xlarge",
    "testingAmi": "soaking_linux",
    "dataType": "otlp",
    "dataMode": "metric",
    "dataRate": 5000,
    "avgCpu": 0.03332926669302766,
    "avgMem": 56.79377066666667,
    "commitId": "dummy_commit",
    "collectionPeriod": 2
}
```

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/models/PerformanceResult.java
+++ b/validator/src/main/java/com/amazon/aoc/models/PerformanceResult.java
@@ -27,6 +27,7 @@ public class PerformanceResult implements Serializable {
   @NonNull private String testingAmi;
 
   @NonNull private String dataType;
+  @NonNull private String dataMode;
   @NonNull private Integer dataRate;
 
   // Calculated average statistics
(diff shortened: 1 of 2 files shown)


validator/src/main/java/com/amazon/aoc/models/PerformanceResult.java
```diff
@@ -27,6 +27,7 @@
   @NonNull private String testingAmi;
 
   @NonNull private String dataType;
+  @NonNull private String dataMode;
```
wyTrivail, line 30: should terraform send datamode to validator?

kohrapha: It's already being sent via config.

kohrapha: https://github.com/aws-observability/aws-otel-test-framework/blob/e550c5d523133f5df82dff802cd93e51606f992e/validator/src/main/java/com/amazon/aoc/models/ValidationConfig.java#L50

wyTrivail (approved):
thanks

Issue #51, opened 2020-10-29
Trace data validation issue : Can't use to validate complicated trace data


JohnWu20:
Actually trace data is a tree whose nodes are segments. When xray SDK emits a trace data which has more than two levels segments, the validator can't validate the segments. So even the template trace data is same with the retrieved trace data from xray API, the integration test will still fail. 

For now, we just simple the validation templates which will just validate two level's segments data. 

Mark this issue here, in case someone else will meet the same problem. We can improve the validator's ability in the future.

PR #169, opened 2020-12-03
Add support for Lambda trace validation

willarmiros:
This PR adds the required templates for validating an AWS-observability-instrumented Lambda function. Lambda segments have two susbsegments (invocation and overhead) that we are validating, but they are not guaranteed to be in the order expected by the validation template. So we sort the subsegments by mapping them to a barebones Entity POJO and recursively sorting all nested subsegments in a segment. Because the Entity class can accept Segments as well, I removed the existing segment-sorting logic and used the new recursive sorter instead.

The drawback of casting to the Entity class is that it will throw an exception if there are top-level fields on the retrieved (sub)segment that are not modeled in the current Entity class. But new fields are added to the entity model rarely, so it's probably worth having an exception and addressing such a new field in that event anyway.

The current Lambda trace template has function names specific to my account, but I plan on addressing that by adding regex support in a future PR.


willarmiros:
This kind of addresses #51, but I think a more thorough solution of removing flattening altogether and replacing it with JSON subset detection is required to fully address that issue.

wyTrivail:
can @bhautikpip take a look?

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
```diff
@@ -64,23 +73,35 @@ public void validate() throws Exception {
...
+        for (Map.Entry<String, Object> entry : storedTrace.entrySet()) {
+          String targetKey = entry.getKey();
+          if (retrievedTrace.get(targetKey) == null) {
+            log.error("mis target data: {}", targetKey);
+            throw new BaseException(ExceptionCode.DATA_MODEL_NOT_MATCHED);
+          }
+          if (!entry
+                  .getValue()
+                  .toString()
+                  .equalsIgnoreCase(retrievedTrace.get(targetKey).toString())) {
```
wangzlei, line 92: Can we support regex here? If not, we can remove `name` from template like serverlessrepo-aot-py38-sar/api, serverlessrepo-aot-py38-sar-function-XMKMYMQM3776

willarmiros: Yeah I'm going to add regex support in a separate PR, didn't want to clutter this one too much. Will remove the `name` in the meantime.

validator/src/main/java/com/amazon/aoc/helpers/SortUtils.java
```diff
@@ -0,0 +1,36 @@
...
+   *
+   * @param entities - list of X-Ray entities to sort recursively. Modified in place.
+   */
+  public static void recursiveEntitySort(List<Entity> entities) {
+    if (entities == null || entities.size() == 0) {
+      return;
+    }
+
+    for (Entity entity : entities) {
+      if (entity.getSubsegments() != null && entity.getSubsegments().size() > 1) {
```
bhautikpip: I think it's a good idea if we keep recursion depth limit. Major reason being since this is validation tests we can absolutely control the trace data. Also, we wouldn't expect data more than certain depth and if we do program should not go in the loop. I don't see any point of doing recursion after some depth especially in testing scenario. breaking condition looks fine but i strongly suggest we should also break the recursion by some level of depth (maybe 6?).

bhautikpip:
I think over all code looks good to me. I would suggest that we should run tests on springboot and spark docker image before merging in this change in order to avoid any issues beforehand.

validator/src/main/java/com/amazon/aoc/helpers/SortUtils.java
willarmiros: That's a good callout, I can add a limit to prevent infinite looping if, say, a subsegment had itself as a child (which wouldn't be possible in practice but still good to check for). This also made me realize another bug with the stop condition that it shouldn't just ignore if an entity only has a single child, because that single child could itself have many children that need sorting. So I'll address that too.

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
```diff
@@ -113,30 +134,33 @@ public void validate() throws Exception {
...
 
-    // build the segment's document as a jsonarray and flatten it for easy comparison
-    StringBuilder segmentsJson = new StringBuilder("[");
+    // Parse retrieved segment documents into a barebones Entity POJO
     for (Segment segment : segmentList) {
-      segmentsJson.append(segment.getDocument());
-      segmentsJson.append(",");
+      Entity entity;
+      try {
+        entity = MAPPER.readValue(segment.getDocument(), Entity.class);
```
bhautikpip, line 143: Also, not sure how often this scenario would occur but do you think mapper should not be able to map some values from document to the class you have created? If this is the case then we will also have to do some handling on fields because for example let's say if mapper didn't translate `name` then validation would fail if we are checking `name` value in stored validation template.

willarmiros: Hmm well `Entity` has all the fields that we support in the (sub)segment model, in addition to `inferred` which is added by the backend service. If we want to validate against any field that isn't yet a part of the Entity class, we can just add it to the Entity class so it's read properly then validate it normally. Like I mentioned new fields for entities are allow-listed pretty rarely, so I think it's acceptable to have to update that class when they do come in.

wangzlei (approved):
looks good to me, thanks!

validator/src/main/java/com/amazon/aoc/validators/TraceValidator.java
bhautikpip: sounds fair.

willarmiros:
@bhautikpip addressed comments :)

bhautikpip (approved):

wyTrivail:
can we use squash to merge so that there's only one commit generated in main branch?

PR #171, opened 2020-12-04
Update springboot-otel-trace-metric-validation.yml

bhautikpip:
fix: springboot template metric validation type


anuraaga (approved):
No idea what this is doing but if it fixes my test failure LGTM+ :)

PR #174, opened 2020-12-04
Expose transactions per minute via mocked server endpoint

kohrapha:
In this PR, we keep track of the number of transactions hitting the `/put-data` endpoint of the mocked server and expose the average number of transactions per minute via the `/tpm` `GET` endpoint.

### DIFF
--- a/mocked_server/index.js
+++ b/mocked_server/index.js
@@ -6,14 +6,25 @@ const http = require('http');
 const fs = require('fs');
 const app = express();
 
-var get_data = ""
+let data = "";
+let numTransactions = 0;
+const startTime = new Date();
 
-app.get("/check-data", function(req, res){
-    res.send(get_data);
+// Retrieve number of transactions per minute
+app.get("/tpm", function (req, res) {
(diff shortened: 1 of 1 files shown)


mocked_server/index.js
```diff
@@ -6,14 +6,25 @@ const http = require('http');
...
+});
+
+app.get("/check-data", function (req, res) {
+    res.send(data);
 });
 
 app.all('/put-data*', function (req, res) {
-    get_data = "success";
+    data = "success";
+    numTransactions++;
```
wyTrivail, line 27: there's a concurrent issue since ++ is not atomic

wyTrivail (changes requested):

mocked_server/index.js
kohrapha: Since Node.js is single-threaded, all operations on global vars are atomic by default.

wyTrivail: oh yeah, make sense, it will only return once i/o happens

wyTrivail (approved):

PR #176, opened 2020-12-05
Add test case for gRPC exporter

shaochengwang:
**Description:**
Add new test case for gRPC exporter

**Tests:**
Manually run tests for `terraform/ec2`.

```
cd terraform/ec2 && terraform init && terraform apply -auto-approve -lock=false -var-file="../testcases/otlp_grpc_exporter_metric_mock/parameters.tfvars" -var="aoc_version=latest" -var="testcase=../testcases/otlp_grpc_exporter_metric_mock"
cd terraform/ec2 && terraform init && terraform apply -auto-approve -lock=false -var-file="../testcases/otlp_grpc_exporter_trace_mock/parameters.tfvars" -var="aoc_version=latest" -var="testcase=../testcases/otlp_grpc_exporter_trace_mock"
```
Also made corresponding changes in ECS and EKS but haven't manually tested yet. I will test it out and send a separate PR if necessary.


mocked_server/grpc_metrics/metrics_handler_server.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+function CheckData(call, callback) {
+    callback(null, {message: get_data});
+}
+
+/**
+ * Starts an RPC server that receives requests for the data handler service at the
+ * sample server port
+ */
+function main() {
```
wyTrivail: can this server also expose a http port 8080 as well so that we don't need to write code in validator to call grpc?

mocked_server/grpc_trace/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:14
```
wyTrivail: for now, we have three mocked_server,
1. https mocked server
2. grpc metric mocked server
3. grpc trace mocked server.

can we create a folder named `mocked_servers` and subfolders for each mocked server. 

then, in the imagebuild folder, we make some modification to loop through all the subfolder under `mocked_servers`, build them and push them to ecr with its name as the image tag.  just like how we do for sample_apps: https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/imagebuild/main.tf#L38

after that, we can define a var `mocked_server` in common.tf just like `sample_app` https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/imagebuild/main.tf#L38,  use `mocked_server` if `mocked_server_image` is not provided. and get the actual mocked_server image link base on the `mocked_server` value just like how we do for sample apps here: https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/basic_components/outputs.tf#L45

wyTrivail: we might also define a var for mocked_server_port since different mocked_server would use different port.

wyTrivail (changes requested):

mocked_server/grpc_trace/Dockerfile
wyTrivail: will that be possible we can all use 443 for the grpc port, so that we don't need to specify the port?

mocked_server/grpc_metrics/metrics_handler_server.js
wyTrivail: to get validator to access the grpc from public internet, we will need to create a lb in front of it. if we can have the 8080 http port, then we don't need to change the code in ecs and eks

terraform/templates/local/docker_compose.tpl
```diff
@@ -2,7 +2,7 @@ version: "3.8"
 services:
   mocked-server:
     build:
-      context: ../../mocked_server
+      context: ../../mocked_servers
```
wyTrivail: it should be `../../mocked_servers/${mocked_server}`?

terraform/templates/local/docker_compose_from_source.tpl
```diff
@@ -2,7 +2,7 @@ version: "3.8"
 services:
   mocked-server:
     build:
-      context: ../../mocked_server
+      context: ../../mocked_servers/https
```
wyTrivail: it should be ../../mocked_servers/${mocked_server}?

terraform/testcases/otlp_grpc_exporter_metric_mock/otconfig.tpl
```diff
@@ -0,0 +1,26 @@
...
+  logging:
+    loglevel: debug
+  otlp:
+    endpoint: ${mock_endpoint}
+    insecure: true
+
+service:
+  pipelines:
+    metrics:
+      receivers: [otlp]
```
wyTrivail, line 23: can you add a batch here?

terraform/testcases/otlp_grpc_exporter_metric_mock/otconfig.tpl
```diff
@@ -0,0 +1,26 @@
...
+    endpoint: 0.0.0.0:1777
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:${grpc_port}
+
+processors:
+  batch/metrics:
+    timeout: 60s
```
wyTrivail: can we remove the timeout so that all the testcases will use the default batch config

wyTrivail (commented):
thanks, just some small nits. could you also run a test with ecs/eks for otlp_mock to ensure this change won't break it?

wyTrivail (approved):
thanks!

PR #183, opened 2020-12-08
Remove fault values from expected trace templates

NathanielRN:
Because of the update to the collector in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/1740, a span with no status set will _not_ produce a `fault` value.

This PR removes the expectation that the `fault` value is set.

### DIFF
--- a/validator/src/main/resources/expected-data-template/otelSDKexpectedAWSSDKTrace.mustache
+++ b/validator/src/main/resources/expected-data-template/otelSDKexpectedAWSSDKTrace.mustache
@@ -1,6 +1,4 @@
 [{
-  "fault": false,
-  "error": false,
   "http": {
     "request": {
       "url": "{{endpoint}}/aws-sdk-call",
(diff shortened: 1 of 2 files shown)


anuraaga (commented):
Ideally the matcher could be smart enough to treat false and missing as the same. But this might be a reasonable short-term fix.

That being said - is it only fault that's not set? I thought error would have the same change.

wyTrivail:
do we really care the values of some of the fields like fault and error?

wyTrivail (approved):

NathanielRN:
@anuraaga You were definitely right! Thanks for the call out, I've removed the `error` field 😄

NathanielRN:
@wyTrivail Hm I think it's a good way to validate what the collector is doing? For example if we add a test in the future that _expects_ a failure to occur, we could check that `fault` was `true` and make sure it was not `false`.

I wouldn't say it's a huge priority though.

PR #184, opened 2020-12-08
Add JS expected trace and metric template

KKelvinLo:
This PR adds a data-validation template for JavaScript for both traces and metrics.


wyTrivail (approved):
thanks

PR #195, opened 2020-12-10
Add case-insensitive regex support

willarmiros:
Copy of #192, this time with case-insensitive comparisons. Tested with values that had mismatched case this time. Also added `UnusedImports` to checkstyle.

### DIFF
--- a/validator/src/main/resources/expected-data-template/otelSDKexpectedHTTPTrace.mustache
+++ b/validator/src/main/resources/expected-data-template/otelSDKexpectedHTTPTrace.mustache
@@ -18,7 +18,7 @@
       "name": "HTTP GET",
       "http": {
         "request": {
-          "url": "https://aws.amazon.com/",
+          "url": "https://aws\\.amazon\\.com/",
           "method": "GET"
         },
         "response": {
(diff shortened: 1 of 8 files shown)


bhautikpip (approved):
LGTM

validator/src/main/resources/expected-data-template/otelSDKexpectedHTTPTrace.mustache
```diff
@@ -18,7 +18,7 @@
       "name": "HTTP GET",
       "http": {
         "request": {
-          "url": "https://aws.amazon.com/",
+          "url": "https://aws\\.amazon\\.com/",
```
wyTrivail, line 21: you might need to change all the trace templates?

wyTrivail (changes requested):

validator/src/main/resources/expected-data-template/otelSDKexpectedHTTPTrace.mustache
willarmiros: Ah I missed all the ones nested in packages. Good catch!

willarmiros: Also FWIW, even if a period wasn't escaped like this (for example in the interpolated `endpoint` variable), it wouldn't break validation. Because the unescaped period matches any character, including a literal period, so the correct endpoint would still be considered a match. But of course it's still better to escape them and match it exactly where we can.

wyTrivail: gotcha!

wyTrivail (approved):

PR #200, opened 2020-12-11
Remove nested subsegment for AWS SDK call

anuraaga:
I think we may have had a nested transport span in previous versions but not anymore. At least, that's how I read the failure n https://github.com/aws-observability/aws-otel-java-instrumentation/runs/1535106837


wyTrivail (approved):

NathanielRN:
Confirmed in text diff that this is the only difference!

image

PR #205, opened 2020-12-16
added ecs support for prometheus static and mock

amanbrar1999:
* Added ecs task definition files for the prometheus sample app within `prometheus_static` and `prometheus_mock` (task definitions are identical in both directories)
* increased retries in mock validator because ecs containers were taking long to start and got dangerously close to 10 retries
* added some parameters in ecs terraform files that need to be passed into the otconfig and ecs taskdef for prometheus tests
* updated aoc image reference to the latest public ecr image. Let me know if this is not desired.


terraform/common.tf
```diff
@@ -31,11 +31,11 @@ variable "sample_app_image" {
 }
 
 variable "aoc_image_repo" {
-  default = "611364707713.dkr.ecr.us-west-2.amazonaws.com/aws/aws-otel-collector"
+  default = "public.ecr.aws/aws-observability/aws-otel-collector"
```
amanbrar1999, line 34: I updated this to the public ecr image, let me know if this is not desired

terraform/common/variables.tf
```diff
@@ -38,9 +38,9 @@ variable "aoc_vpc_name" {
 }
 
 variable "aoc_image_repo" {
-  default = "josephwy/ttt"
+  default = "public.ecr.aws/aws-observability/aws-otel-collector"
```
amanbrar1999, line 41: same here but I think this gets overwritten by the other one anyways

terraform/common.tf
```diff
@@ -31,11 +31,11 @@ variable "sample_app_image" {
...
 }
 
 variable "aoc_version" {
-  default = "v0.4.0-382067259"
+  default = "v0.6.0"
```
hossain-rayhan: Wondering, why can't we make it `latest`?

amanbrar1999: We probably can, I just did a specific version because previously a specific version was also used

validator/src/main/java/com/amazon/aoc/enums/GenericConstants.java
```diff
@@ -22,7 +22,7 @@
   // retry
   SLEEP_IN_MILLISECONDS("10000"), // ms
   SLEEP_IN_SECONDS("30"),
-  MAX_RETRIES("10"),
+  MAX_RETRIES("20"),
```
hossain-rayhan: nit: Can we rather increase the sleep time than mex_retries? Then we will see less logs in the terminal which is good in my opinion.

amanbrar1999: good point, will change

hossain-rayhan (approved):
Added two minor comments. Overall, LGTM.

terraform/common.tf
mxiamxia: since the integration test is testing the latest build that not being released. `latest` won't work. It will be always override from the terraform input. We can leave it as is.

terraform/testcases/prometheus_mock/ecs_taskdef.tpl
```diff
@@ -0,0 +1,101 @@
+[
```
mxiamxia: Can we move all the template files in `templates` folder under `terraform` folder? in this case, we can create `prometheus` folder under `templates` folder for this template. The test case folder only leave `.var` and `otconfig.tpl` for all the testcases

mxiamxia: can you pls help rename this file to `ecs_taskdef_sidecar.tpl` so we can differentiate it with other deployment mode?

terraform/testcases/prometheus_static/ecs_taskdef.tpl
```diff
@@ -0,0 +1,101 @@
+[
```
mxiamxia: ditto

terraform/common.tf
mxiamxia: The integration should always test against on the newly built images that's why you see "v0.4.0-38xxxx". We always take it from terraform input in the workflow. So a fixed version number or `latest` doesnt' matter too much here.

terraform/testcases/prometheus_mock/ecs_taskdef.tpl
amanbrar1999: 👍  will do

amanbrar1999: hmm both are still sidecar deployments though right?

amanbrar1999: the only difference here is that I changed the sample app, as prometheus sample app takes different environment variable inputs than spark sample app

mxiamxia: > hmm both are still sidecar deployments though right?

Yea, you sent two examples - deploy collector as sidecar with sample app, with file name `ecs_taskdef.tpl `. I mean what if in the future we want to deploy the collector as DaemonSet or a Pure service, what would be the example file name?

mxiamxia: I've done some example configs to deploy the collector as  service, daemonset and sidecar before. I gave each example config a different name. 
https://github.com/aws-observability/aws-otel-collector/tree/eks_daemonset/examples/eks

mxiamxia (approved):
Thanks

PR #208, opened 2021-01-05
add soaking test validation metrics

mxiamxia:
Emit soaking test validation metrics for setting up the dashboard.
image

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/validators/AlarmPullingValidator.java
+++ b/validator/src/main/java/com/amazon/aoc/validators/AlarmPullingValidator.java
@@ -8,6 +8,8 @@ import com.amazon.aoc.helpers.RetryHelper;
 import com.amazon.aoc.models.Context;
 import com.amazon.aoc.models.ValidationConfig;
 import com.amazon.aoc.services.CloudWatchAlarmService;
+import com.amazon.aoc.services.CloudWatchService;
+import com.amazonaws.services.cloudwatch.model.Dimension;
 import com.amazonaws.services.cloudwatch.model.MetricAlarm;
 import lombok.extern.log4j.Log4j2;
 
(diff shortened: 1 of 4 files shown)


validator/src/main/java/com/amazon/aoc/validators/AlarmPullingValidator.java
```diff
@@ -21,6 +23,11 @@
   private Integer pullDuration;
   private Integer pullTimes;
   private CloudWatchAlarmService cloudWatchAlarmService;
+  private CloudWatchService cloudWatchService;
+
+  private static final String SOAKING_NAMESPACE = "AWSOtelCollector/SoakingTest";
```
wyTrivail, line 28: can this namespace be sent from terraform to validator instead of define a constant here?

wyTrivail (approved):

validator/src/main/java/com/amazon/aoc/validators/AlarmPullingValidator.java
```diff
@@ -64,11 +76,14 @@ public void validate() throws Exception {
...
                   metricAlarm.getStateValue(),
                   metricAlarm.getMetrics());
+              cloudWatchService.putMetricData(SOAKING_NAMESPACE, METRIC_NAME, 0.0, dimension);
               System.exit(1);
             }
           }
 
-          log.info("alarms look good, continue to bake");
+          log.info("all alarms look good, continue to bake");
+          cloudWatchService.putMetricData(SOAKING_NAMESPACE, METRIC_NAME, 1.0, dimension);
```
wyTrivail, line 85: i think we just emit one metric after all the retries, but looks like we are going to emit metrics for every retry?

mxiamxia: yes, it will emit metrics for every pulling alarm check. I feel not too much difference on emitting the metrics for each retry and all retries. For every retry gets better granularity on the metrics.

PR #210, opened 2021-01-06
Expand ec2 test for more amis and arm

wyTrivail:


terraform/ec2/amis.tf
```diff
@@ -65,26 +65,185 @@ EOF
...
+      os_family = "centos"
+      ami_search_pattern = "CentOS Linux 6 x86_64 HVM EBS 2002*"
+      family = "linux"
+      ami_owner = "679593333241"
+      arch = "amd64"
+      user_data = <<EOF
+#! /bin/bash
+sudo iptables -I INPUT -p tcp -m tcp --dport 55680 -j ACCEPT
+sudo iptables -I INPUT -p tcp -m tcp --dport 55690 -j ACCEPT
+sudo service iptables save
```
mxiamxia, line 185: just curious. why CentOS need update iptables but other OSs do not need?

mxiamxia (approved):

terraform/ec2/amis.tf
wyTrivail: Yeah, centos6 blocks almost all the ports by default

PR #213, opened 2021-01-08
fix the iptables command for ami centos6

wyTrivail:
by allowing udp port 55690 to get xray receiver test passed


shaochengwang (approved):
LGTM!

PR #215, opened 2021-01-13
Metric integ test enhancement

shaochengwang:
Enhance sample app to include all 6 metric types

**Tests:**
```
terraform apply -auto-approve -lock=false -var-file="../testcases/otlp_metric/parameters.tfvars" -var="aoc_version=latest" -var="testcase=../testcases/otlp_metric" -var="testing_ami=amazonlinux2"
```

### DIFF
--- a/sample-apps/spark/Dockerfile
+++ b/sample-apps/spark/Dockerfile
@@ -7,7 +7,7 @@ COPY ./settings.gradle ./settings.gradle
 
 RUN gradle build
 RUN tar -xvf build/distributions/spark-sample-app-1.0.tar
-RUN wget https://github.com/aws-observability/aws-otel-java-instrumentation/releases/download/v0.10.1-aws.1/aws-opentelemetry-agent.jar
+RUN wget https://github.com/aws-observability/aws-otel-java-instrumentation/releases/download/v0.12.1-aws.1/aws-opentelemetry-agent.jar
 
 FROM amazoncorretto:11 
 
--- a/sample-apps/spark/Dockerfile
+++ b/sample-apps/spark/Dockerfile
@@ -7,7 +7,7 @@ COPY ./settings.gradle ./settings.gradle
 
 RUN gradle build
 RUN tar -xvf build/distributions/spark-sample-app-1.0.tar
-RUN wget https://github.com/aws-observability/aws-otel-java-instrumentation/releases/download/v0.12.1-aws.1/aws-opentelemetry-agent.jar
+RUN wget https://github.com/aws-observability/aws-otel-java-instrumentation/releases/download/v0.10.1-aws.1/aws-opentelemetry-agent.jar
 
 FROM amazoncorretto:11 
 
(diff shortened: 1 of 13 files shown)


sample-apps/spark/Dockerfile
```diff
@@ -7,7 +7,7 @@ COPY ./settings.gradle ./settings.gradle
 
 RUN gradle build
 RUN tar -xvf build/distributions/spark-sample-app-1.0.tar
-RUN wget https://github.com/aws-observability/aws-otel-java-instrumentation/releases/download/v0.12.1-aws.1/aws-opentelemetry-agent.jar
+RUN wget https://github.com/aws-observability/aws-otel-java-instrumentation/releases/download/v0.10.1-aws.1/aws-opentelemetry-agent.jar
```
mxiamxia: I thought only v0.13.0 doesn't work. Does v0.12.1 also has issue for emitting the compatible metrics?

sample-apps/spark/src/main/java/com/amazon/sampleapp/MetricEmitter.java
```diff
@@ -31,22 +42,63 @@
...
+                    .setChannel(
+                            ManagedChannelBuilder.forTarget(otelExporterOtlpEndpoint).usePlaintext().build())
+                    .build();
+
+    IntervalMetricReader.builder()
+            .setMetricProducers(
+                    Collections.singleton(OpenTelemetrySdk.getGlobalMeterProvider().getMetricProducer()))
+            .setExportIntervalMillis(1000)
+            .setMetricExporter(metricExporter)
+            .build();
```
mxiamxia: why we don't use auto-instr-agent to send metrics anymore?

shaochengwang: I added this when I tested it locally without java-agent and docker. Will remove this part.

sample-apps/spark/Dockerfile
shaochengwang: Same reason as in the other comment. I was testing it locally without java-agent and docker and `v0.12.1` is not compatible with the `IntervalMetricReader` builder. Will change this back.

validator/src/main/resources/expected-data-template/defaultExpectedMetric.mustache
```diff
@@ -11,3 +11,73 @@
     -
       name: statusCode
       value: 200
+
```
wyTrivail, line 14: i'm afraid changing this template would break the test in SDK repo which would probably be using this template as well with another sample app, we might either use another template or let them to use our latest sample app image.

wyTrivail: looped through the sdk repos, where the default template is not used, so it's not a concern.

wyTrivail (approved):

PR #216, opened 2021-01-20
support patching on soak and canary

wyTrivail:
1 adding a flag in ec2 module
2. adding a tag in ec2 instance when patch is enabled
3. check patch status once instance is created.

### DIFF
--- a/terraform/canary/main.tf
+++ b/terraform/canary/main.tf
@@ -54,4 +54,6 @@ module "ec2_setup" {
 
   aws_access_key_id = var.aws_access_key_id
   aws_secret_access_key = var.aws_secret_access_key
+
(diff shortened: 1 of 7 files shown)


terraform/canary/main.tf
```diff
@@ -54,4 +54,6 @@ module "ec2_setup" {
 
   aws_access_key_id = var.aws_access_key_id
   aws_secret_access_key = var.aws_secret_access_key
+
+  patch = true
```
mxiamxia, line 58: can we get the name more specific?

wyTrivail: like `enable_patch`?

mxiamxia (approved):

PR #226, opened 2021-02-03
Set service.name for spring boot app too.

anuraaga:
We weren't setting this before, but now that service.name is required by otel we need it to be a reasonable value.


mxiamxia (approved):
Thanks.

PR #237, opened 2021-03-03
Fix AMI filter and docker installation

vastin:
* Refine the filter string to pick up correct AMI image.
* Install docker by following instructions in [Docker basics for Amazon ECS](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/docker-basics.html)


mxiamxia (approved):
Thanks!

PR #238, opened 2021-03-09
tf: Use latest AL2 AMI and right Debian 10

pingleig:
Follow up on #237

- debian filter need to use product code to avoid getting the dshop
  debian 10 AMI ...
- al2 need to use specific format to avoid getting sql expres ami
- `ami_product_code` is added and every AMI need to have at least an
  empty list so the dynamic block is happy

Reference

- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/finding-an-ami.html#finding-quick-start-ami
- https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-images.html#options
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ami
- https://www.terraform.io/docs/language/expressions/dynamic-blocks.html

### DIFF
--- a/terraform/ec2/amis.tf
+++ b/terraform/ec2/amis.tf
@@ -81,6 +81,7 @@ variable "amis" {
       os_family = "ubuntu"
       ami_search_pattern = "ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-*"
       ami_owner = "099720109477"
+      ami_product_code = []
       family = "debian"
       arch = "amd64"
       login_user = "ubuntu"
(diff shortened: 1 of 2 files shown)


terraform/ec2/amis.tf
```diff
@@ -97,6 +99,9 @@ variable "amis" {
       os_family = "debian"
       ami_search_pattern = "debian-10-amd64*"
       ami_owner = "679593333241"
+      # NOTE: we need product code to pick the right debian 10.
+      ami_product_code = [
+        "auhljmclkudu651zy27rih2x2"]
```
mxiamxia, line 104: Q: why do we only need to apply `product_code` filter for this AMI? How about the others?

pingleig: For AMI owned by amazon, there is no product code, and if you apply the filter with `*`, the API returns nothing. I didn't check other marketplace AMI, based on subscription in the test account we only have subscription for Debian 10 and CentOS.

The reason we need product code for debian 10 is the filter pattern alone will return two products, one from Debian, one from dshop. Those two types of AMIs can only be distinguished using product code.

mxiamxia (approved):
Thanks!

PR #239, opened 2021-03-10
tf: Fix ami for canaray and soaking

pingleig:
In #238 forgot to update the AMI for canary and soaking, they need to have the empty product code and strict al2 filter


mxiamxia (approved):
Thanks!

PR #240, opened 2021-03-11
Update New Relic config

alanwest:
Noted in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/2481:

> `spans_url_override` and `metrics_url_override` has been renamed to `spans_host_override` and `metrics_host_override` to reflect the fact that you only need to specify host[:port] when overriding these endpoints.

I believe this should resolve the issue with the integration tests noted in https://github.com/aws-observability/aws-otel-collector/pull/392.


mxiamxia (approved):
Thanks!

PR #245, opened 2021-03-16
[ECScontainerReceiver]Update the config file for the new resources

JohnWu20:
Update the config file to delete some resource data we don't want to validate.

Test: Local test.


pingleig (approved):
LGTM

PR #246, opened 2021-03-16
Remove jcenter

pingleig:
The build is failing because it can't pull jar from jcenter.

- https://blog.gradle.org/jcenter-shutdown
- https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,9 +12,7 @@ plugins {
 }
 
 repositories {
-    // Use jcenter for resolving dependencies.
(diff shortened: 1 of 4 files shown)


pingleig:
It seems the jcenter issue is intermediate because #245 passed ... but since it's going down, we better remove it now instead of getting surprise later.

bjrara (approved):
/lgtm

mxiamxia (approved):

PR #247, opened 2021-03-23
Add AppMesh metric integration test

bjrara:
**Why do we need it?**
Add container insight integration test of collecting AppMesh metrics on EKS

### DIFF
--- a/terraform/templates/defaults/validator_docker_compose.tpl
+++ b/terraform/templates/defaults/validator_docker_compose.tpl
@@ -26,6 +26,7 @@ services:
       - "ecsTaskDefFamily=${ecs_taskdef_family}"
       - "--ecs-context"
       - "ecsTaskDefVersion=${ecs_taskdef_version}"
+      - "--cloudwatch-context=${cloudwatch_context_json}"
       - "--alarm-names=${cpu_alarm}"
       - "--alarm-names=${mem_alarm}"
       - "--alarm-names=${incoming_packets_alarm}"
--- a/validator/src/main/java/com/amazon/aoc/services/CloudWatchService.java
+++ b/validator/src/main/java/com/amazon/aoc/services/CloudWatchService.java
@@ -19,18 +19,29 @@ import com.amazonaws.services.cloudwatch.AmazonCloudWatch;
 import com.amazonaws.services.cloudwatch.AmazonCloudWatchClientBuilder;
 import com.amazonaws.services.cloudwatch.model.Datapoint;
 import com.amazonaws.services.cloudwatch.model.Dimension;
+import com.amazonaws.services.cloudwatch.model.GetMetricDataRequest;
+import com.amazonaws.services.cloudwatch.model.GetMetricDataResult;
 import com.amazonaws.services.cloudwatch.model.GetMetricStatisticsRequest;
 import com.amazonaws.services.cloudwatch.model.ListMetricsRequest;
 import com.amazonaws.services.cloudwatch.model.Metric;
+import com.amazonaws.services.cloudwatch.model.MetricDataQuery;
+import com.amazonaws.services.cloudwatch.model.MetricDataResult;
 import com.amazonaws.services.cloudwatch.model.MetricDatum;
+import com.amazonaws.services.cloudwatch.model.MetricStat;
 import com.amazonaws.services.cloudwatch.model.PutMetricDataRequest;
 import com.amazonaws.services.cloudwatch.model.PutMetricDataResult;
 import com.amazonaws.services.cloudwatch.model.StandardUnit;
 
+import java.util.Date;
 import java.util.List;
 
-/** a wrapper of cloudwatch client. */
+/**
+ * a wrapper of cloudwatch client.
+ */
 public class CloudWatchService {
+  private static final int MAX_QUERY_PERIOD = 60;
+  private static final String REQUESTER = "integrationTest";
+
   AmazonCloudWatch amazonCloudWatch;
 
   /**
@@ -53,6 +64,25 @@ public class CloudWatchService {
     final ListMetricsRequest listMetricsRequest =
         new ListMetricsRequest().withNamespace(namespace).withMetricName(metricName);
     return amazonCloudWatch.listMetrics(listMetricsRequest).getMetrics();
+  }
+
+
+  /**
+   * getMetricData fetches the history data of the given metric from CloudWatch.
+   * @param metric metric query object
+   * @param startTime the start timestamp
+   * @param endTime the end timestamp
+   * @return List of MetricDataResult
+   */
+  public List<MetricDataResult> getMetricData(Metric metric, Date startTime, Date endTime) {
+    MetricStat stat = new MetricStat().withMetric(metric).withStat("Average")
+            .withPeriod(MAX_QUERY_PERIOD);
+    MetricDataQuery query = new MetricDataQuery().withMetricStat(stat).withId(REQUESTER);
+    final GetMetricDataRequest request = new GetMetricDataRequest().withMetricDataQueries(query)
+            .withStartTime(startTime).withEndTime(endTime);
+
+    GetMetricDataResult result = amazonCloudWatch.getMetricData(request);
+    return result.getMetricDataResults();
   }
 
   /**
--- a/validator/src/main/java/com/amazon/aoc/validators/ContainerInsightMetricsValidator.java
+++ b/validator/src/main/java/com/amazon/aoc/validators/ContainerInsightMetricsValidator.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.aoc.validators;
+
+import com.amazon.aoc.models.CloudWatchContext;
+import com.amazon.aoc.callers.ICaller;
+import com.amazon.aoc.exception.BaseException;
+import com.amazon.aoc.exception.ExceptionCode;
+import com.amazon.aoc.fileconfigs.FileConfig;
+import com.amazon.aoc.helpers.MustacheHelper;
+import com.amazon.aoc.helpers.RetryHelper;
+import com.amazon.aoc.models.Context;
+import com.amazon.aoc.models.ValidationConfig;
+import com.amazon.aoc.services.CloudWatchService;
+import com.amazonaws.services.cloudwatch.model.Dimension;
+import com.amazonaws.services.cloudwatch.model.Metric;
+import com.amazonaws.services.cloudwatch.model.MetricDataResult;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import lombok.extern.log4j.Log4j2;
+
+import java.nio.charset.StandardCharsets;
+import java.sql.Date;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * ContainerInsightMetricsValidator is diverged from CWMetricValidator for container insight
+ * metrics validation. In this validator, we exclude specific dimension checking, e.g. OTelLib
+ * and metrics rolling up.
+ */
+// TODO: We hope to merge it with CWMetricValidator when the validation process is finalized.
+@Log4j2
+public class ContainerInsightMetricsValidator implements IValidator {
+  private CloudWatchService cloudWatchService;
+  private final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+
+  private List<Metric> expectedMetrics;
+  private Set<String> validateNamespaces;
+
+  private static final int MAX_RETRY_COUNT = 6;
+  private static final int CHECK_INTERVAL_IN_MILLI = 30 * 1000;
+  private static final int CHECK_DURATION_IN_SECONDS = 2 * 60;
+
+
+  @Override
+  public void init(Context context, ValidationConfig validationConfig, ICaller caller,
+                   FileConfig expectedDataTemplate) throws Exception {
+    cloudWatchService = new CloudWatchService(context.getRegion());
+    validateNamespaces = getNamespacesToValidate(context.getCloudWatchContext());
+
+    MustacheHelper mustacheHelper = new MustacheHelper();
+    String templateInput = mustacheHelper.render(expectedDataTemplate, context);
+    expectedMetrics = mapper.readValue(templateInput.getBytes(StandardCharsets.UTF_8),
+          new TypeReference<List<Metric>>() {
+          });
+  }
+
+  @Override
+  public void validate() throws Exception {
+    log.info("[ContainerInsight] start validating metrics, pause 60s for metric collection");
+    TimeUnit.SECONDS.sleep(60);
+    log.info("[ContainerInsight] resume validation");
+
+    RetryHelper.retry(MAX_RETRY_COUNT, CHECK_INTERVAL_IN_MILLI, true, () -> {
+      Instant startTime = Instant.now().minusSeconds(CHECK_DURATION_IN_SECONDS)
+              .truncatedTo(ChronoUnit.MINUTES);
+      Instant endTime = startTime.plusSeconds(CHECK_DURATION_IN_SECONDS);
+
+      for (Metric expectedMetric : expectedMetrics) {
+        String namespace = extractNamespaceDimension(expectedMetric);
+        if (validateNamespaces.contains(namespace)) {
+          List<MetricDataResult> batchResult = cloudWatchService.getMetricData(expectedMetric,
+                  Date.from(startTime), Date.from(endTime));
+          boolean found = false;
+          for (MetricDataResult result : batchResult) {
+            if (result.getValues().size() > 0) {
+              found = true;
+            }
+          }
+          if (!found) {
+            throw new BaseException(ExceptionCode.EXPECTED_METRIC_NOT_FOUND,
+                    String.format("[ContainerInsight] metric %s not found under namespace %s",
+                            expectedMetric.getMetricName(), namespace));
+          }
+        }
+      }
+    });
+    log.info("[ContainerInsight] finish validation successfully");
+  }
+
+  private static Set<String> getNamespacesToValidate(CloudWatchContext cwContext) {
+    Set<String> namespaces = new HashSet<>();
+    if (cwContext.getAppMesh() != null) {
+      namespaces.add(cwContext.getAppMesh().getNamespace());
+    }
+    if (cwContext.getNginx() != null) {
+      namespaces.add(cwContext.getNginx().getNamespace());
+    }
+    if (cwContext.getHaproxy() != null) {
+      namespaces.add(cwContext.getHaproxy().getNamespace());
+    }
+    if (cwContext.getJmx() != null) {
+      namespaces.add(cwContext.getJmx().getNamespace());
+    }
+    if (cwContext.getMemcached() != null) {
+      namespaces.add(cwContext.getMemcached().getNamespace());
+    }
+    return namespaces;
+  }
+
+  private static String extractNamespaceDimension(Metric metric) {
+    for (Dimension dimension : metric.getDimensions()) {
+      if ("Namespace".equals(dimension.getName())) {
+        return dimension.getValue();
+      }
+    }
+    return null;
+  }
+}
--- a/terraform/containerinsight-eks-prometheus/appmesh/appmesh.tf
+++ b/terraform/containerinsight-eks-prometheus/appmesh/appmesh.tf
@@ -1,0 +1,186 @@
+# ------------------------------------------------------------------------
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+variable "provider_url" {
+  type    = string
+  default = ""
+}
+
+variable "sample_app_image_repo" {
+  type    = string
+  default = null
+}
+
+variable "testing_id" {
+  type    = string
+  default = ""
+}
+
+variable "region" {
+  type    = string
+  default = "us-west-2"
+}
+
+variable "testcase" {
+  type    = string
+  default = "../testcases/container_insight"
+}
+
+variable "kubeconfig" {
+  type = string
+  default = "kubeconfig"
+}
+
+output "metric_dimension_namespace" {
+  value = kubernetes_namespace.traffic_ns.metadata[0].name
+}
+
+data "aws_caller_identity" "current" {}
+
+locals {
+  mesh_name      = "howto-k8s-http-headers-${var.testing_id}"
+  aws_account_id = data.aws_caller_identity.current.account_id
+}
+
+data "aws_iam_policy" "appmesh_policy" {
+  arn = "arn:aws:iam::${local.aws_account_id}:policy/AWSAppMeshK8sControllerIAMPolicy"
+}
+
+module "iam_assumable_role_with_oidc" {
+  source       = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
+  version      = "~> 3.0"
+  provider_url = var.provider_url
+
+  create_role = true
+  role_name   = "AppMeshControllerRole-${var.testing_id}"
+  role_policy_arns = [
+  data.aws_iam_policy.appmesh_policy.arn]
+
+  oidc_fully_qualified_subjects = [
+  "system:serviceaccount:${kubernetes_namespace.appmesh_ns.metadata[0].name}:appmesh-controller"]
+  number_of_role_policy_arns = 1
+}
+
+resource "kubernetes_namespace" "appmesh_ns" {
+  metadata {
+    name = "appmesh-system-${var.testing_id}"
+  }
+}
+
+resource "kubernetes_service_account" "appmesh_sa" {
+  metadata {
+    name      = "appmesh-controller"
+    namespace = kubernetes_namespace.appmesh_ns.metadata[0].name
+    annotations = {
+      "eks.amazonaws.com/role-arn" = module.iam_assumable_role_with_oidc.this_iam_role_arn
+    }
+  }
+  automount_service_account_token = true
+}
+
+resource "helm_release" "eks" {
+  name      = "eks"
+  namespace = kubernetes_namespace.appmesh_ns.metadata[0].name
+
+  repository = "https://aws.github.io/eks-charts"
+  chart      = "appmesh-controller"
+  version    = "1.3.0"
+
+  set {
+    name  = "serviceAccount.create"
+    value = "false"
+  }
+
+  set {
+    name  = "serviceAccount.name"
+    value = kubernetes_service_account.appmesh_sa.metadata[0].name
+  }
+
+  set {
+    name  = "region"
+    value = var.region
+  }
+
+  provisioner "local-exec" {
+    command = "kubectl --kubeconfig=${var.kubeconfig} apply -k \"github.com/aws/eks-charts/stable/appmesh-controller/crds?ref=v0.0.47\""
+  }
+
+  wait = true
+  depends_on = [
+  module.iam_assumable_role_with_oidc]
+}
+
+resource "kubernetes_namespace" "traffic_ns" {
+  metadata {
+    name = "howto-k8s-http-headers-${var.testing_id}"
+    labels = {
+      "mesh"                                   = local.mesh_name
+      "appmesh.k8s.aws/sidecarInjectorWebhook" = "enabled"
+    }
+  }
+  depends_on = [
+  null_resource.delete_mesh]
+}
+
+resource "null_resource" "delete_mesh" {
+  triggers = {
+    mesh_name = local.mesh_name
+    kubeconfig = var.kubeconfig
+  }
+  provisioner "local-exec" {
+    when    = destroy
+    command = "kubectl --kubeconfig=${self.triggers.kubeconfig} delete mesh ${self.triggers.mesh_name} --ignore-not-found=true"
+  }
+  depends_on = [
+  helm_release.eks]
+}
+
+resource "null_resource" "appmesh_readiness_check" {
+  provisioner "local-exec" {
+    when    = create
+    command = "kubectl --kubeconfig=${var.kubeconfig} rollout status deployment eks-appmesh-controller -n${kubernetes_namespace.appmesh_ns.metadata[0].name}"
+  }
+  depends_on = [
+  helm_release.eks]
+}
+
+data "template_file" "traffic_deployment_file" {
+  template = file("${var.testcase}/appmesh_traffic_sample.tpl")
+  vars = {
+    APP_NAMESPACE   = kubernetes_namespace.traffic_ns.metadata[0].name
+    MESH_NAME       = local.mesh_name
+    FRONT_APP_IMAGE = "${var.sample_app_image_repo}:feapp"
+    COLOR_APP_IMAGE = "${var.sample_app_image_repo}:colorapp"
+  }
+
+  depends_on = [
+  null_resource.appmesh_readiness_check]
+}
+
+resource "local_file" "appmesh_deployment" {
+  filename = "appmesh_deployment_${var.testing_id}.yaml"
+  content  = data.template_file.traffic_deployment_file.rendered
+  depends_on = [
+  helm_release.eks]
+}
+
+resource "null_resource" "traffic_deployment" {
+  triggers = {
+    config_contents = md5(local_file.appmesh_deployment.content)
+  }
+  provisioner "local-exec" {
+    command = "kubectl --kubeconfig=${var.kubeconfig} apply -f ${local_file.appmesh_deployment.filename}"
+  }
+}
--- a/terraform/containerinsight-eks-prometheus/common.auto.tf
+++ b/terraform/containerinsight-eks-prometheus/common.auto.tf
@@ -1,0 +1,1 @@
+../common.tf
--- a/terraform/containerinsight-eks-prometheus/kubeconfig.tpl
+++ b/terraform/containerinsight-eks-prometheus/kubeconfig.tpl
@@ -1,0 +1,18 @@
+apiVersion: v1
+clusters:
+- cluster:
+    certificate-authority-data: ${CA_DATA}
+    server: ${SERVER_ENDPOINT}
+  name: testing_cluster
+contexts:
+- context:
+    cluster: testing_cluster
+    user: terraform_user
+  name: integ_test
+current-context: integ_test
+kind: Config
+preferences: {}
+users:
+- name: terraform_user
+  user:
+    token: ${TOKEN}
--- a/terraform/containerinsight-eks-prometheus/main.tf
+++ b/terraform/containerinsight-eks-prometheus/main.tf
@@ -1,0 +1,189 @@
+# ------------------------------------------------------------------------
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+# creating a eks cluster takes around 10 minutes typically.
+# so in the eks/k8s test, we need tester to provide the cluster instead of creating it in terraform
+# so that we can shorten the execution time
+
+module "common" {
+  source = "../common"
+
+  aoc_image_repo = var.aoc_image_repo
+  aoc_version = var.aoc_version
+}
+
+# region
+provider "aws" {
+  region = var.region
+}
+
+# get eks cluster by name
+data "aws_eks_cluster" "testing_cluster" {
+  name = var.eks_cluster_name
+}
+
+data "aws_eks_cluster_auth" "testing_cluster" {
+  name = var.eks_cluster_name
+}
+
+provider "kubernetes" {
+  // Note: copy from eks module. Please avoid use shorted-lived tokens when running locally.
+  // For more information: https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#exec-plugins
+  host = data.aws_eks_cluster.testing_cluster.endpoint
+  cluster_ca_certificate = base64decode(data.aws_eks_cluster.testing_cluster.certificate_authority[0].data)
+  token = data.aws_eks_cluster_auth.testing_cluster.token
+  load_config_file = false
+}
+
+data "template_file" "kubeconfig_file" {
+  template = file("./kubeconfig.tpl")
+  vars = {
+    CA_DATA: data.aws_eks_cluster.testing_cluster.certificate_authority[0].data
+    SERVER_ENDPOINT: data.aws_eks_cluster.testing_cluster.endpoint
+    TOKEN = data.aws_eks_cluster_auth.testing_cluster.token
+  }
+}
+
+resource "local_file" "kubeconfig" {
+  filename = "kubeconfig"
+  content = data.template_file.kubeconfig_file.rendered
+}
+
+provider "helm" {
+  kubernetes {
+    host = data.aws_eks_cluster.testing_cluster.endpoint
+    cluster_ca_certificate = base64decode(data.aws_eks_cluster.testing_cluster.certificate_authority[0].data)
+    token = data.aws_eks_cluster_auth.testing_cluster.token
+  }
+}
+
+# create a unique namespace for each run
+resource "kubernetes_namespace" "aoc_ns" {
+  metadata {
+    name = "aoc-ns-${module.common.testing_id}"
+  }
+}
+
+resource "kubernetes_service_account" "aoc_role" {
+  metadata {
+    name = "aoc-role-${module.common.testing_id}"
+    namespace = kubernetes_namespace.aoc_ns.metadata[0].name
+  }
+  automount_service_account_token = true
+}
+
+resource "kubernetes_cluster_role_binding" "aoc_role_binding" {
+  metadata {
+    name = "aoc-role-binding-${module.common.testing_id}"
+  }
+  role_ref {
+    api_group = "rbac.authorization.k8s.io"
+    kind = "ClusterRole"
+    name = "cluster-admin"
+  }
+  subject {
+    kind = "ServiceAccount"
+    name = "aoc-role-${module.common.testing_id}"
+    namespace = kubernetes_namespace.aoc_ns.metadata[0].name
+  }
+}
+resource "kubernetes_deployment" "standalone_aoc_deployment" {
+  metadata {
+    name = "aoc"
+    namespace = kubernetes_namespace.aoc_ns.metadata[0].name
+    labels = {
+      app = "aoc"
+    }
+  }
+
+  spec {
+    replicas = 1
+
+    selector {
+      match_labels = {
+        app = "aoc"
+      }
+    }
+
+    template {
+      metadata {
+        labels = {
+          app = "aoc"
+        }
+      }
+
+      spec {
+        service_account_name = "aoc-role-${module.common.testing_id}"
+        automount_service_account_token = true
+        # aoc
+        container {
+          name = "aoc"
+          image = module.common.aoc_image
+          image_pull_policy = "Always"
+          args = [
+            "--config",
+            "/etc/eks/prometheus/config-all.yaml"]
+          env {
+            name = "AWS_REGION"
+            value = var.region
+          }
+          env {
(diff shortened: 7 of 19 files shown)


terraform/templates/local/docker_compose_from_source.tpl
```diff
@@ -9,7 +9,7 @@ services:
 
   aws-ot-collector:
     build:
-      context: ../../../aws-otel-collector
+      context: $GOPATH/src/github.com/aws-observability/aws-otel-collector
```
pingleig: why change this? If you clone the test in the gopath way, the relative path should work.

pingleig (commented):
I didn't see [aws_appmesh_mesh](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/appmesh_mesh) so you created the mesh without terraform? You might want to add the instruction here.

terraform/templates/local/docker_compose_from_source.tpl
bjrara: Why not? `$GOPATH` is more accurate than relative path.

bjrara:
> I didn't see [aws_appmesh_mesh](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/appmesh_mesh) so you created the mesh without terraform? You might want to add the instruction here.

I don't create appmesh using terraform. All the test resources are created using a single yaml.
https://github.com/aws-observability/aws-otel-test-framework/blob/828a85e6a47c535554fe48bacbb7e230e60a174d/terraform/testcases/container_insight_appmesh/appmesh_traffic_sample.tpl#L1-L8

terraform/eks-cloudwatch/app_mesh.tf
```diff
@@ -0,0 +1,156 @@
+# ------------------------------------------------------------------------
```
wyTrivail: may i know the reason why we don't do the logic in `eks` folder? the folder name `eks-cloudwatch` seems not verbose enough? hard to guess its usage

terraform/eks-cloudwatch/app_mesh.tf
```diff
@@ -0,0 +1,156 @@
...
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+# creating a eks cluster takes around 10 minutes typically.
+# so in the eks/k8s test, we need tester to provide the cluster instead of creating it in terraform
+# so that we can shorten the execution time
+
+locals {
+  aws_account_id = var.aws_account_id != "" ? var.aws_account_id : data.aws_caller_identity.current.account_id
```
wyTrivail: if we configure a different account_id, would it be a cross-account call?

terraform/eks-cloudwatch/app_mesh.tf
```diff
@@ -0,0 +1,156 @@
...
+  arn = "arn:aws:iam::${local.aws_account_id}:policy/AWSAppMeshK8sControllerIAMPolicy"
+}
+
+module "iam_assumable_role_with_oidc" {
+  source = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
+  version = "~> 3.0"
+  provider_url = data.aws_eks_cluster.testing_cluster.identity[0].oidc[0].issuer
+
+  create_role = true
+  role_name = "appmesh-controller_role"
```
wyTrivail: should we use underscore or hyphens?

terraform/eks-cloudwatch/app_mesh.tf
```diff
@@ -0,0 +1,156 @@
...
+data "aws_iam_policy" "appmesh_policy" {
+  arn = "arn:aws:iam::${local.aws_account_id}:policy/AWSAppMeshK8sControllerIAMPolicy"
+}
+
+module "iam_assumable_role_with_oidc" {
+  source = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
+  version = "~> 3.0"
+  provider_url = data.aws_eks_cluster.testing_cluster.identity[0].oidc[0].issuer
+
+  create_role = true
```
wyTrivail: since the role name is fixed, when we run multiple tests in one aws account, any conflict would be expected?

terraform/eks-cloudwatch/app_mesh.tf
```diff
@@ -0,0 +1,156 @@
...
+    annotations = {
+      "eks.amazonaws.com/role-arn" = module.iam_assumable_role_with_oidc.this_iam_role_arn
+    }
+  }
+  automount_service_account_token = true
+}
+
+provider "helm" {
+  kubernetes {
+    config_path = "~/.kube/config"
```
wyTrivail: do we make an assumption that "~/.kube/config" is provided before we run the test? or i guess it's generated during the test?

terraform/eks-cloudwatch/app_mesh.tf
```diff
@@ -0,0 +1,156 @@
...
+  }
+}
+
+resource "helm_release" "eks" {
+  name = "eks"
+  namespace = kubernetes_namespace.appmesh_ns.metadata[0].name
+
+  repository = "https://aws.github.io/eks-charts"
+  chart = "appmesh-controller"
+  version = "1.3.0"
```
wyTrivail: why 1.3.0?

terraform/eks-cloudwatch/app_mesh.tf
```diff
@@ -0,0 +1,156 @@
...
+    when = destroy
+    command = "kubectl delete mesh howto-k8s-http-headers --ignore-not-found=true"
+  }
+  depends_on = [helm_release.eks]
+}
+
+resource "null_resource" "appmesh_readiness_check" {
+  provisioner "local-exec" {
+    when = create
+    command = "kubectl rollout status deployment eks-appmesh-controller -n${kubernetes_namespace.appmesh_ns.metadata[0].name}"
```
wyTrivail: it's the only way to deploy it? any possibility to use terraform?

terraform/eks-cloudwatch/main.tf
```diff
@@ -0,0 +1,158 @@
...
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+# creating a eks cluster takes around 10 minutes typically.
+# so in the eks/k8s test, we need tester to provide the cluster instead of creating it in terraform
+# so that we can shorten the execution time
+
+variable "aws_account_id" {
```
wyTrivail: better define vars in a separate file

terraform/eks-cloudwatch/main.tf
```diff
@@ -0,0 +1,158 @@
...
+}
+
+# set up kubectl
+provider "kubernetes" {
+  config_path = "~/.kube/config"
+  version = "~> 1.13"
+}
+
+# create a unique namespace for each run
+resource "kubernetes_namespace" "aoc_ns" {
```
wyTrivail: would that be possible we can somehow extract the existing `eks` module a bit so that we can reuse some of the common things?

terraform/eks-cloudwatch/main.tf
```diff
@@ -0,0 +1,158 @@
...
+        service_account_name = "aoc-role-${module.common.testing_id}"
+        automount_service_account_token = true
+        # aoc
+        container {
+          name = "aoc"
+          image = module.common.aoc_image
+          image_pull_policy = "Always"
+          args = [
+            "--config",
+            "/etc/eks/prometheus/config-all.yaml"]
```
wyTrivail: i might missed this part, who generates this config?

terraform/eks-cloudwatch/variables.tf
```diff
@@ -0,0 +1,22 @@
...
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+variable "eks_cluster_name" {
+  default = "aws-otel-testing-framework-eks"
+}
+
+variable "mock_endpoint" {
+  default = "localhost/put-data"
```
wyTrivail: would it be a common var in the common.tf?

terraform/templates/local/docker_compose_from_source.tpl
wyTrivail: some people don't really have gopath there, the assumption we are making here is testing-framework code is at the same folder of collector code.  if we want to change it, please change the document also.

validator/src/main/java/com/amazon/aoc/MetricContext.java
```diff
@@ -0,0 +1,9 @@
+package com.amazon.aoc;
+
+import lombok.Data;
+
+@Data
+public class MetricContext {
```
wyTrivail: since this validator aims to support multiple component, not just cloudwatch, maybe giving it a concrete name would be better?

validator/src/main/java/com/amazon/aoc/services/CloudWatchService.java
```diff
@@ -45,22 +40,40 @@ public CloudWatchService(String region) {
...
+  /**
+   * getMetricData fetches the history data of the given metric from CloudWatch.
+   * @param metric metric query object
+   * @param startTime the start timestamp
+   * @param endTime the end timestamp
+   * @return List of MetricDataResult
+   */
+  public List<MetricDataResult> getMetricData(Metric metric, Date startTime, Date endTime) {
+    MetricStat stat = new MetricStat().withMetric(metric).withStat("Average").withPeriod(60);
+    MetricDataQuery query = new MetricDataQuery().withMetricStat(stat).withId("integrationTest");
```
wyTrivail: small nits, it would be better if we can define magic code `60`, `integrationTest` as vars.

validator/src/main/java/com/amazon/aoc/services/CloudWatchService.java
```diff
@@ -45,22 +40,40 @@ public CloudWatchService(String region) {
...
+
+    GetMetricDataResult result = amazonCloudWatch.getMetricData(request);
+    return result.getMetricDataResults();
+  }
+
   /**
    * putMetricData publish metric to CloudWatch.
    *
-   * @param namespace the metric namespace on CloudWatch
+   * @param namespace  the metric namespace on CloudWatch
```
wyTrivail: additonal spaces?

validator/src/main/java/com/amazon/aoc/validators/ContainerInsightMetricsValidator.java
```diff
@@ -0,0 +1,76 @@
+package com.amazon.aoc.validators;
```
wyTrivail, line 16: i guess we forgot to put license header in validator sometimes :)

validator/src/main/java/com/amazon/aoc/validators/ContainerInsightMetricsValidator.java
```diff
@@ -0,0 +1,76 @@
...
+    cloudWatchService = new CloudWatchService(context.getRegion());
+    MustacheHelper mustacheHelper = new MustacheHelper();
+    String templateInput = mustacheHelper.render(expectedDataTemplate, context);
+    expectedMetrics = mapper.readValue(templateInput.getBytes(StandardCharsets.UTF_8),
+          new TypeReference<List<Metric>>() {
+          });
+  }
+
+  @Override
+  public void validate() throws Exception {
```
wyTrivail, line 78: what's the difference between appmesh validation and the normal cloudwatch metric validation? seems we just do a simple check on it? is it intended or we had a "todo" for it?

wyTrivail (commented):
just some small nits, thanks

terraform/eks-cloudwatch/app_mesh.tf
bjrara: * EKS seems to have its own testcases. The current folder `eks` is too general yet serves for a very specific testing purpose. For instance, it requests user to set up cortex in the account, which is of no interest when testing container insight features. Vice versa, the framework should not ask users to install AppMesh when testing EKS merely.
* The new `eks-cloudwatch` uses a separate container insight configuration rather than the default one, which conflicts with the one that eks should use. Related PR: https://github.com/aws-observability/aws-otel-collector/blob/92414f1de2fba3a9c846922304b4bce81252a952/config/eks/prometheus/config-all.yaml

So I don't feel it's a good idea to couple eks and the test cases I'm adding. Perhaps `eks-containerinsight` is more self-explained? btw the name should be able to split `ecs` and `eks` cuz we would have container insight test cases on ECS soon as well.

validator/src/main/java/com/amazon/aoc/services/CloudWatchService.java
bjrara: It aligns with the next param spec.

terraform/eks-cloudwatch/app_mesh.tf
bjrara: Removed it from the variables.

terraform/templates/local/docker_compose_from_source.tpl
bjrara: Thanks for the explanation. Reverted.

terraform/eks-cloudwatch/app_mesh.tf
bjrara: Changed the format to `AppMeshControllerRole-${var.testing_id}`

terraform/eks-cloudwatch/app_mesh.tf
bjrara: https://github.com/aws-observability/aws-otel-test-framework/pull/247#discussion_r601029859

terraform/eks-cloudwatch/app_mesh.tf
bjrara: To finalize the version of AppMesh to avoid version conflict

terraform/eks-cloudwatch/app_mesh.tf
bjrara: The AppMesh controller is deployed through helm. `appmesh_readiness_check` waits the controller to be ready, otherwise the injection of envoy sidecar would fail for the following traffic application deployment.

terraform/eks-cloudwatch/main.tf
bjrara: Removed.

terraform/eks-cloudwatch/main.tf
bjrara: https://github.com/aws-observability/aws-otel-test-framework/pull/247#discussion_r600981387

Let me know if you have other suggestions.

terraform/eks-cloudwatch/main.tf
bjrara: Sorry about the confusion. This integration test will be introduced in the PR: https://github.com/aws-observability/aws-otel-collector/pull/394

terraform/eks-cloudwatch/variables.tf
bjrara: Code cleansing done. Nice catch!

validator/src/main/java/com/amazon/aoc/MetricContext.java
bjrara: Changed to `CloudWatchContext`

validator/src/main/java/com/amazon/aoc/services/CloudWatchService.java
bjrara: Removed unrelated format.

validator/src/main/java/com/amazon/aoc/validators/ContainerInsightMetricsValidator.java
bjrara: Thanks for the reminder!

validator/src/main/java/com/amazon/aoc/validators/ContainerInsightMetricsValidator.java
bjrara: Yes, I guess we need a discussion on how to merge the two validators. There're some presumption that breaks container insight validations in the original `CWMetricValidator`.

terraform/eks-cloudwatch/app_mesh.tf
bjrara: Thanks for pointing it out. Updated.

wyTrivail:
please let me know when you are ready for review the code :)

bjrara:
> please let me know when you are ready for review the code :)

Thanks for keeping track of this. Please help review.

bjrara:
> please let me know when you are ready for review the code :)

Thanks for keeping track of the PR. Please review again.

terraform/eks-cloudwatch/appmesh/appmesh.tf
```diff
@@ -0,0 +1,186 @@
...
+  aws_account_id = data.aws_caller_identity.current.account_id
+}
+
+data "aws_iam_policy" "appmesh_policy" {
+  arn = "arn:aws:iam::${local.aws_account_id}:policy/AWSAppMeshK8sControllerIAMPolicy"
+}
+
+module "iam_assumable_role_with_oidc" {
+  source       = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
+  version      = "v3.13.0"
```
mxiamxia: should we use version constraint to get the latest version? eg `"~> 3.0"`

terraform/eks-cloudwatch/main.tf
mxiamxia: Q: should we use `config-all.yaml`? Or make it as variable and each test case (eg, appmesh, nginx. etc) only apply their own configurations?

terraform/templates/defaults/validator_docker_compose.tpl
```diff
@@ -26,6 +26,7 @@ services:
       - "ecsTaskDefFamily=${ecs_taskdef_family}"
       - "--ecs-context"
       - "ecsTaskDefVersion=${ecs_taskdef_version}"
+      - --cloudwatch-context=${metric_dimension_json}
```
mxiamxia: nit: pls add `" "` to match code style

validator/src/main/java/com/amazon/aoc/services/CloudWatchService.java
```diff
@@ -17,19 +17,14 @@
...
-import com.amazonaws.services.cloudwatch.model.Datapoint;
-import com.amazonaws.services.cloudwatch.model.Dimension;
-import com.amazonaws.services.cloudwatch.model.GetMetricStatisticsRequest;
-import com.amazonaws.services.cloudwatch.model.ListMetricsRequest;
-import com.amazonaws.services.cloudwatch.model.Metric;
-import com.amazonaws.services.cloudwatch.model.MetricDatum;
-import com.amazonaws.services.cloudwatch.model.PutMetricDataRequest;
-import com.amazonaws.services.cloudwatch.model.PutMetricDataResult;
-import com.amazonaws.services.cloudwatch.model.StandardUnit;
+import com.amazonaws.services.cloudwatch.model.*;
```
mxiamxia: we should avoid wildcard imports

terraform/eks-cloudwatch/main.tf
bjrara: In the final PR of this integration test series, we test all the workload all at once. Metrics and logs are separated by namespaces, testing workload one by one makes no difference in validation with testing in a batch. However, testing separately surely will make integration test more expensive because of the extra effort on building and compiling and initiating new containers for testing. Also test order should be taken care of, which is what terraform is poor at. 

CloudWatch agent doesn't even provide separated config for each workload IIUC. I'm still questioning myself whether those separated configs are valuable.

terraform/eks-cloudwatch/appmesh/appmesh.tf
bjrara: Done

terraform/templates/defaults/validator_docker_compose.tpl
bjrara: Done

validator/src/main/java/com/amazon/aoc/services/CloudWatchService.java
bjrara: Done

validator/src/main/java/com/amazon/aoc/MetricContext.java
wyTrivail: would this context include "log" and "metric"? or just metric?

validator/src/main/java/com/amazon/aoc/validators/ContainerInsightMetricsValidator.java
wyTrivail: so, what's the path forward now? we're gonna use this simple validation instead of making the validator more generic? i'd prefer we do the latter, but if we have to do so, let's add a "todo" here

terraform/eks-cloudwatch/appmesh/appmesh.tf
```diff
@@ -0,0 +1,186 @@
...
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+variable "provider_url" {
```
wyTrivail: better put variables in variables.tf

terraform/eks-cloudwatch/appmesh/appmesh.tf
```diff
@@ -0,0 +1,186 @@
...
+  type    = string
+  default = ""
+}
+
+variable "sample_app_image_repo" {
+  type    = string
```
wyTrivail: i wonder do we need to specify type here as string is the default type?

wyTrivail: and i remember this is a var in common.tf, why's the reason to create one here?

terraform/eks-cloudwatch/appmesh/appmesh.tf
```diff
@@ -0,0 +1,186 @@
...
+  default = null
+}
+
+variable "testing_id" {
+  type    = string
+  default = ""
+}
+
+variable "region" {
```
wyTrivail: looks like many vars are duplicate with the ones in common.tf?

terraform/eks-cloudwatch/appmesh/appmesh.tf
```diff
@@ -0,0 +1,186 @@
...
+  type    = string
+  default = "../testcases/container_insight"
+}
+
+variable "kubeconfig" {
+  type = string
+  default = "kubeconfig"
+}
+
+output "metric_dimension_namespace" {
```
wyTrivail: we use it for debugging?

validator/src/main/java/com/amazon/aoc/services/CloudWatchService.java
bjrara: Done

validator/src/main/java/com/amazon/aoc/validators/ContainerInsightMetricsValidator.java
pingleig: I prefer we add todo and deal with it later, this PR is already pretty big and we are not enabling it until the related features merge into adot repo. We have plenty time to refactor those validators, both me and @pxaws will use cw metrics and log validators for other container insight related features and we can figure out the generic way along the way.

terraform/eks-cloudwatch/app_mesh.tf
pingleig: I think `containerinsight-eks-prometheus` should be good and I can do `containerinsight-ecs-prometheus`, cause they are just setting up container insight auto dashboard specific workloads. btw: there is little to share between eks and ecs containersight when setting up environment so we don't need a common module for containerinsight.

terraform/eks-cloudwatch/variables.tf
```diff
@@ -0,0 +1,22 @@
...
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+variable "eks_cluster_name" {
+  default = "aws-otel-testing-framework-eks"
+}
+
+variable "sample_app_image_repo" {
+  default = "611364707713.dkr.ecr.us-west-2.amazonaws.com/otel-test/container-insight-samples"
```
wyTrivail: where's the source code of this sample app? i wonder if we could put it under the sample-apps folder so that we can manage it?

terraform/templates/defaults/validator_docker_compose.tpl
```diff
@@ -26,6 +26,7 @@ services:
       - "ecsTaskDefFamily=${ecs_taskdef_family}"
       - "--ecs-context"
       - "ecsTaskDefVersion=${ecs_taskdef_version}"
+      - "--cloudwatch-context=${metric_dimension_json}"
```
wyTrivail: are these dimension values from eks?

terraform/testcases/container_insight/parameters.tfvars
```diff
@@ -0,0 +1,2 @@
+# this file is defined in validator/src/main/resources/validations
+validation_config="eks-container-insight.yml"
```
wyTrivail: i guess this folder is just for eks container insight but not for ecs?

validator/src/main/java/com/amazon/aoc/MetricContext.java
bjrara: Good question. Prometheus metrics are the only one being tested. However these metrics are pushed to CloudWatch using structured log, which explains why we test metrics and logs both.

terraform/eks-cloudwatch/app_mesh.tf
bjrara: @wyTrivail Let me know if any of the proposed names sounds good to you.

wyTrivail (commented):
my major concern is still, we are creating a new folder for eks, i'm still thinking about how we can minimize the duplicate parts, for example:
1. we are using testing_id to create namespace,
2. we are deploying aoc with configmap.
3. we will need to support mocked server which is already supported in eks folder today.

the eks folder today can serve all the testcases,  `cortex` is just an optional parameter. this [file](https://github.com/aws-observability/aws-otel-collector/blob/main/e2etest/testcases.json#L4) lists all the testcases which could be running via eks folder. 

i'd like to ask myself if we want to have a separate folder for some testcases, 
1. if the difference is in ot config, we might be able to just configure it in testcase?
2. if the difference is on how to deploy otcollector, why can't we add a new "starting mode" into eks folder?
3. if the difference is on how to deploy sample apps, why can't we make the deployment of sample apps configurable? or add a file under eks folder?
4. if the difference is on validation, we might be able to just configure the parameters in testcase?
5. how hard would it be to make eks folder more generic?  

If we can have a plan around how do we maintain the separate folders, either we have solid reasons to keep them separate, or we have a vision to merge them someday, i will be okay with that.

terraform/eks-cloudwatch/app_mesh.tf
wyTrivail: containerinsight-eks-prometheus sounds good

wyTrivail: btw, let's don't forget to change this script: https://github.com/aws-observability/aws-otel-collector/blob/main/e2etest/get-testcases.py

terraform/eks-cloudwatch/appmesh/appmesh.tf
bjrara: To make sure we're on the same page, appmesh works as an independent stateless module, which only accepts input vars, and export variables that would be used by its caller. It self doesn't rely on any other module, that's why we don't put the variables in a common place.

terraform/eks-cloudwatch/appmesh/appmesh.tf
bjrara: https://github.com/aws-observability/aws-otel-test-framework/pull/247#discussion_r604486146

terraform/eks-cloudwatch/appmesh/appmesh.tf
bjrara: https://github.com/aws-observability/aws-otel-test-framework/pull/247#discussion_r604486146

terraform/eks-cloudwatch/appmesh/appmesh.tf
bjrara: https://github.com/aws-observability/aws-otel-test-framework/pull/247#discussion_r604486146

bjrara: It's referred in the main function: https://github.com/aws-observability/aws-otel-test-framework/pull/247/files#diff-2d154fdde544d49c61326e5b38c46e422cf1c06757fc916fe2edd5d33c7ed0f0R182

terraform/eks-cloudwatch/appmesh/appmesh.tf
bjrara: Also the variables in `common.tf` is poorly documented. I'm not sure how it is used and how it should be used. As you may see appmesh requires multiple app image, and apparently a single variable naming `sample_app_image` can't fulfill my needs.

terraform/eks-cloudwatch/main.tf
wyTrivail: i'm not sure if that's a good approach to bake all the testing configurations into image and release it to customer? for example, if we want to test with negative endpoint where we need to modify the endpoint value in config, how would we do that?

wyTrivail: my point is not about we should do it right away, but instead, we need to think about what's the overlapping parts we are making with the eks folder, how do we extract the common parts, if there's a contributor who wants to add a new testcase, how do we guide him to add.

wyTrivail (approved):
Ship it. i still have many questions which i hope we address them with opening an issue though.

terraform/eks-cloudwatch/variables.tf
bjrara: The source code can be found following our official setup guidance, it is maintained in another AWS repo: https://github.com/aws/aws-app-mesh-examples/tree/master/walkthroughs/howto-k8s-http-headers.
Ref:
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/ContainerInsights-Prometheus-Sample-Workloads-appmesh-EKS.html

I didn't put it under the sample-apps folder because I personally think:
1. It increases the effort to sync this copy with the original one.
2. I can't see any potential user that is interested in running the integration tests locally. If this image would only be used by aoc integration test, I don't see a strong reason to maintain it. What's the benefit?
3. We don't aim to test sample apps, they serve for testing aoc. A stable version of sample app is enough to meet the requirement.

terraform/eks-cloudwatch/appmesh/appmesh.tf
wyTrivail: maybe you want to add a comment above `sample_app_image_repo` to explain it's for multiple app images, which i was not aware of when i saw this code.

terraform/eks-cloudwatch/variables.tf
wyTrivail: 1. It increases the effort to sync this copy with the original one. make sense.
2. I can't see any potential user that is interested in running the integration tests locally. If this image would only be used by aoc integration test, I don't see a strong reason to maintain it. What's the benefit? -> who is maintaining this sample app now? does it have workflow for auto-build? how do we track the version?
3. We don't aim to test sample apps, they serve for testing aoc. A stable version of sample app is enough to meet the requirement. ->  would we need performance test?

wyTrivail: i wonder if this sample app is just for aoc integration-test, why don't we move it into this repo? i'm okay with that if this sample app has multiple usages than aoc inregration-test

terraform/eks-cloudwatch/main.tf
bjrara: Not sure if I understand your concern correctly, but if `the testing configurations` refer to the [prometheus container insight configuration](https://github.com/aws-observability/aws-otel-collector/blob/main/config/eks/prometheus/config-all.yaml), it's not TESTing configurations, but official configurations that end users could use when they switch to aoc for container insight hands-on. Same as what [cloudwatch agent prometheus-eks.yaml](https://github.com/aws/amazon-cloudwatch-agent/blob/master/amazon-cloudwatch-container-insights/k8s-deployment-manifest-templates/deployment-mode/service/cwagent-prometheus/prometheus-eks.yaml) is doing.

And I still remember the conversation we had on why not merge into eks foler: https://github.com/aws-observability/aws-otel-test-framework/pull/247#discussion_r600981387.

I, as a contributor, am confused what eks folder is served for. Maybe a document to illustrate the existing eks module would be greatly helpful for developers to follow.

terraform/eks-cloudwatch/variables.tf
bjrara: > who is maintaining this sample app now? does it have workflow for auto-build? 

I believe it's EKS team, checking the contributors. 
Per [README](https://github.com/aws/aws-app-mesh-examples/blob/master/walkthroughs/howto-k8s-http-headers/README.md), they've provided a deploy script that will build the image, and deploy the relating resources to the cluster (cleansing should still be taken cared of by users).

> how do we track the version?

Images are for sample apps, we don't need version tracking, or so I thought.

> would we need performance test?

I'd like to see it happen. We don't have concrete plan. Yet that's unrelated to sample app images IIUC.

> this sample app is just for aoc integration-test, why don't we move it into this repo? 

Apparently it's not for aoc integration-test, but is recommended by CloudWatch for AppMesh quick start.

mxiamxia: pls help to leave source code link in the comment.

terraform/testcases/container_insight/parameters.tfvars
bjrara: Renamed the folder

terraform/templates/defaults/validator_docker_compose.tpl
bjrara: I'm hoping containerinsight-ecs-prometheus would reuse this configuration too.

terraform/eks-cloudwatch/app_mesh.tf
bjrara: Renamed the folder.

terraform/eks-cloudwatch/variables.tf
bjrara: Added def and source code links to the variable.

bjrara:
> my major concern is still, we are creating a new folder for eks, i'm still thinking about how we can minimize the duplicate parts, for example:
> 
> 1. we are using testing_id to create namespace,
> 2. we are deploying aoc with configmap.
> 3. we will need to support mocked server which is already supported in eks folder today.
> 
> the eks folder today can serve all the testcases, `cortex` is just an optional parameter. this [file](https://github.com/aws-observability/aws-otel-collector/blob/main/e2etest/testcases.json#L4) lists all the testcases which could be running via eks folder.
> 
> i'd like to ask myself if we want to have a separate folder for some testcases,
> 
> 1. if the difference is in ot config, we might be able to just configure it in testcase?
> 2. if the difference is on how to deploy otcollector, why can't we add a new "starting mode" into eks folder?
> 3. if the difference is on how to deploy sample apps, why can't we make the deployment of sample apps configurable? or add a file under eks folder?
> 4. if the difference is on validation, we might be able to just configure the parameters in testcase?
> 5. how hard would it be to make eks folder more generic?
> 
> If we can have a plan around how do we maintain the separate folders, either we have solid reasons to keep them separate, or we have a vision to merge them someday, i will be okay with that.

Add some thoughts for people who are interested in this comment. Please note this is still an open question needs to be addressed later in the future.

**Correction:**
The otel configuration used in the integration test is _NOT_ a testing config but rather a config that would be released to users for container insight on EKS. It will be built in the image, accessible directly. We're _NOT_ deploying otel collector with configmaps. No extra setup process is required when deloying aoc in the testing cluster.

**Can containerinsight-eks as well as containerinsight-ecs be merged into existing modules?**
Before this question can be answered, I personally would like to be provided a best-practice guidance for adding new test cases on difference platforms. For instance, is `eks` used to validate the main functions from the platform perspective or is it served for more general purpose, to hold all the validations happen on EKS? Is the current design/implementation capable of doing that? What's the minimum test set of `eks` module? How are different submodules isolated during runtime? Is it possible to run a subset that I'm interested in? What are the semantics of the common variables? Docs?

Without these instructions and agreed contracts, it's hard for contributors to follow what is unspoken.

wyTrivail:
> > my major concern is still, we are creating a new folder for eks, i'm still thinking about how we can minimize the duplicate parts, for example:
> > 
> > 1. we are using testing_id to create namespace,
> > 2. we are deploying aoc with configmap.
> > 3. we will need to support mocked server which is already supported in eks folder today.
> > 
> > the eks folder today can serve all the testcases, `cortex` is just an optional parameter. this [file](https://github.com/aws-observability/aws-otel-collector/blob/main/e2etest/testcases.json#L4) lists all the testcases which could be running via eks folder.
> > i'd like to ask myself if we want to have a separate folder for some testcases,
> > 
> > 1. if the difference is in ot config, we might be able to just configure it in testcase?
> > 2. if the difference is on how to deploy otcollector, why can't we add a new "starting mode" into eks folder?
> > 3. if the difference is on how to deploy sample apps, why can't we make the deployment of sample apps configurable? or add a file under eks folder?
> > 4. if the difference is on validation, we might be able to just configure the parameters in testcase?
> > 5. how hard would it be to make eks folder more generic?
> > 
> > If we can have a plan around how do we maintain the separate folders, either we have solid reasons to keep them separate, or we have a vision to merge them someday, i will be okay with that.
> 
> Add some thoughts for people who are interested in this comment. Please note this is still an open question needs to be addressed later in the future.
> 
> **Correction:**
> The otel configuration used in the integration test is _NOT_ a testing config but rather a config that would be released to users for container insight on EKS. It will be built in the image, accessible directly. We're _NOT_ deploying otel collector with configmaps. No extra setup process is required when deloying aoc in the testing cluster.
> 
> **Can containerinsight-eks as well as containerinsight-ecs be merged into existing modules?**
> Before this question can be answered, I personally would like to be provided a best-practice guidance for adding new test cases on difference platforms. For instance, is `eks` used to validate the main functions from the platform perspective or is it served for more general purpose, to hold all the validations happen on EKS? Is the current design/implementation capable of doing that? What's the minimum test set of `eks` module? How are different submodules isolated during runtime? Is it possible to run a subset that I'm interested in? What are the semantics of the common variables? Docs?
> 
> Without these instructions and agreed contracts, it's hard for contributors to follow what is unspoken.

1. when we designed this framework, we were trying to decouple otconfig, sample app, validation, and computing platforms as a test case is constructed by the four, so that when we define a testcase in testcases folder, we define what's the otconfig, what's the sample app, and what's the validation config, also define which platform this testcase will be running. it's not about should we hold all the validations happen on EKS as validation and computing platform are decoupled.  

2.  Is the current design/implementation capable of doing that?  At least we covered all the testcases so far. As maintainers, we should know it's not mature and understand that there will be new types of testcases coming in anyway and we need to keep improving it gradually. 

3.  What's the minimum test set of `eks` module? again, we were trying to decouple that, EKS module is a module being able to deploy collector and sample apps, if there's a testcase we can't cover, we need to think about which part is missing.

4. How are different submodules isolated during runtime? what kind of isolation? there will be no isolation issue as each testcase is running separately in different container,  and i don't think they need to be isolated.

5. What are the semantics of the common variables? Docs?  could you point out which parameter is confusing so that we can improve it? 

what's the biggest intention bring us to create a new folder? can we define more test cases in this folder?

mxiamxia (approved):
LGTM. thx

bjrara:
Just to make myself clear, I'm not opposed to merge container insight test cases to `eks`. I'm just proposing more requirement to the framework, specifically the documentation.
As you said,
> if there's a contributor who wants to add a new testcase, how do we guide him to add.

They clearly wouldn't be guided by my PR, but documents of `HOW TO`s.

> 4\. How are different submodules isolated during runtime? what kind of isolation? there will be no isolation issue as each testcase is running separately in different container,  and i don't think they need to be isolated.
> 5\. What are the semantics of the common variables? Docs?  could you point out which parameter is confusing so that we can improve it?
> 
> what's the biggest intention bring us to create a new folder? can we define more test cases in this folder?

Add more thoughts on the testing,

4. Isolation means the ability to run one/multiple test scenario(s) at one time. Users should be able to trigger a subset of the provided test cases with the minimum resources to be requested. Assuming containerinsight is finally merged with eks, can it be run without creating "cortex" and and "sample_app_deployment"?
Furthermore, whether testcases can be run locally in separate containers depends on where the docker-compose file is created. I'm not sure about the github flow, but in my local environment, if they are in the same folder, running in different containers would result in sharing the same network, and will fail if running concurrently. That's another thing we need to take care of in the following actions.

5. I think a clear definition on common variables is always a good way to avoid misunderstanding and confusion. I just don't feel safe to use any variable that I don't understand completely: https://github.com/aws-observability/aws-otel-test-framework/blob/main/terraform/common.tf.

bjrara:
> Ship it. i still have many questions which i hope we address them with opening an issue though.

Created an issue: https://github.com/aws-observability/aws-otel-test-framework/issues/258. I thought merging the folders is the only one left unresolved, isn't it?

wyTrivail:
> > Ship it. i still have many questions which i hope we address them with opening an issue though.
> 
> Created an issue: #258. I thought merging the folders is the only one left unresolved, isn't it?

and merging the validation part i guess?

bjrara:
> > > Ship it. i still have many questions which i hope we address them with opening an issue though.
> > 
> > 
> > Created an issue: #258. I thought merging the folders is the only one left unresolved, isn't it?
> 
> and merging the validation part i guess?

Yup. Added. Thanks!

wyTrivail:
> 1. Isolation means the ability to run one/multiple test scenario(s) at one time. Users should be able to trigger a subset of the provided test cases with the minimum resources to be requested. Assuming containerinsight is finally merged with eks, can it be run without creating "cortex" and and "sample_app_deployment"?
>    Furthermore, whether testcases can be run locally in separate containers depends on where the docker-compose file is created. I'm not sure about the github flow, but in my local environment, if they are in the same folder, running in different containers would result in sharing the same network, and will fail if running concurrently. That's another thing we need to take care of in the following actions.

i'm not sure if it's easy to achieve running multiple test together, we made assumption that each test is running separately and rely on github workflow for parallelization.   
But we should okay to use a subset of functions with flags, like right now, "cortex" is an optional one, we can define different deploy mode for sample apps and allow the sample apps deployments being to able to be plugged into the eks module.  so that contributor will only need to develop the sample app deployment part but not the whole eks things.

bjrara:
> > 1. Isolation means the ability to run one/multiple test scenario(s) at one time. Users should be able to trigger a subset of the provided test cases with the minimum resources to be requested. Assuming containerinsight is finally merged with eks, can it be run without creating "cortex" and and "sample_app_deployment"?
> >    Furthermore, whether testcases can be run locally in separate containers depends on where the docker-compose file is created. I'm not sure about the github flow, but in my local environment, if they are in the same folder, running in different containers would result in sharing the same network, and will fail if running concurrently. That's another thing we need to take care of in the following actions.
> 
> i'm not sure if it's easy to achieve running multiple test together, we made assumption that each test is running separately and rely on github workflow for parallelization.
> But we should okay to use a subset of functions with flags, like right now, "cortex" is an optional one, we can define different deploy mode for sample apps and allow the sample apps deployments being to able to be plugged into the eks module. so that contributor will only need to develop the sample app deployment part but not the whole eks things.

A separate `project` needs to be assigned when executing `docker-compose` to run tests simultaneously. It can be fixed during layout refactoring as long as the agreement is made.

wyTrivail:
> > > 1. Isolation means the ability to run one/multiple test scenario(s) at one time. Users should be able to trigger a subset of the provided test cases with the minimum resources to be requested. Assuming containerinsight is finally merged with eks, can it be run without creating "cortex" and and "sample_app_deployment"?
> > >    Furthermore, whether testcases can be run locally in separate containers depends on where the docker-compose file is created. I'm not sure about the github flow, but in my local environment, if they are in the same folder, running in different containers would result in sharing the same network, and will fail if running concurrently. That's another thing we need to take care of in the following actions.
> > 
> > 
> > i'm not sure if it's easy to achieve running multiple test together, we made assumption that each test is running separately and rely on github workflow for parallelization.
> > But we should okay to use a subset of functions with flags, like right now, "cortex" is an optional one, we can define different deploy mode for sample apps and allow the sample apps deployments being to able to be plugged into the eks module. so that contributor will only need to develop the sample app deployment part but not the whole eks things.
> 
> A separate `project` needs to be assigned when executing `docker-compose` to run tests simultaneously. It can be fixed during layout refactoring as long as the agreement is made.

probably check https://github.com/nektos/act, which enable us run workflow locally. It was not too mature last year where credentials can not be used, not sure if it's okay to use now.

And we will need to think about do we have such requirement of running test simultaneously in local, normally we develop one testcase one time, and should be able to use workflow to run regression test.  From resource consumption wise, although tests are running together, the resources are still the same, the cpu/memory consumption should be the same, each sample apps/collector will still take its resource. Furthermore, since these tests are short-time test, i'm not sure if we really care about the consumption.  We might be concerning on soaking and performance test, but it's better to isolate these tests on instance level so that we can get a better tuning result. So far I don't see too much benefit to run test simultaneously, happy to know if we have any idea on it.

bjrara:
> > > > 1. Isolation means the ability to run one/multiple test scenario(s) at one time. Users should be able to trigger a subset of the provided test cases with the minimum resources to be requested. Assuming containerinsight is finally merged with eks, can it be run without creating "cortex" and and "sample_app_deployment"?
> > > >    Furthermore, whether testcases can be run locally in separate containers depends on where the docker-compose file is created. I'm not sure about the github flow, but in my local environment, if they are in the same folder, running in different containers would result in sharing the same network, and will fail if running concurrently. That's another thing we need to take care of in the following actions.
> > > 
> > > 
> > > i'm not sure if it's easy to achieve running multiple test together, we made assumption that each test is running separately and rely on github workflow for parallelization.
> > > But we should okay to use a subset of functions with flags, like right now, "cortex" is an optional one, we can define different deploy mode for sample apps and allow the sample apps deployments being to able to be plugged into the eks module. so that contributor will only need to develop the sample app deployment part but not the whole eks things.
> > 
> > 
> > A separate `project` needs to be assigned when executing `docker-compose` to run tests simultaneously. It can be fixed during layout refactoring as long as the agreement is made.
> 
> probably check https://github.com/nektos/act, which enable us run workflow locally. It was not too mature last year where credentials can not be used, not sure if it's okay to use now.
> 
> And we will need to think about do we have such requirement of running test simultaneously in local, normally we develop one testcase one time, and should be able to use workflow to run regression test. From resource consumption wise, although tests are running together, the resources are still the same, the cpu/memory consumption should be the same, each sample apps/collector will still take its resource. Furthermore, since these tests are short-time test, i'm not sure if we really care about the consumption. We might be concerning on soaking and performance test, but it's better to isolate these tests on instance level so that we can get a better tuning result. So far I don't see too much benefit to run test simultaneously, happy to know if we have any idea on it.

Some statistics:
A history discussion can be found: https://github.com/aws-observability/aws-otel-test-framework/pull/247#discussion_r604332152
For each workload in containerinsight, setup (install resources in EKS) 1 minute, docker compile 2 minutes, wait logs and metrics ready 3 minutes.
If workload is run one by one in different test cases (although I took a different approach), it would take 30 minutes to run the entire tests. Running concurrently takes 6 minutes. 

We had 6 nodes in the testing cluster, for all the workload in containerinsight, it requires a successful creation of 19 Pods (sample, traffic generator, component control plane). How many integration flow can be executed at one time without possible impact of resource shortage. 

If contributors want to run on their on EKS cluster, how many nodes is recommended to be created for our framework? They pay for it.

bjrara:
@wyTrivail Thanks for the careful review and patient discussion with me about the understanding of design and what the framework should/would be! It was a great honour to share opinions with you. But I think there's too much information in this PR and I like to end the discussion for now. I believe most of the topics can be better understood by future maintainers.

PR #248, opened 2021-03-26
Bump OTel Java SDK to 1.0.x for sample apps

mxiamxia:
## Description
Bump OTel Java SDK to 1.0.x for sample apps

## Test
local run


terraform/templates/local/docker_compose_from_source.tpl
```diff
@@ -40,7 +40,7 @@ services:
       - AWS_REGION=${region}
       - OTEL_RESOURCE_ATTRIBUTES=${otel_resource_attributes}
       - INSTANCE_ID=${testing_id}
-      - OTEL_EXPORTER_OTLP_ENDPOINT=aws-ot-collector:${grpc_port}
+      - OTEL_EXPORTER_OTLP_ENDPOINT=http://aws-ot-collector:${grpc_port}
```
pingleig, line 43: I think we don't need `http://` for grpc right? Or something has changed in sdk and requires `http`?

And will it cause problem for other language? Though I think we only have java sample app push to otlp receiver.

load-generator/build.gradle
```diff
@@ -29,9 +29,14 @@ dependencies {
     implementation "io.grpc:grpc-api:1.34.1"
 
     implementation "io.grpc:grpc-netty-shaded:1.34.1"
-    implementation "io.opentelemetry:opentelemetry-sdk:0.12.0"
-    implementation "io.opentelemetry:opentelemetry-exporter-otlp:0.12.0"
-    implementation "io.opentelemetry:opentelemetry-sdk-extension-aws:0.12.0"
+
+    implementation platform("io.opentelemetry:opentelemetry-bom:1.0.1")
```
pingleig, line 33: Can we use same style for import? https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.dsl.DependencyHandler.html#N17433

I think we can use `implementation 'foo.bar'` for most packages that does not require extra notation (e.g. platform)

sample-apps/spark/Dockerfile
```diff
@@ -7,7 +7,7 @@ COPY ./settings.gradle ./settings.gradle
 
 RUN gradle build
 RUN tar -xvf build/distributions/spark-sample-app-1.0.tar
-RUN wget https://github.com/aws-observability/aws-otel-java-instrumentation/releases/download/v0.12.1-aws.1/aws-opentelemetry-agent.jar
+RUN wget https://github.com/aws-observability/aws-otel-java-instrumentation/releases/download/v0.18.0-aws.1/aws-opentelemetry-agent.jar
```
pingleig, line 10: Maybe we should make it an `ARG` in Dockerfile and give it a default value. In that case we can bump the version in tf files.
Does all otel sdk (for different languages) follows same version number? Or each language has its own?

sample-apps/spark/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Tue Mar 16 11:14:39 PDT 2021
```
pingleig, line 1: Actually we can bump gradle wrapper to a new major version (though I think this is modified by IDEA because it used to suggest me to download full gradle to get better competition etc.

pingleig (approved):
LGTM Just some style issues (btw: I didn't run it locally to validate the code ...)

terraform/templates/local/docker_compose_from_source.tpl
mxiamxia: yes, OTel SDK added this recently.

```
 /**
   * Sets the OTLP endpoint to connect to. If unset, defaults to {@value DEFAULT_ENDPOINT_URL}. The
   * endpoint must start with either http:// or https://.
   */
  public OtlpGrpcMetricExporterBuilder setEndpoint(String endpoint) {
```

sample-apps/spark/Dockerfile
mxiamxia: This is Java Auto Instrument Agent for Tracing. It is not OTel SDK so the version # is not in v1.0.x yet.

This sample app image doesn't change frequently yet. I like your `ARG` idea. I'll save it for later. :)

sample-apps/spark/gradle/wrapper/gradle-wrapper.properties
mxiamxia: yea. it is updated by IDE.

mxiamxia:
> LGTM Just some style issues (btw: I didn't run it locally to validate the code ...)

Thanks, Addressed the style issue

PR #252, opened 2021-03-28
sample: Add jmx prometheus metrics with tomcat

pingleig:
Port from cwagent public doc https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/ContainerInsights-Prometheus-Sample-Workloads-ECS-javajmx.html

It only expose prometheus metrics and does not support trace nor does it push anything to anywhere. The `/` for app returns 404 but we only care about the port from prometheus agent. Though visit `/samplejmx/` would update some prometheus metrics.


bjrara (approved):
LGTM

bjrara (approved):
LGTM

bjrara (approved):
LGTM

Issue #253, opened 2021-03-29
[aws][ci] Install SSM agent for all distros in EC2 test and enable patch by default


pingleig:
Currently only AMI from Amazon ships with SSM agent by default, which is required for patching instances (e.g. #216 ).

## TODO

- [x] list AMIs that does not come with SSM #254 
- [x] list AMIs that does not come with python (e.g. RHEL 8)
- [x] use user data to install SSM agent
- [x] set patch to true when running in github action aws-observability/aws-otel-collector#431

pingleig:
Done, though short term tests didn't enable patch because the CI would be too slow with it.

PR #254, opened 2021-03-29
Install SSM agent for all AMIs

vastin:
Install SSM agent for all AMIs which hasn't it by default. It is part of https://github.com/aws-observability/aws-otel-test-framework/issues/253
Refer to https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-manual-agent-install.html

### DIFF
--- a/terraform/ec2/amis.tf
+++ b/terraform/ec2/amis.tf
@@ -25,6 +25,7 @@ variable "ami_family" {
       start_command = "sudo /opt/aws/aws-otel-collector/bin/aws-otel-collector-ctl -c /tmp/ot-default.yml -a start"
       connection_type = "ssh"
       user_data = ""
+      wait_cloud_init = "for i in {1..60}; do [ ! -f /var/lib/cloud/instance/boot-finished ] && echo 'Waiting for cloud-init...' && sleep 1 || break; done"
     }
     linux = {
       login_user = "ec2-user"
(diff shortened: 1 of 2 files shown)


terraform/ec2/amis.tf
```diff
@@ -221,6 +267,12 @@ EOF
       family = "linux"
       arch = "arm64"
       instance_type = "t4g.nano"
+      user_data = <<EOF
```
pingleig, line 270: I think all AMI owned by amz has ssm by default? Though it won't hurt to reinstall ssm even if AMI has ssm agent already, as it simply update ssm agent.

terraform/ec2/amis.tf
```diff
@@ -150,6 +167,14 @@ variable "amis" {
       family = "linux"
       login_user = "ec2-user"
       arch = "amd64"
+      user_data = <<EOF
```
pingleig, line 170: terraform does not allow you to define string constant that can be used in variable defaults, and you can not use local or variable in variable defaults as well, so copy the userdata everywhere is the only way ... /w

Another way is we add userdata based on os family, but some os need extra init logic like install python (for patching #253 ). I prefer we just inline the userdata here as we don't need to update them most of the time.

pingleig (approved):
LGTM

pingleig:
Actually we forgot to update all the`amis.tf`,  there are separated files for canary and soaking, I made same mistake last time https://github.com/aws-observability/aws-otel-test-framework/pull/239/files

It is causing 

```
Error: Invalid index: The given key does not identify an element in this collection value.
```

PR #257, opened 2021-03-30
Fix nil string issue in remote-exec

vastin:
The remote-exec provisioner doesn't support empty string as inline. Fix following error for windows AMI which failed canary test.
 Error: Error parsing <nil> as a string

Refer to:
https://github.com/hashicorp/terraform/issues/13970
https://github.com/hashicorp/terraform/pull/14134

### DIFF
--- a/terraform/canary/amis.tf
+++ b/terraform/canary/amis.tf
@@ -36,7 +36,7 @@ net start winrm
 Set-NetFirewallProfile -Profile Public -Enabled False
 </powershell>
 EOF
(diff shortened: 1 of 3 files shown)


pingleig (approved):
LGTM

pingleig:
Though you might want to add one line comment above those 'empty' strings.

vastin:
> 
> 
> Though you might want to add one line comment above those 'empty' strings.

Thx for comment. I add reason in commit message instead.

PR #262, opened 2021-04-01
Restrict kubernetes version to be installed

bjrara:
Got the following error when running as CI job remotely. It is caused by terraform downloading kubernetes v2 instead of v1.
```
Error: Unsupported argument
  on main.tf line 47, in provider "kubernetes":
  47:   load_config_file = false
```

This PR is created to fix this issue, and is validated by `terraform init -update && terraform apply`.


pingleig (approved):
LGTM

Issue #255, opened 2021-03-30
[framework] Port resource clean up logic from deprecating branch


pingleig:
terraform fails to delete resource for various reasons (e.g. test took too long and its k8s token is outdated, leaving namespace around).

- [ ] [ec2](https://github.com/aws-observability/aws-otel-test-framework/blob/deprecating/integ-test/src/main/java/com/amazon/aocagent/tasks/EC2Clean.java) 
- [ ] [eks namespace](https://github.com/aws-observability/aws-otel-test-framework/blob/deprecating/integ-test/src/main/java/com/amazon/aocagent/tasks/EKSClean.java)
- [ ] efs mount and efs file system

Most env setup logic is in tf, but it's hard to write resource clean up logic in tf. Further more we need to deal with problems like #253, which we are using shell script, I plan to write a go util to handle all those util tasks so it's easier to extend and maintain in the future.

pingleig:
I think existing resource cleaner does not work because it can't match tag. And we didn't distinguish ec2 instances for canary and soaking, so it's hard to terminate instances. Until we are able to distinguish those there is no good way to clean up instances generated from CI.yaml ...

vasireddy99:
Closing this issue by referencing to available [weekly resources](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/weekly-resources-clean.yml#L36). Please re-open this issue if this is not the expected implementation.

PR #264, opened 2021-04-02
Add SSM package test for ADOT Collector in ec2 test

vastin:
Add SSM package validation to ADOT collect test framework. The usage of new testcase `ssm_package` is as same as existing ec2 testcases. The var `ssm_package_name` and var `ssm_config` are parameters of the new testcase.

### DIFF
--- a/terraform/ec2/main.tf
+++ b/terraform/ec2/main.tf
@@ -70,6 +70,9 @@ locals {
 
   # get instance subnet, use separate subnet for soaking and performance test as it takes more instances and some times eat up all the available ip address in the subnet
   instance_subnet = var.testing_type == "e2e" ? tolist(module.basic_components.aoc_public_subnet_ids)[1] : tolist(module.basic_components.aoc_public_subnet_ids)[0]
+
+  # get SSM package version, latest is the default version
+  ssm_package_version = var.aoc_version == "latest" ? "\"\"" : trimprefix(var.aoc_version, "v")
 }
 
 ## launch a sidecar instance to install data emitter and the mocked server
@@ -228,6 +231,7 @@ resource "null_resource" "download_collector_from_s3" {
 }
 
 resource "null_resource" "start_collector" {
+  count = var.install_package_source == "ssm" ? 0 : 1
   # either getting the install package from s3 or from local
   depends_on = [null_resource.download_collector_from_local, null_resource.download_collector_from_s3]
   provisioner "file" {
@@ -260,6 +264,38 @@ resource "null_resource" "start_collector" {
   }
 }
 
+resource "aws_ssm_parameter" "setup_aoc_config" {
+  count = var.install_package_source == "ssm" ? 1 : 0
+  name  = format("aoc-config-%s", module.common.testing_id)
+  type  = "String"
+  value = var.ssm_config
+}
+
+resource "null_resource" "install_collector_from_ssm" {
+  depends_on = [null_resource.check_patch, aws_ssm_parameter.setup_aoc_config]
+  count = var.install_package_source == "ssm" ? 1 : 0
+
+  provisioner "remote-exec" {
+    inline = [
+      local.ami_family["wait_cloud_init"],
+    ]
+
+    connection {
+      type = local.connection_type
+      user = local.login_user
+      private_key = local.connection_type == "ssh" ? local.private_key_content : null
+      password = local.connection_type == "winrm" ? rsadecrypt(aws_instance.aoc.password_data, local.private_key_content) : null
+      host = aws_instance.aoc.public_ip
+    }
+  }
+
+    provisioner "local-exec" {
+      command = <<-EOT
+        bash ../templates/local/ssm-install-aoc.sh ${aws_instance.aoc.id} ${var.ssm_package_name} ${local.ssm_package_version} ${aws_ssm_parameter.setup_aoc_config[0].name}
+    EOT
+  }
+}
+
 ########################################
 # Start Sample app and mocked server
 #########################################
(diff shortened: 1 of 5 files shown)


terraform/common.tf
```diff
@@ -71,3 +71,16 @@ variable "cortex_instance_endpoint" {
   # change to your cortex endpoint
   default = "https://aps-workspaces-gamma.us-west-2.amazonaws.com/workspaces/ws-31eb305d-3208-42d5-a7f4-32ce1191e699"
 }
+
+variable "ssm_mode" {
```
pingleig: maybe change it to `install_mode` with some 'enums' like package manager (i.e. download deb/rpm directly in user data) or `ssm`

Or reuse `install_package_source` (See comments in other files).

terraform/templates/local/ssm-install-aoc.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
pingleig, line 1: Actually we can convert those shell script into go code to make life easier. I haven't created the PR for https://github.com/aws-observability/aws-otel-test-framework/issues/255 it has some ssm wrapper for patching related ssm status query.

terraform/ec2/main.tf
```diff
@@ -210,7 +213,7 @@ resource "null_resource" "download_collector_from_local" {
 
 resource "null_resource" "download_collector_from_s3" {
   depends_on = [null_resource.check_patch]
-  count = var.install_package_source == "s3" ? 1 : 0
+  count = var.install_package_source == "s3" && !var.ssm_mode ? 1 : 0
```
pingleig: Can we reuse `install_package_source` to have `ssm` instead of having both `install_package_source` and `ssm_mode`?

terraform/ec2/main.tf
```diff
@@ -124,7 +127,7 @@ data "template_file" "mocked_server_cert_for_windows" {
 }
 resource "null_resource" "setup_mocked_server_cert_for_windows" {
   depends_on = [null_resource.check_patch]
-  count = local.selected_ami["family"] == "windows" ? 1 : 0
+  count = local.selected_ami["family"] == "windows" && !var.ssm_mode ? 1 : 0
```
pingleig: Why we want to disable setup mocked server cert when using ssm?

terraform/ec2/main.tf
```diff
@@ -260,6 +264,38 @@ resource "null_resource" "start_collector" {
...
+      type = local.connection_type
+      user = local.login_user
+      private_key = local.connection_type == "ssh" ? local.private_key_content : null
+      password = local.connection_type == "winrm" ? rsadecrypt(aws_instance.aoc.password_data, local.private_key_content) : null
+      host = aws_instance.aoc.public_ip
+    }
+  }
+
+    provisioner "local-exec" {
+    command = <<-EOT
```
pingleig: There should be indent here, if you install terraform plugin, it can format the file like other languages. Though existing format in ec2/main.tf is not consistent, and you can configure tf to align properties in different ways.

I will create another pr for the formatting, format the entire file will the pr hard to read. https://github.com/aws-observability/aws-otel-test-framework/issues/265

Screen Shot 2021-04-03 at 2 45 39 PM

pingleig (commented):
It seems there is no test for reload config,  there will be other PRs?

btw: Is the ssm doc already published or it's in your own account?

vastin:
> It seems there is no test for reload config, there will be other PRs?
The SSM command document for reload config is in development. Right now it only support installation with configuration.

> btw: Is the ssm doc already published or it's in your own account?
The default SSM package document is a test document in my account. The publish workflow will be added in CI/CD.

terraform/common.tf
vastin: Agree. I will use install_package_source for installation methods. However I want to keep ssm_mode to disable terraform resources for SSM testcase.

terraform/templates/local/ssm-install-aoc.sh
vastin: Agree. Golang will make error handling easier. No matter what language, the logic of async execution is required. The testcase should fail when any error occurs. We can use bash script for now and convert its logic to golang later.

terraform/ec2/main.tf
vastin: Agree.

terraform/ec2/main.tf
vastin: Actually SSM testcase don't use mocked server instance. It will only check whether ADOT Collector is running in aoc instance.

terraform/ec2/main.tf
vastin: Agree

terraform/ec2/main.tf
pingleig: I think it's better to add a var to disable mocked server to common e.g. `var.disable_mocked_server`, because there are other tests that don't require mocked server. Then when testing ssm you can set that variable to true. I am not sure if there is one hanging around, though since ssm is only for ec2 using ssm_mode should be fine as well.

pingleig (commented):
It seems we are still using `ssm_mode` instead of `install_package_source` ? Do you want to change it in this PR or have another PR for that?

terraform/ec2/main.tf
vastin: Good idea. I add `disable_mocked_server` and `enable_ssm_validate` to replace `ssm_mode`. All new variable definitions are moved into terraform/ec2/variables.tf because it is only for ec2 test.

pingleig (approved):
LGTM

vastin:
It shouldn't impact existing CI/CD. So merge it now.

PR #267, opened 2021-04-05
aotutil: Init go util for monitor ec2 instance patching via ssm

pingleig:
Part of #253  A working action is in [my fork](https://github.com/pingleig/aws-otel-collector/runs/2272240057?check_suite_focus=true#step:10:151)

- a new go util to do the ssm patch check logic
- not enabled yet because it will break current CI until we update current CI
- all the EC2 tests will now run at least for 30 minutes (when patch is enabled) so it can report the patched status (if you find this behaviour strange, the reason is we can't run the report document directly as it can only triggered by association, and the minimal interval is 30 minutes)

This go util can be used for other things as well

Also updated some doc

- how to create eks cluster
- how to patch eks cluster


.gitignore
```diff
@@ -24,6 +24,8 @@ sshkey.pem
 # Local .terraform directories
 **/.terraform/*
 terraform/**/.terraform.lock.hcl
+terraform/**/creds.env
+terraform/**/validator_docker_compose.yml
```
mxiamxia, line 28: 👍

hack/aotutil/.gitignore
```diff
@@ -0,0 +1,5 @@
+# binary
+aotutil
+
```
mxiamxia: can we not use `hack` as folder name?

pingleig: There are other projects using hack folder as well, e.g. k8s https://github.com/kubernetes/kubernetes/tree/master/hack

Though we can change it to something like `tools` or move it to a `cmd/aotutil` so it's more go project style, but this is a java project ...

pingleig: Moved to cmd cc @bjrara PTAL

cmd/aotutil/main.go
```diff
@@ -0,0 +1,130 @@
...
+	// Wait Patch Report
+	watchPatchReport := cobra.Command{
+		Use:   "wait-patch-report",
+		Short: "Wait until a patched instance is reported",
+		Run: func(cmd *cobra.Command, args []string) {
+			instance := oneInstanceId(args)
+			if err := ssmWrapper.WaitPatchReported(ctx, instance, ssmWaitTimeout); err != nil {
+				logger.Fatal("WaitPatch failed", zap.Error(err))
+			} else {
+
```
bjrara: Could you please remove the empty block?

cmd/aotutil/util.go
```diff
@@ -0,0 +1,58 @@
...
+const (
+	WaitContinue WaitAction = iota + 1
+	WaitDone
+)
+
+func Wait(interval, timeout time.Duration, f func() (WaitAction, error)) error {
+	// Run f once
+	act, err := f()
+	if err != nil {
+		return err
```
bjrara, line 37: I would prefer checking `act` as well before return in case of failure retries.

pingleig: it stops on error and we expect f to swallow the error if it thinks it is retry-able. I have updated the comment.

bjrara (approved):
LGTM

PR #268, opened 2021-04-08
Add http prefix for ECS and EKS OTLP receiver endpoint

mxiamxia:
Add http prefix for ECS and EKS OTLP receiver endpoint


pingleig (approved):
Same SDK change (like the java one?)

PR #269, opened 2021-04-08
patch: Use aotutil for check patch and wait report

pingleig:
For #253 

- switch from shell script to go binary.
- wait 30 mins before destroy ... 

Example in my [fork](https://github.com/pingleig/aws-otel-collector/runs/2301293915?check_suite_focus=true) test itself failed but all the patching and reporting logic are working ...


vastin (approved):
LGTM

pingleig:
We need to hold on merge until https://github.com/aws-observability/aws-otel-collector/pull/431 is merged.

Though I expect resource cleaner logic to be broken #255 ... we can fix it later as we can do resource cleaning without terraform.

mxiamxia (approved):
Thanks!

PR #272, opened 2021-04-12
Fix validate build with gradle v7.0'

vastin:
Validator docker build failed after gradle docker upgrade to v7.0. This PR should fix it and recover integration test.

### DIFF
--- a/validator/build.gradle
+++ b/validator/build.gradle
@@ -31,28 +31,28 @@ repositories {
 }
 
 dependencies {
-    testCompile group: 'junit', name: 'junit', version: '4.12'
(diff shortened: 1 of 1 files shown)


bjrara (approved):
LGTM

pingleig (approved):
LGTM

btw: we are using gradle wrapper so our gradle version is locked to 6.x I suppose?

https://github.com/aws-observability/aws-otel-test-framework/blob/e1cf7b7befa4a4a79e187162e841417774924a02/gradle/wrapper/gradle-wrapper.properties#L2

PR #277, opened 2021-04-15
tf: Add ssm policy for ecs ec2 instances

pingleig:
This would allow ssm to report host status, patch does not work until we pass the patch tag


vastin (approved):
LGTM

Issue #258, opened 2021-03-31
Module layout for Container Insight integration tests


bjrara:
When initiating integration tests for container insight in PR: https://github.com/aws-observability/aws-otel-test-framework/pull/247, we debated a lot on whether test cases should be merged into `/terraform/eks` or remain in a new folder `/terraform/containerinsight-eks-prometheus`.

This issue is created to keep track of the remaining tasks:
1. Merge `/terraform/containerinsight-eks-prometheus` into `/terraform/eks`. (PR in review)
2. Merge `CWMetricValidator` and `ContainerInsightMetricValidator`.
3. Provide design docs, and guidance for adding new test cases. 

Full history records can be found in https://github.com/aws-observability/aws-otel-test-framework/pull/247#issuecomment-810725653.

wyTrivail:
thanks Mengyi! that's pretty good point we lack of design doc, i will put the design doc here later.

pingleig:
I suppose it's done? And @wyTrivail is there any design doc? :doge:

wyTrivail:
https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/docs/testing-framework-design.md

PR #280, opened 2021-04-16
Modularize EKS terraform workflow

bjrara:
**Why do we need it?**
The PR optimizes the current EKS terraform with a clearer structure.
1. `/eks/main.tf` contains the components that are shared among eks and container insights.
2. `/eks/oltp.tf` creates aoc deployment and mocked server.
2. In the `/eks/otlp/` folder where resource management is modularized and stateless, we create sample apps based on the pull/push mode. 
3. The aoc deployment and mocked server for general otlp testcases can be toggled by `aoc_base` to prepare for the merge of container insights. (The container insights aoc testing doesn't provision mocked server related resources because it didn't configure oltp as a receiver.)

This is part of the refactoring work for https://github.com/aws-observability/aws-otel-test-framework/issues/258


terraform/eks/otlp.tf
```diff
@@ -0,0 +1,203 @@
+module "basic_components" {
```
wyTrivail, line 16: license header?

terraform/eks/otlp/shared.tf
```diff
@@ -0,0 +1,40 @@
+# ------------------------------------------------------------------------
```
wyTrivail: i assume shared.tf will not only be used for config file, but all the shared resources?

terraform/eks/variables.tf
```diff
@@ -20,3 +20,7 @@ variable "eks_cluster_name" {
 variable "mock_endpoint" {
   default = "localhost/put-data"
 }
+
+variable "aoc_base" {
+  default = "oltp"
```
wyTrivail, line 27: what's the other possible values for aoc_base? trying to understand it's meaning.

wyTrivail (commented):
small nits and questions, thanks

terraform/eks/variables.tf
bjrara: Added comments to the variable. The deployment spec of AOC will be different under different scenarios, this is used to identify the testing purpose.

terraform/eks/otlp/shared.tf
bjrara: Changed the name to `oltp.tf` to avoid confusion. Basically it's the file that is used for both `pull mode` and `push mode`, sth. like the `main.tf` in a module, but I couldn't think of a good name. Any suggestion is also welcomed.

terraform/eks/otlp.tf
bjrara: Nice catch, thx!

wyTrivail (approved):

PR #281, opened 2021-04-16
Merge container insights prometheus and eks

bjrara:
**Why do we need it?**
This is a follow-up PR of https://github.com/aws-observability/aws-otel-test-framework/pull/280, and is the second part of refactoring work to address https://github.com/aws-observability/aws-otel-test-framework/issues/258.

This PR
1. Add IAM policy creation in the setup flow.
2. Merge container insights prometheus into eks folder.

**Note**
This PR should be merged after container insights integration test cases are fully merged and tested. So hold the review for now.


terraform/eks/agent.tf
```diff
@@ -0,0 +1,53 @@
+data "template_file" "cluster_role_file" {
```
pxaws: I would suggest to add some comments for this file to explain it is for container insights. Or you could rename the file to `CIAgent.tf`

terraform/eks/agent.tf
```diff
@@ -0,0 +1,53 @@
...
+  vars = {
+    NAMESPACE = kubernetes_namespace.aoc_ns.metadata[0].name
+  }
+}
+
+data "template_file" "daemonset_file" {
+  template = file("./node-agent/daemonset.tpl")
+  vars = {
+    NAMESPACE       = kubernetes_namespace.aoc_ns.metadata[0].name
+    SERVICE_ACCOUNT = kubernetes_service_account.aoc-agent-role.metadata[0].name
```
pxaws: So the service account is shared for all eks test cases? Just wondering if there is any issues because different test cases require different permissions. If two test cases are running at the same time, will this service account be assigned with different permissions.

terraform/eks/agent.tf
bjrara: Oops, I thought CI stood for continuous integration...

terraform/eks/agent.tf
bjrara: I don't think so. `ServiceAccount` is namespaced level resource. Also `ClusterRole` and `ClusterRoleBinding` are all suffixed with testing id (although I do feel we can downgrade RBAC into namespaced level too). As long as the two test cases are not executed in a same container at the same time, we'll be good.

terraform/eks/appmesh/appmesh.tf
```diff
@@ -155,7 +150,7 @@ resource "null_resource" "appmesh_readiness_check" {
 }
 
 data "template_file" "traffic_deployment_file" {
-  template = file("${var.testcase}/appmesh_traffic_sample.tpl")
+  template = file("./appmesh/appmesh_traffic_sample.tpl")
```
wyTrivail, line 153: would be possible we place template files under `https://github.com/aws-observability/aws-otel-test-framework/tree/terraform/terraform/templates`?

terraform/eks/container_insights_agent.tf
```diff
@@ -0,0 +1,53 @@
+data "template_file" "cluster_role_file" {
+  template = file("./container-insights-agent/cluster_role.tpl")
+  vars = {
+    NAMESPACE = "aoc-ns-${module.common.testing_id}"
```
wyTrivail: better have a var as namespace and construct in one place (maybe in `shared`)?

terraform/eks/main.tf
```diff
@@ -22,6 +22,10 @@ terraform {
     kubernetes = {
       version = "~> 1.13"
     }
+    kubectl = {
+      source  = "gavinbunney/kubectl"
+      version = ">= 1.7.0"
```
wyTrivail, line 31: we use >= for kubectl and ~> for kubernetes, is it intended? or we are okay with big version upgrade?

terraform/eks/main.tf
```diff
@@ -81,47 +57,46 @@ provider "kubernetes" {
...
-resource "kubernetes_config_map" "aoc_config_map" {
-  metadata {
-    name      = "otel-config"
-    namespace = kubernetes_namespace.aoc_ns.metadata[0].name
-  }
-
-  data = {
-    "aoc-config.yml" = module.basic_components.otconfig_content
-  }
+provider "kubectl" {
```
wyTrivail, line 64: would every testcase need kubectl? 
but i think it's okay to leave it created since it makes not hurts and save "ifelse".

terraform/eks/main.tf
```diff
@@ -81,47 +57,46 @@ provider "kubernetes" {
...
-    "ca-bundle.crt" = module.basic_components.mocked_server_cert_content
+provider "helm" {
+  kubernetes {
+    host                   = data.aws_eks_cluster.testing_cluster.endpoint
+    cluster_ca_certificate = base64decode(data.aws_eks_cluster.testing_cluster.certificate_authority[0].data)
+    token                  = data.aws_eks_cluster_auth.testing_cluster.token
   }
 }
 
 locals {
```
wyTrivail: better put local vars definition at the beginning of the file

terraform/eks/main.tf
```diff
@@ -81,47 +57,46 @@ provider "kubernetes" {
...
-  eks_pod_config = yamldecode(data.template_file.eksconfig.rendered)["sample_app"]
+  eks_pod_config_path = fileexists("${var.testcase}/eks_pod_config.tpl") ? "${var.testcase}/eks_pod_config.tpl" : module.common.default_eks_pod_config_path
+}
+
+# create a unique namespace for each run
+resource "kubernetes_namespace" "aoc_ns" {
+  metadata {
+    name = "aoc-ns-${module.common.testing_id}"
```
wyTrivail, line 98: better have a var for namespace name as we use it multiplaces?

terraform/eks/main.tf
```diff
@@ -469,14 +145,38 @@ module "validator" {
   region                       = var.region
   testing_id                   = module.common.testing_id
   metric_namespace             = "${module.common.otel_service_namespace}/${module.common.otel_service_name}"
-  sample_app_endpoint          = "http://${var.sample_app_mode == "push" ? kubernetes_service.sample_app_service[0].load_balancer_ingress.0.hostname : kubernetes_service.pull_mode_sample_app_service[0].load_balancer_ingress.0.hostname}:${module.common.sample_app_lb_port}"
-  mocked_server_validating_url = "http://${kubernetes_service.mocked_server_service.load_balancer_ingress.0.hostname}/check-data"
-
+  sample_app_endpoint          = length(kubernetes_service.sample_app_service) > 0 ? "http://${kubernetes_service.sample_app_service.0.load_balancer_ingress.0.hostname}:${module.common.sample_app_lb_port}" : ""
+  mocked_server_validating_url = length(kubernetes_service.mocked_server_service) > 0 ? "http://${kubernetes_service.mocked_server_service.0.load_balancer_ingress.0.hostname}/check-data" : ""
+  cloudwatch_context_json = var.aoc_base == "prometheus" ? jsonencode({
```
wyTrivail: now i kinda understand what's `aoc_base`.  this name seems not quite verbose, and in some cases like testing promotheus exporter we uses mock server as well.  maybe think of another name? for example: `launch_mock_server` ?

if the only difference is mockserver, maybe we just leave mockserver there so that we don't need those ifelses?

terraform/eks/otlp/variables.tf
```diff
@@ -0,0 +1,33 @@
+variable "region" {
+  default = "us-west-2"
+}
+
+variable "testing_id" {
+  type = string
+}
+
+variable "eks_pod_config_path" {
```
wyTrivail: do we need a default value here?

terraform/setup/appmesh.tf
```diff
@@ -0,0 +1,99 @@
+resource "aws_iam_policy" "appmesh_k8s_iam_policy" {
```
wyTrivail, line 1: would be possible move it to the setup folder? and have a parameter to control whether these policies are needed? so that we keep setup material in one place.

terraform/testcases/containerinsight_eks/parameters.tfvars
```diff
@@ -1,2 +1,4 @@
 # this file is defined in validator/src/main/resources/validations
-validation_config = "eks-container-insight.yml"
\ No newline at end of file
+validation_config = "eks-container-insight.yml"
+
+aoc_base = "infra"
```
wyTrivail: i seems don't find "infra" used any places?

terraform/eks/appmesh/appmesh.tf
bjrara: I'm thinking would it be better to keep those related file together.

terraform/eks/main.tf
bjrara: This is a file ported from @pxaws 's PR. I think he could possibly give a good answer~

terraform/eks/main.tf
bjrara: No. Currently only container insight infra tests would use `kubectl` provider. As providers are global-wise configuration, I would prefer to have them stayed in `main.tf` to avoid duplicated declaration in the future. That's my original thinking.

terraform/setup/appmesh.tf
bjrara: I think it's already in setup folder, isn't it?

terraform/testcases/containerinsight_eks/parameters.tfvars
bjrara: It is used to overwrite the default value of `otlp` to avoid creating otlp related resources.

terraform/eks/otlp/variables.tf
bjrara: We could have one. I thought it was unnecessary because `/eks/otlp` is actually an inner module of `eks` (to isolate bootstrap process for otls, prometheus and infra), the var input is taken cared of by us, and is not exposed to end user.

terraform/eks/main.tf
bjrara: Thanks for the second thought. Another difference is declared here. Let me know if the current name would look any better. https://github.com/aws-observability/aws-otel-test-framework/pull/280#discussion_r624341233

terraform/eks/container_insights_agent.tf
bjrara: No. There shouldn't be any hardcode elsewhere for namespace value. Updated to use reference.

terraform/eks/main.tf
bjrara: Done

terraform/eks/main.tf
bjrara: Thanks for pointing out the issue. I don't think we should hardcode namespace elsewhere.
https://github.com/aws-observability/aws-otel-test-framework/pull/281#discussion_r624363642

wyTrivail (approved):

PR #286, opened 2021-04-21
fix zipkin, jaeger, ecsmetric testcases

mxiamxia:
### Description
1. Allow zipkin and jaeger receiver default http port to EC2 security group
2. Update the ECS metric template to fix the validation errors


vastin (approved):
LGTM

PR #290, opened 2021-04-22
tf: Fix ignore error flag position for ec2 patch report

pingleig:
Introduced in https://github.com/aws-observability/aws-otel-test-framework/pull/285 I put the quote at the wrong place and didn't test it before commit.


mxiamxia (approved):
Thx

PR #291, opened 2021-04-23
Increase timeout for cloud-init and SSM operations

vastin:
The cloud-init might take up to 3 mins in redhat 7/8 AMI. Increase wait timeout of cloud-init and SSM operations in order to make SSM package test more stable. It won't increase running duration of most test cases.


pingleig (approved):
LGTM

I think the extra cloud init time  on redhat could be  installing python3 for os patching.

mxiamxia (approved):
Thanks

PR #292, opened 2021-04-23
reduce the canary testing cycles due to long running time

mxiamxia:
reduce the canary testing cycles because canary workflow took more than 1 hour to finish now.

https://github.com/aws-observability/aws-otel-collector/actions/runs/778860168


pingleig (approved):
LGTM

Though shouldn't we make this configurable? We can use 20 as default.

PR #293, opened 2021-04-29
Use SSM package version name start with 'v'

vastin:
The SSM package is published as v0.9.0. So don't trim prefix 'v' from version.


mxiamxia (approved):
Thanks!

PR #295, opened 2021-05-04
support local path template for validation

wyTrivail:
1.  add support for configuring local path data expected template as long as the template name start with "file://".  Ex. the yaml config could be 

```
-
  validationType: "cw-metric"
  httpPath: "/outgoing-http-call"
  httpMethod: "get"
  callingType: "http"
  expectedMetricTemplate: "file:///opt/aws/test.mustache"
```

2. Introduce mock lib so that we can add unit-tests gradually, which was a defect of validator.

3. reformat the code.

4. refactor the code a bit.


pingleig:
I think @bjrara should take a look because the metric validator logic has also changed a bit in this PR

validator/src/main/java/com/amazon/aoc/fileconfigs/PredefinedExpectedTemplate.java
```diff
@@ -15,18 +15,34 @@
...
-import lombok.Getter;
+import java.net.URL;
 
-@Getter
-public enum ExpectedTrace implements FileConfig {
+/**
+ * ExpectedTemplate includes all the built-in expected data templates,
+ * which are under resources/expected-data-templates.
+ */
+public enum PredefinedExpectedTemplate implements FileConfig {
```
bjrara, line 24: I would prefer merge `LocalPathExpectedTemplate` and `PredefinedExpectedTemplate` into one, and register the predefined templated into local. It looks like predefined is just a subset of local path.

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
```diff
@@ -0,0 +1,155 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * helper class for getting expected metrics from templates.
+ */
+public class CWMetricHelper {
```
bjrara, line 35: Is it possible to make the functions in Helper class static?

validator/src/main/java/com/amazon/aoc/helpers/MustacheHelper.java
```diff
@@ -38,14 +39,14 @@
    * @return generated content
    * @throws IOException when the template file is not existed
    */
-  public String render(FileConfig fileConfig, Object dataToInject) throws IOException {
+  public String render(FileConfig fileConfig, Object dataToInject) throws Exception {
     return render(fileConfig.getPath(), dataToInject);
   }
 
-  private String render(String path, Object dataToInject) throws IOException {
+  private String render(URL path, Object dataToInject) throws Exception {
```
bjrara: It's sad to see the exception is less specific. Any thoughts?

validator/src/main/java/com/amazon/aoc/validators/ContainerInsightPrometheusMetricsValidator.java
```diff
@@ -18,14 +19,20 @@
...
     List<CloudWatchContext.App> validateApps = getAppsToValidate(context.getCloudWatchContext());
     MustacheHelper mustacheHelper = new MustacheHelper();
     for (CloudWatchContext.App app : validateApps) {
-      String templateInput = mustacheHelper.render(new JsonSchemaFileConfig(
-              FilenameUtils.concat(templatePath,
-                      app.getName() + "_metrics.mustache")), context);
+      FileConfig fileConfig = new LocalPathExpectedTemplate(FilenameUtils.concat(
+          expectedDataTemplate.getPath().getProtocol()
+            + "://"
+            + expectedDataTemplate.getPath().getPath(),
```
bjrara: Can the transformation be simplified as `getPath().toString()`?

bjrara (commented):
Only some trivial questions. Overall looks good to me. Thanks for adding test cases for the validators!

validator/src/main/java/com/amazon/aoc/fileconfigs/PredefinedExpectedTemplate.java
wyTrivail: there're two differences driving me divide them into two.
1.  PredefinedExceptedTemplate uses different way to get its file path: `GetResource.GetPath()`, as its interface FileConfig defined, the function `getPath` is the key for different implementation.

2. PredefinedExcecptedTemplate is an enum class, i'm not sure how to merge them together, it will be great if you can shed me some lights?

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
wyTrivail: i wonder what's the benefit to make it static?  I tended to use non-static class/function as they are more friendly to unit-test(it's not easy to mock static functions/classes), but i'm open to change it if it's the best practice to use static here.

validator/src/main/java/com/amazon/aoc/helpers/MustacheHelper.java
wyTrivail: good point! we should try our best to have specific exceptions defined in most of the cases. will change it back.

validator/src/main/java/com/amazon/aoc/validators/ContainerInsightPrometheusMetricsValidator.java
wyTrivail: yes, we can, thanks for pointing out!

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
bjrara: Usually `static` could better explain that methods are stateless, it would also slightly improve the runtime performance, which won't make any difference here in the integration test... I don't think it worths much effort to change it to static.

validator/src/main/java/com/amazon/aoc/fileconfigs/PredefinedExpectedTemplate.java
bjrara: Get it. The pain is we need to convert a string to a built-in FileConfig. Let's keep it as what it is, merging two won't make the whole thing easier. Please help update the comment of `PredefinedExpectedTemplate`, I think we need to refine the class name in the java doc.

bjrara:
Please rebase the new main branch, thx!

bjrara (approved):
LGTM

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
wyTrivail: thanks! learned now :)

PR #297, opened 2021-05-05
Emit canary metrics for SSM package test

vastin:
Send success metrics for SSM package testcase in canary test.

### DIFF
--- a/terraform/ec2/main.tf
+++ b/terraform/ec2/main.tf
@@ -78,6 +78,7 @@ locals {
 
   # get SSM package version, latest is the default version
   ssm_package_version = var.aoc_version == "latest" ? "\"\"" : var.aoc_version
+  testcase_name       = split("/", var.testcase)[2]
 }
 
 ## launch a sidecar instance to install data emitter and the mocked server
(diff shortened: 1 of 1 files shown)


terraform/ec2/main.tf
```diff
@@ -477,6 +478,17 @@ resource "null_resource" "ssm_validation" {
   }
 }
 
+resource "null_resource" "ssm_canary_metrics" {
+  depends_on = [null_resource.ssm_validation]
+  count      = !var.skip_validation && var.enable_ssm_validate && var.canary ? 1 : 0
+
+  provisioner "local-exec" {
+    command = <<-EOT
+        aws cloudwatch put-metric-data --metric-name Success --dimensions testcase=${local.testcase_name} --namespace "Otel/Canary" --value 1.0
```
pingleig, line 487: Maybe we can send more detail info like how long it takes for ssm installation? Though that logic can be put into the python script.

terraform/ec2/main.tf
```diff
@@ -78,6 +78,7 @@ locals {
 
   # get SSM package version, latest is the default version
   ssm_package_version = var.aoc_version == "latest" ? "\"\"" : var.aoc_version
+  testcase_name       = split("/", var.testcase)[2]
```
pingleig, line 81: Is there a way to get the last element in list in terraform?

terraform/ec2/main.tf
vastin: SSM service has sent such metrics to owner of SSM package. This is only for canary test and will align with other canary testcases.

terraform/ec2/main.tf
vastin: Something like element(split("/", var.testcase), length(split("/", var.testcase))-1) should work. However the same statement is used to get testcase name by performance and soaking tests. Prefer to refactoring all of them in separate PR.

pingleig (approved):
LGTM

PR #309, opened 2021-06-21
Update validator readme with working example

NathanielRN:
# Description

Made the following updates to the readme:
- In my local tests I found that `--endpoint 127.0.0.1:8080` will **not work** while `--endpoint http://127.0.0.1:8080` _will_ work.
- It seems like there is no `default-validation.yml` file, so I replaced it with one that I know works for OTel Python at least.

### DIFF
--- a/validator/README.md
+++ b/validator/README.md
@@ -7,7 +7,7 @@ validator is one of the component of the testing framework, and can be run indep
 run this command in the root dir of the testing framework
 
 ```shell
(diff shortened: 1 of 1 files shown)


pingleig (approved):
The doc was outdated https://github.com/aws-observability/aws-otel-test-framework/blob/2c95cb76e080125f955bdf6734936a6a3979e0b3/validator/src/main/java/com/amazon/aoc/callers/HttpCaller.java#L39

PR #311, opened 2021-06-28
Fix EC2 AMI lookup filter issue

mxiamxia:
Fix EC2 AMI lookup filter issue and SUSE integ test cases

https://github.com/aws-observability/aws-otel-collector/runs/2936855704?check_suite_focus=true


pingleig (approved):
LGTM

PR #312, opened 2021-06-29
Fix the ec2 ami arch setup

mxiamxia:
Revert arch value back to amd64 and only apply x86 in ec2 lookup filter. 

Tested in local


pingleig (approved):
LGTM

(pretending I am not the one approving previous PR

PR #316, opened 2021-08-02
Add AMP Validator

KKelvinLo:
In this PR, there is a new static Prometheus metric validator. As we pull expected metrics from a provided mustache file, we want to verify those metrics against the actual metrics pulled from AMP(when given a query endpoint). (This mustache file is temporary and will be changed in the future).

This validator shares some similarity to the dynamic prometheus metric [validator](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/validator/src/main/java/com/amazon/aoc/validators/PrometheusMetricValidator.java) but since we do have an access to the timestamp of a metric time series we will be querying all metrics within the last hour. PromQL is not designed to be able to query all metrics without a timestamp hence the decision to query all metrics in the last hour was made.

### DIFF
--- a/validator/src/main/resources/validations/prometheus-static-metric-validation.yml
+++ b/validator/src/main/resources/validations/prometheus-static-metric-validation.yml
@@ -2,4 +2,3 @@
   validationType: "prom-static-metric"
   shouldValidateMetricValue: false
   expectedMetricTemplate: "AMP_EXPECTED_METRIC"
-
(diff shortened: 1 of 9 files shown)


validator/src/main/resources/validations/prometheus-static-metric-validation.yml
```diff
@@ -0,0 +1,4 @@
+-
```
mxiamxia, line 1: do we need to retire `prometheus-sd-validation.yml` which is used to validate AMP exporter test case?

KKelvinLo: If the current prometheus-sd-validation.yml is still used to validate AMP exporter with the current sample-apps setup, we should not retire it.

KKelvinLo: This validator will also be used for validating the metrics coming in from aws-otel-lambda.

mxiamxia (approved):

PR #320, opened 2021-08-13
Fix For Gradle 7

sethAmazon:
Tested by building the java sample apps with Java 11 and gradle 7


sethAmazon:
@mxiamxia should I also increase the wrapper to 7?

mxiamxia (approved):
thx

PR #330, opened 2021-08-24
Updated soaking test load generator image.

jefchien:
Pushed new image to public ECR.


mxiamxia (approved):
nice. thx for the fixings.

PR #331, opened 2021-08-25
Disable Counter metrics in Prom Sample App

mxiamxia:
Disable Counter metrics in Prom Sample App Since there was some breaking changes in Prom receiver previously. Will need to follow up with AMP team.


jefchien (approved):
LGTM

PR #347, opened 2021-09-21
Fix Operator test case to use the AOC image form the latest commit

mxiamxia:
Make Operator test case to use the AOC image built from the latest commit so it can verify the breaking changes.


willarmiros (approved):
lgtm

Aneurysm9 (approved):

PR #381, opened 2021-11-01
Reduced number of retries for soaking test alarm validation

jefchien:
There are multiple soaking tests that are cancelled before completion (ex: [run-soaking-test (statsd, soaking_linux)](https://github.com/aws-observability/aws-otel-collector/runs/4054558138?check_suite_focus=true)). This is because the jobs exceed the 6 hour limit.

https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits
> Job execution time - Each job in a workflow can run for up to 6 hours of execution time. If a job reaches this limit, the job is terminated and fails to complete.

Reducing the number of retries for the test should reduce the likelihood of timeout.

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/services/CloudWatchAlarmService.java
+++ b/validator/src/main/java/com/amazon/aoc/services/CloudWatchAlarmService.java
@@ -21,8 +21,8 @@ public class CloudWatchAlarmService {
    * @return the list of MetricAlarm Object
    */
   public List<MetricAlarm> listAlarms(List<String> alarmNameList) {
-    DescribeAlarmsResult describeALarmsResult =
+    DescribeAlarmsResult describeAlarmsResult =
         amazonCloudWatch.describeAlarms(new DescribeAlarmsRequest().withAlarmNames(alarmNameList));
-    return describeALarmsResult.getMetricAlarms();
+    return describeAlarmsResult.getMetricAlarms();
   }
 }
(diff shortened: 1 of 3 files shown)


validator/src/main/java/com/amazon/aoc/services/CloudWatchAlarmService.java
```diff
@@ -21,8 +21,8 @@ public CloudWatchAlarmService(String region) {
    * @return the list of MetricAlarm Object
    */
   public List<MetricAlarm> listAlarms(List<String> alarmNameList) {
-    DescribeAlarmsResult describeALarmsResult =
+    DescribeAlarmsResult describeAlarmsResult =
```
sethAmazon, line 24: Good catch

validator/src/main/java/com/amazon/aoc/validators/AlarmPullingValidator.java
```diff
@@ -60,7 +60,7 @@ public void validate() throws Exception {
           alarmList.sort(Comparator.comparing(MetricAlarm::getAlarmName));
           for (int i = 0; i != context.getAlarmNameList().size(); ++i) {
             if (!context.getAlarmNameList().get(i).equals(alarmList.get(i).getAlarmName())) {
-              log.error("alarm {} can not be found", context.getAlarmNameList().get(i));
+              log.error("alarm {} cannot be found", context.getAlarmNameList().get(i));
```
sethAmazon, line 63: can't?

sethAmazon (approved):

PR #384, opened 2021-11-03
Fixing soak test with insufficient IP address on subnet

khanhntd:
**Relate issues**: https://github.com/aws-observability/aws-otel-collector/issues/698
**Short term fix:** change subnet public region a to subnet public region a-1
**Long term fix:** 
- Delete the unused EC2 with the associate network interfaces to free the IP address in one subnet
- Let's the terraform destroy the EC2 resources even failing (relating to state but still trying to figure that out)

Note: Already created another subnet that has the same route table,... as the previous one
Screen Shot 2021-11-03 at 4 37 10 PM

### DIFF
--- a/terraform/basic_components/main.tf
+++ b/terraform/basic_components/main.tf
@@ -54,7 +54,7 @@ data "aws_subnet_ids" "aoc_public_subnet_ids" {
   filter {
     name = "tag:Name"
     values = [
-      "${module.common.aoc_vpc_name}-public-${var.region}a",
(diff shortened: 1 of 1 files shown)


aateeqi:
How do we know that this is working?

khanhntd:
> How do we know that this is working?

I would say that waiting for the current soak test to finish and re-run it again. If it does not work out, we could revert back again.
Since there are two options for us: delete the current subnet and make a same-subnet-name with the associate EC2 but it's not available for now or we could replace with the substitute one and only this files are the one associate with the subnet.

PR #395, opened 2021-11-10
Refactored nodejs mock_servers to go

### LINKED ISSUE #393: Refactor mock servers to use golang
The node servers may not be performant, which is affecting our benchmarking tests.

jefchien:
**Description:** Refactored mock servers. Added Makefile to generate definitions from protos.

**Link to tracking Issue:** https://github.com/aws-observability/aws-otel-test-framework/issues/393

Will need to push the images to ECR once merged.


mocked_servers/Makefile
```diff
@@ -0,0 +1,17 @@
+ADDLICENSE=$(PWD)/bin/addlicense
+
+.PHONY: install-tools
+install-tools:
```
anuraaga: Think it's better to use the OTel docker build tool

https://github.com/open-telemetry/build-tools/blob/main/protobuf/README.md

mocked_servers/grpc_metrics/Dockerfile
```diff
@@ -1,11 +1,17 @@
...
+RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux go build -o=/bin/main .
 
-COPY *.js ./
-COPY package.json ./
-COPY metrichandler.proto ./
-
-RUN npm install
-
-CMD ["node", "metrics_handler_server.js"]
+FROM scratch
```
anuraaga, line 8: Recommend `gcr.io/distroless/static` in case the lack of TLS certificates becomes an issue even in a mocked server in the future since the errors can be very mysterious

mocked_servers/grpc_metrics/Dockerfile
```diff
@@ -1,11 +1,17 @@
-FROM node:14
+FROM golang:1.17 AS mod
```
Aneurysm9: Is it necessary to have a separate stage for this?  What advantage does it provide?

mocked_servers/grpc_metrics/Dockerfile
Aneurysm9: Perhaps even `gcr.io/distroless/static:nonroot` since we shouldn't need to be running as root.

mocked_servers/Makefile
jefchien: Ended up just using https://github.com/open-telemetry/opentelemetry-proto-go instead as suggested by @Aneurysm9

mocked_servers/grpc_trace/main.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (h *checkDataHandler) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
+	if _, err := w.Write([]byte(*h.data)); err != nil {
+		log.Printf("Unable to write response: %v", err)
+	}
+}
+
+// Export Implements the RPC method.
+func (s *server) Export(_ context.Context, _ *pb.ExportTraceServiceRequest) (*pb.ExportTraceServiceResponse, error) {
```
Aneurysm9: Does this require a completely separate implementation?  This looks almost identical to the metrics implementation.

Aneurysm9 (approved):

mocked_servers/grpc_trace/main.go
jefchien: It's identical just uses different protos.

mocked_servers/grpc_metrics/Dockerfile
jefchien: I copied it from the prometheus sample app https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/sample-apps/prometheus/Dockerfile

I don't think it's necessary to have the separate stage.

mocked_servers/grpc_metrics/Dockerfile
jefchien: Once gcr.io use has been verified, I can make the changes.

.gitignore
```diff
@@ -59,4 +58,6 @@ terraform.rc
 !gradle/wrapper/gradle-wrapper.jar
 
 # Ignore nodejs builds
-node_modules 
+node_modules
```
sethAmazon: Is this still needed?

jefchien: It's not. Good catch!

sethAmazon:
How have you tested this commit?

jefchien:
Tested it locally against `otlp_*_exporter_*_mock`, `otlp_mock`, `sapm_exporter_trace_mock`, `signalfx_exporter_metric_mock`, `zipkin_mock`

`prometheus_mock` fails, but this occurs before the changes. Created an issue.
Link to tracking issue: https://github.com/aws-observability/aws-otel-test-framework/issues/404

mocked_servers/https/Dockerfile
```diff
@@ -1,11 +1,9 @@
...
-COPY certificates ./certificates
-COPY *.js ./
-COPY package.json ./
-
-RUN npm install
-
-CMD ["node", "index.js"]
+FROM golang:1.17 as build
+WORKDIR $GOPATH/main
+COPY . .
```
sethAmazon, line 3: Is the entire directory needed or do you only need to copy certain files?

sethAmazon: I would assume yes on this.

sethAmazon (approved):

mocked_servers/https/Dockerfile
jefchien: There's only 4 files in the directory to begin with. I guess I don't need to pull in the Dockerfile, but I don't think that'll make much of a difference.

mocked_servers/https/main.go
```diff
@@ -0,0 +1,123 @@
...
+		log.Printf("Unable to write response: %v", err)
+	}
+}
+
+func (h *transactionStore) dataReceived(w http.ResponseWriter, _ *http.Request) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	h.data = SuccessMessage
+	h.transactions++
```
Aneurysm9:
```suggestion
	h.mu.Lock()

	h.data = SuccessMessage
	h.transactions++
	
	h.mu.Unlock()
```

No need to retain the lock while sleeping.  I suspect that the lock could be removed entirely using [`sync/atomic`](https://pkg.go.dev/sync/atomic) types.  `h.data` is basically a boolean indicating whether data was ever received and thus can be replaced with `h.transactions > 0`.  That can be addressed in a followup PR if throughput becomes an issue.

mocked_servers/https/main.go
```diff
@@ -0,0 +1,123 @@
...
+
+// Retrieve number of transactions per minute
+func (h *transactionStore) tpm(w http.ResponseWriter, _ *http.Request) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	// Calculate duration in minutes
+	now := time.Now()
+	duration := now.Sub(h.startTime)
+	tpm := h.transactions / int(duration.Minutes())
```
Aneurysm9:
```suggestion
	// Calculate duration in minutes
	now := time.Now()
	duration := now.Sub(h.startTime)
	
	h.mu.Lock()
	tpm := h.transactions / int(duration.Minutes())
	h.mu.Unlock()
```

Again, minimize lock contention where you can.

mocked_servers/https/main.go
```diff
@@ -0,0 +1,123 @@
...
+
+// Starts an HTTPS server that receives requests for the data handler service at the sample server port
+// Starts an HTTP server that receives request from validator only to verify the data ingestion
+func main() {
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	store := transactionStore{
+		data:         "",
+		transactions: 0,
```
Aneurysm9:
```suggestion
```

These are the zero values for their types, no need to explicitly initialize them.

Aneurysm9 (approved):
A few minor suggestions on lock usage and style.

mocked_servers/https/main.go
jefchien: Switched to using `AddUint32`, `LoadUint32`, and `StoreUint32` instead of locking.

PR #396, opened 2021-11-12
Added Makefile to handle certificate updates.

jefchien:
Re-ran certificate generation with openssl.conf


mocked_servers/https/README.md
```diff
@@ -0,0 +1,5 @@
+# HTTPS Server
```
sethAmazon, line 1: Please add a comment saying open ssl must be installed without sudo requirement for this to work. At least I think you will need to install open ssl for this to work.

sethAmazon (approved):

aateeqi (approved):

PR #403, opened 2021-11-15
Add Test Case Mock For Otlp To AWS Prometheus Remote Write

sethAmazon:


terraform/testcases/otlp_metric_amp_mock/otconfig.tpl
```diff
@@ -0,0 +1,30 @@
+extensions:
+  pprof:
+    endpoint: 0.0.0.0:1777
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:${grpc_port}
```
aateeqi, line 8: is there a default for this grpc_port value or does it have to be specified?

sethAmazon: There is a default

sethAmazon:
Tested local with command 
terraform apply -auto-approve -lock=false -var-file="../testcases/otlp_metric_amp_mock/parameters.tfvars" -var="testcase=../testcases/otlp_metric_amp_mock"

terraform/testcases/otlp_metric_amp_mock/otconfig.tpl
sethAmazon: 4317

aateeqi (approved):

validator/src/main/resources/validations/otlp-mocked-server-prometheus-validation.yml
```diff
@@ -0,0 +1,5 @@
+-
```
jefchien: This looks just like the `default-mocked-server-validation.yml`

terraform/testcases/otlp_metric_amp_mock/otconfig.tpl
```diff
@@ -0,0 +1,30 @@
...
+
+processors:
+  batch:
+
+exporters:
+  logging:
+    loglevel: info
```
jefchien: Most of the other tests set the logging level to debug.

sethAmazon: The other tests do not run the log exporter. Look at the exports portion of the test

validator/src/main/resources/validations/otlp-mocked-server-prometheus-validation.yml
sethAmazon: Yes it is the same since mock server tests do not validate data, it only validates if data is sent to the mock server do you think I should use that yml file?

jefchien: What's the difference between these files? https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/validator/src/main/resources/validations/default-mocked-server-validation.yml

jefchien: If they're the same, then I don't see a reason to make another one.

terraform/testcases/otlp_metric_amp_mock/otconfig.tpl
sethAmazon: I can change this to what other tests do if you think I should

jefchien: https://github.com/aws-observability/aws-otel-test-framework/blob/a61d96e03129b20f252ab41b256de6faf5ffa540/terraform/testcases/otlp_metric/otconfig.tpl#L13-L24

jefchien: It's more of a nitpick. All of the other tests use debug.

PR #405, opened 2021-11-16
Add Otlp To AMP Test

sethAmazon:


validator/src/main/resources/expected-data-template/otlpToAMPExpectedMetric.mustache
```diff
@@ -0,0 +1,18 @@
+[
+  {
+    "metric": {
```
sethAmazon, line 3: This has to do with when there is one metric it will set the quantile as 0 and 1. This is really 1 metric

validator/src/main/resources/validations/otlp-prometheus-validation.yml
```diff
@@ -0,0 +1,7 @@
+-
+  validationType: "prom-static-metric"
+  httpPath: "/outgoing-http-call"
+  httpMethod: "get"
+  callingType: "http"
+  shouldValidateMetricValue: false
```
sethAmazon, line 6: Values can be random since we are testing a latency metric. It is not a static value.

validator/src/main/java/com/amazon/aoc/validators/PrometheusStaticMetricValidator.java
```diff
@@ -46,9 +46,15 @@
   private FileConfig expectedMetric;
   private ValidationConfig validationConfig;
   private CortexService cortexService;
+  private ICaller caller;
 
   @Override
   public void validate() throws Exception {
+    // hit the endpoint to generate data if need be
```
sethAmazon, line 53: The spark app sends out the otlp metrics. Thus call the generator is this step.

sethAmazon:
Tested via 
terraform apply -auto-approve -lock=false -var-file="../testcases/otlp_metric_amp/parameters.tfvars" -var="testing_ami=amazonlinux" -var="aoc_version=v0.14.0-1450694987" -var="testcase=../testcases/otlp_metric_amp"

jefchien (approved):
LGTM

khanhntd:
LGTM

PR #411, opened 2021-11-26
Add Initial Support For EKS Fargate Testing In Terraform

sethAmazon:


docs/run-testing-framework.md
```diff
@@ -86,10 +86,46 @@ cd terraform/eks && terraform init && terraform apply \
     -var="testcase=../testcases/{{your test case folder name}}" \
     -var-file="../testcases/{{your test case folder name}}/parameters.tfvars"
 ````
- 
+
+### 2.3.1 Run in EKS Fargate
+#### Set Up
+* Install kubectl and eksctl
```
anuraaga, line 92: These should be provided as a terraform config for making it easier to run. If the reason to not include the k8s setup in the test tf files is the time it takes to provision a cluster (it is indeed best practice to decouple cluster creation and use) it's still good to provide these as a separate terraform deployment instead of manual steps.

anuraaga: Sorry read your doc now and realized that this is your plan - that being said, it might be easier to achieve it if this is a separate installation step, but using terraform.

sethAmazon: The current functionality for eks testing is we do not bring up an eks cluster.

terraform/eks/main.tf
```diff
@@ -152,7 +158,7 @@ module "validator" {
   region                       = var.region
   testing_id                   = module.common.testing_id
   metric_namespace             = "${module.common.otel_service_namespace}/${module.common.otel_service_name}"
-  sample_app_endpoint          = length(kubernetes_service.sample_app_service) > 0 ? "http://${kubernetes_service.sample_app_service.0.load_balancer_ingress.0.hostname}:${module.common.sample_app_lb_port}" : ""
+  sample_app_endpoint          = length(kubernetes_ingress.app) > 0 && var.deployment_type == "fargate" ? "http://${kubernetes_ingress.app.0.load_balancer_ingress.0.hostname}:80" : length(kubernetes_ingress.app) > 0 ? "http://${kubernetes_service.sample_app_service.0.load_balancer_ingress.0.hostname}:${module.common.sample_app_lb_port}" : ""
```
aateeqi, line 161: Could we use parentheses to clarify the ternary statements or could it be rewritten like this?
```suggestion
  sample_app_endpoint          = length(kubernetes_ingress.app) > 0 ? (var.deployment_type == "fargate" ? "http://${kubernetes_ingress.app.0.load_balancer_ingress.0.hostname}:80" : "http://${kubernetes_service.sample_app_service.0.load_balancer_ingress.0.hostname}:${module.common.sample_app_lb_port}") :  ""
```

docs/run-testing-framework.md
anuraaga: Yeah but here we're documenting complicated steps for starting up a cluster - instead we should provide a simple way to do it, with just a single terraform command. And it should make it easier in the future to move toward automatically bringing up the cluster in tests too. I don't think it's worth adding these complicated steps now instead of modeling them with terraform (a separate deployment, not part of the test, for now)

sethAmazon:
I will open a new pr going to close this

PR #412, opened 2021-11-28
Support testcase for ec2 based on ecs when using otlp to send metrics to amprometheus xray cloudwatch

khanhntd:
**Use case:**
* Can replace the CD test cases since the test case CD only need to make sure the integrations with backends are working instead of the same as CI, while baking one plan per config
* Support the validation for the ECS EC2 console.

**How to test it**:
```cd terraform/ecs  && terraform init && terraform apply -auto-approve -var="testcase=../testcases/otlp_grpc_exporter_cw_amp_xray_ecs" -var-file="../testcases/otlp_grpc_exporter_cw_amp_xray_ecs/parameters.tfvars"```

**Expected result**:
Validation is passed

### DIFF
--- a/terraform/common.tf
+++ b/terraform/common.tf
@@ -31,7 +31,7 @@ variable "sample_app_image" {
 }
 
 variable "aoc_image_repo" {
-  default = "public.ecr.aws/aws-observability/aws-otel-collector"
+  default = "611364707713.dkr.ecr.us-west-2.amazonaws.com/aws/aws-otel-collector"
 }
 
 variable "aoc_version" {
--- a/terraform/common.tf
+++ b/terraform/common.tf
@@ -31,7 +31,7 @@ variable "sample_app_image" {
 }
 
 variable "aoc_image_repo" {
-  default = "611364707713.dkr.ecr.us-west-2.amazonaws.com/aws/aws-otel-collector"
+  default = "public.ecr.aws/aws-observability/aws-otel-collector"
 }
 
 variable "aoc_version" {
--- a/terraform/ecs/main.tf
+++ b/terraform/ecs/main.tf
@@ -328,6 +328,7 @@ module "validator" {
   account_id = data.aws_caller_identity.current.account_id
 
   ecs_context_json = jsonencode({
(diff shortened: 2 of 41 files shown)

### CODE REVIEW

File: terraform/common.tf [hunk 1]
```diff
@@ -31,7 +31,7 @@ variable "sample_app_image" {
 }
 
 variable "aoc_image_repo" {
-  default = "611364707713.dkr.ecr.us-west-2.amazonaws.com/aws/aws-otel-collector"
+  default = "public.ecr.aws/aws-observability/aws-otel-collector"
```
khanhntd: One of the reason I changed this is because the `private repo image does not have the latest tag` except the public one. Therefore, we could change this in this PR or another PR if that make sense.

terraform/ecs/main.tf
```diff
@@ -328,6 +328,7 @@ module "validator" {
   account_id = data.aws_caller_identity.current.account_id
 
   ecs_context_json = jsonencode({
+    ecsClusterName : module.ecs_cluster.cluster_name
```
khanhntd, line 331: One of the reason I added this is because the [expected metric template ](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/validator/src/main/resources/expected-data-template/ecsContainerExpectedMetric.mustache#L2-L7) needs to `get the metrics from the container context`. However, the current context does not have that. Therefore, `it returns NullPointerException whenever variable sample_app_callable return true`.

terraform/templates/defaults/ecs_taskdef.tpl
```diff
@@ -5,11 +5,11 @@
       "cpu": 10,
```
sethAmazon: Can we have the reformat in a diff pr?

sethAmazon (approved):

terraform/templates/defaults/ecs_taskdef.tpl
khanhntd: Yes, we can. Fixed it.

Aneurysm9 (approved):

anuraaga (approved):

PR #425, opened 2021-12-06
Add Initial Support For EKS Fargate Testing In Terraform

sethAmazon:


terraform/validation/main.tf
```diff
@@ -49,6 +49,7 @@ data "template_file" "docker_compose" {
     incoming_packets_alarm = var.incoming_packets_alarm
 
     cortex_instance_endpoint = var.cortex_instance_endpoint
+    rollup                   = var.rollup
```
khanhntd, line 52: What does roll-up for? Any documents on that since I could not fine one.

rltoSD: @khanhntd  Hi, rollup is a new feature for cwmetricsvalidator to enable or disable dimension rollup for expected test templates. By default it was set to true earlier, but with this newly implemented feature we can also turn it off.

terraform/eks/main.tf
```diff
@@ -152,7 +167,7 @@ module "validator" {
   region                       = var.region
   testing_id                   = module.common.testing_id
   metric_namespace             = "${module.common.otel_service_namespace}/${module.common.otel_service_name}"
-  sample_app_endpoint          = length(kubernetes_service.sample_app_service) > 0 ? "http://${kubernetes_service.sample_app_service.0.load_balancer_ingress.0.hostname}:${module.common.sample_app_lb_port}" : ""
+  sample_app_endpoint          = length(kubernetes_ingress.app) > 0 && var.deployment_type == "fargate" ? "http://${kubernetes_ingress.app.0.load_balancer_ingress.0.hostname}:80" : length(kubernetes_service.sample_app_service) > 0 ? "http://${kubernetes_service.sample_app_service.0.load_balancer_ingress.0.hostname}:${module.common.sample_app_lb_port}" : ""
```
khanhntd: For this port, can we set as a variable port instead of default one?

terraform/eks/main.tf
```diff
@@ -108,6 +111,18 @@ resource "kubernetes_service_account" "aoc-role" {
   automount_service_account_token = true
 }
 
+resource "kubernetes_service_account" "aoc-fargate-role" {
+  metadata {
+    name      = "aoc-fargate-role-${module.common.testing_id}"
+    namespace = "default"
+    annotations = {
+      "eks.amazonaws.com/role-arn" : "arn:aws:iam::611364707713:role/ServiceAccount-eks-test-aoc-role"
```
khanhntd: Can we set the account Id to a dynamic one instead of default one?

.gitignore
```diff
@@ -58,3 +58,5 @@ override.tf.json
 terraform.rc
 
 !gradle/wrapper/gradle-wrapper.jar
+
+*/.DS_Store
```
jefchien:
```suggestion
.DS_Store
**/.DS_Store
```
https://git-scm.com/docs/gitignore
>A leading "**" followed by a slash means match in all directories. For example, "**/foo" matches file or directory "foo" anywhere, the same as pattern "foo".

docs/game.yml
```diff
@@ -0,0 +1,59 @@
+---
```
jefchien: I don't think we need to duplicate this file in our repo. We can probably just link to the example in our docs. https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/examples/2048/2048_full.yaml

terraform/eks/main.tf
sethAmazon: This I was going to change in the next phase.

khanhntd: Got cha. Should  catch up with the plan for the current phase.

terraform/eks/container-insights-agent/aoc_fargate_deploy.yml
```diff
@@ -0,0 +1,557 @@
...
+kind: ConfigMap
+metadata:
+  name: adot-collector-config
+  namespace: default
+  labels:
+    app: aws-adot
+    component: adot-collector-config
+data:
+  adot-collector-config: |
+    receivers:
```
jefchien: Are we using a fixed config for each of the tests? Can we use a template and render setup?

terraform/eks/container-insights-agent/config_map_fargate.yml
```diff
@@ -0,0 +1,442 @@
...
+kind: ConfigMap
+metadata:
+  name: adot-collector-config
+  namespace: default
+  labels:
+    app: aws-adot
+    component: adot-collector-config
+data:
+  adot-collector-config: |
+    receivers:
```
jefchien, line 11: Is this the same config as in `aoc_fargate_deploy.yml`?

terraform/validation/main.tf
khanhntd: Reasonable. Thanks for explaining.

khanhntd (approved):
LGTM.

terraform/eks/main.tf
jefchien: It's a bit difficult to parse nested ternary statements. Can we use parentheses to break it up?

docs/game.yml
sethAmazon: That link does not use the default namespace. It uses another namespace.

terraform/eks/container-insights-agent/aoc_fargate_deploy.yml
sethAmazon: I can add that in later commits. So we can get the testing in.

docs/game.yml
jefchien: Is that the only difference?

terraform/eks/container-insights-agent/config_map_fargate.yml
sethAmazon: Yes the first file contains the fill config. I can delete it. Sorry I was using it as a reference.

docs/game.yml
sethAmazon: yes

sethAmazon: Eventually this will be removed when I get the cluster creation fully working.

jefchien: So, could you explain that in the doc instead of adding the file? Like
> * Create Ingress controller fargate
Download config [here](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/examples/2048/2048_full.yaml) and replace all instances of namespace `game-2048` with `default`.

jefchien (approved):

terraform/eks/container-insights-agent/logs_sample_fargate.yml
```diff
@@ -0,0 +1,34 @@
...
+      app: flblogproducer2eks
+  template:
+    metadata:
+      labels:
+        app: flblogproducer2eks
+    spec:
+      containers:
+        - name: flblogproducer2eks
+          #image: 803860917211.dkr.ecr.us-west-2.amazonaws.com/alpine-log-producer:latest
+          image: 803860917211.dkr.ecr.us-west-2.amazonaws.com/python-logger:300sec
```
hossain-rayhan, line 21: This image is in a personal ECR account. Let's replace it with `nginx:latest` and remove the command section.

terraform/eks/container-insights-agent/logs_sample_fargate.yml
```diff
@@ -0,0 +1,34 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: flblogproducer2eks
```
hossain-rayhan, line 4: Let's rename this to `app-container` in 5 places. This is not related to Fluent Bit. It was added only for testing purposes.

PR #426, opened 2021-12-08
Dynamic Assume Role Creation

sethAmazon:


terraform/eks/main.tf
```diff
@@ -111,16 +111,63 @@ resource "kubernetes_service_account" "aoc-role" {
...
+      Version = "2012-10-17"
+      Statement = [
+        {
+          Action = [
+            "ec2:*",
+            "logs:*",
+            "cloudwatch:PutMetricData",
+            "s3:*",
+            "ssm:*",
+            "xray:*"
```
aateeqi: Are these the final state of the policy or will end up going back and specify them out?

sethAmazon: We can specify them out later. I think this should be fine because we only create the policy for the test then we tear it down after test execution so a wild card policy is not floating around.

sethAmazon: Considering we use a role with an admin policy currently this is really stripped down :) aoc-integ-test-dev

jefchien (approved):
LGTM

aateeqi (approved):

PR #437, opened 2021-12-14
Only Create Fargate Profile On Deployed Fargate Task

sethAmazon:


terraform/eks/main.tf
```diff
@@ -160,9 +161,10 @@ resource "kubernetes_service_account" "aoc-role" {
 }
 
 resource "kubernetes_service_account" "aoc-fargate-role" {
+  count = 1
```
jefchien: Shouldn't this also only be created on fargate deployments? It seems fargate specific.

jefchien: Should've refreshed.

jefchien:
Have you tested this?

khanhntd (approved):
LGTM.

jefchien (approved):

sethAmazon:
Yes I was able to bring up terraform in multiple repo. I was not able to run multiple validators since the validator will recreate when run on the same docker network. Our ci runs on multiple docker networks for containers will not be re created.

PR #438, opened 2021-12-15
Avoid having null pointers error when the context is missing

khanhntd:
**Description:**
Sometimes when the [ecs_context is missing or not being added](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/ecs/main.tf#L330-L335), the metrics getting from the container will miss some dimension value (default is empty). Therefore, to avoid having NullPointerException error, we need to add a condition for skipping these dimensions to make it fail faster and easier to know why these metrics are failing.

**How to test**:
For ecs test case such as ecsmetrics, define `sample_app_callable = true` and run the test case on ecs to see the error.
**Example:**  cd terraform/ecs && terraform init && terraform apply -var="testcase=../testcases/ecsmetrics" -var-file="../testcases/ecsmetrics/parameters.tfvars" -auto-approve

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
+++ b/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
@@ -74,7 +74,7 @@ public class CWMetricValidator implements IValidator {
     for (Metric metric : expectedMetricList) {
       for (Dimension dimension : metric.getDimensions()) {
 
-        if (Object.equals(dimension.getValue(), null) || dimension.getValue().equals("")) {
+        if (dimension.getValue() == null || dimension.getValue().equals("")) {
           continue;
         }
 
--- a/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
+++ b/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
@@ -74,7 +74,7 @@ public class CWMetricValidator implements IValidator {
     for (Metric metric : expectedMetricList) {
       for (Dimension dimension : metric.getDimensions()) {
 
-        if (dimension.getValue() == null || dimension.getValue().equals("")) {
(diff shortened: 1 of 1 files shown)


validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
```diff
@@ -73,6 +73,11 @@ public void validate() throws Exception {
     Set<String> skippedDimensionNameList = new HashSet<>();
     for (Metric metric : expectedMetricList) {
       for (Dimension dimension : metric.getDimensions()) {
+
+        if (dimension.getValue() == null || dimension.getValue().equals("")) {
```
sethAmazon, line 77: Might want to look into https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html otherwise this is fine with me

sethAmazon (approved):

validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java
khanhntd: Looked through it. The equals will have problem [when the reference object is null](https://stackoverflow.com/questions/4501061/java-null-check-why-use-instead-of-equals/4501084#4501084) so I cannot use that. Therefore, it makes more sense for me to use [Object.equals ](https://stackoverflow.com/a/44799807) instead and I have already tested it out without any issue.

khanhntd: Seems it is not feasible to use Object.equals since the PR build shows the errors https://github.com/aws-observability/aws-otel-test-framework/runs/4539943758?check_suite_focus=true (Cannot compare null with string )

aateeqi (approved):

PR #449, opened 2021-12-21
Uploading terraform state to s3 to easier in knowing the resources which were created by the test case and destroying

khanhntd:
IMO, there are three ways for us to do this: 
1.  Using backend with s3 config   but hard part is we cannot use dynamic variable such as [the testing-id to define the key path](https://www.terraform.io/language/settings/backends/configuration#using-a-backend-block) since the terraform checkouts the backend first before any terraform resources.
 2.  Using ```aws_s3_bucket_object``` but need to create another folder since ```aws_s3_bucket_object``` can only upload complete file instead of updating files. Therefore, if we upload the resources within the platform folder, we would need to run it twice.
 3.  Uploading all the scripts to s3 after running the terraform apply  and [aws-otel-collector is using this](https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/soaking.yml#L188-L194) but we need to manage every resources within terraform to easier destroy and manage. 
 Currently, I a**m applying  the second option** since the second option can also be used as a managed folder.

**How to test it?**
An example would be:
- **Step 1:** Running the following command to run the test case: ```cd terraform/ec2 && terraform init && terraform apply -var="aoc_version=v0.15.0" -var="testcase=../testcases/otlp_trace_resourcedetection_ec2" -var-file="../testcases/otlp_trace_resourcedetection_ec2/parameters.tfvars" -var="package_s3_bucket=aws-otel-collector" -var="ssm_package_name=AWSDistroOTel-Collector" --auto-approve```
 
- **Step 2:** Running the following command to upload the test case ```cd terraform/add_on/remote_state && terraform init && terraform apply -var="testcase=../../testcases/otlp_trace_resourcedetection_ec2" -var="testing_id=302073711122da11" -var="platform=ec2"```

- **Step 3:** Only need to download the terraform state to the appropriate platform and destroy the resources within AWS account and use ``` terraform init && terraform destroy ```


anuraaga (approved):
This seems fine to me

alolita:
@khanhntd pl resolve conflicts, rerun tests before merging.

khanhntd:
Mostly the **terraform format's error is from defining type string and description** within a variable.

anuraaga (approved):

PR #459, opened 2021-12-28
Support uploading setup's terraform state to s3 bucket 

khanhntd:
**Description**
Currently, if devs shared the accounts with others and setup again, they will run into some issues when creating resources such as the Elastic IP Address problem. Therefore, it would be better to share the terraform state with one-time setup.
- Share the setup configuration with others dev
- Avoid creating duplicate resources when sharing the same account

**Disadvantages: **
- Instead of dynamic variables, this approach is supporting only hard code variables since [https://www.terraform.io/language/settings/backends/configuration#using-a-backend-block]
- We need to un-comment backend manually and run terraform apply again since we cannot control the creation of backend (terraform checkout backend before any modules, resources,..)

**How to test**:
**Step 1:** Following the instructions in https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/docs/setup-basic-components-in-aws-account.md
**Step 2:** Uncomment the [backend configuration](https://github.com/khanhntd/aws-otel-test-framework/blob/support_s3_bucket_setup/terraform/setup/backend.tf#L17-L25) 
**Step 3:** Run terraform init && terraform apply

### DIFF
--- a/terraform/setup/backend.tf
+++ b/terraform/setup/backend.tf
@@ -13,7 +13,10 @@
 # permissions and limitations under the License.
 # -------------------------------------------------------------------------
 
-#In this module, we are uploading
+#In this module, we are uploading setup's terraform state to s3 bucket for two reasons:
+#-Share the setup configuration with others dev
+#-Avoid creating duplicate resources when sharing the same account
+
 #terraform {
 #  backend "s3" {
 #    bucket           = "setup-remote-state-s3-bucket"
--- a/terraform/setup/backend.tf
+++ b/terraform/setup/backend.tf
@@ -1,0 +1,62 @@
+# ------------------------------------------------------------------------
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+#In this module, we are uploading
+#terraform {
+#  backend "s3" {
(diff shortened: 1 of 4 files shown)


terraform/setup/backend.tf
```diff
@@ -0,0 +1,65 @@
...
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+#In this module, we are uploading setup's terraform state to s3 bucket for two reasons:
+#-Share the setup configuration with others dev
+#-Avoid creating duplicate resources when sharing the same account
+
+#terraform {
```
sethAmazon, line 20: What are the comments for?

khanhntd: The purpose of terraform backend s3 is to `uploading  terraform.tfstate to s3.` However, since the backend is not supporting condition checking with S3 bucket and DynamoDB Table currently; so first time dev setup which does not have DynamoDB Table and S3 bucket on their aws account will not able to pass the error. Therefore, it would be good for us to treat this as optional feature to upload it and let's the dev decide to use it or not. Moreover, I have documented it out for everyone to know more: https://github.com/aws-observability/aws-otel-test-framework/pull/459/files#diff-0776469b54a50b9aaf826b5d5b9d11f4f34df0af615498fe9222343fdd9d8d33R26-R35

sethAmazon: Maybe this can be added to setup?

khanhntd: Currently, it is in the setup. My intention was uploading the terraform setup state file but currently found no way to upload it besides ```terraform apply``` a second time though.

sethAmazon:
Can you give a demo on this. Just like take a video of this feature being used and post it here.

khanhntd:
https://user-images.githubusercontent.com/91758108/149237590-defc09f8-14f8-47dd-ad76-f8fdf4254310.mp4

Yes. Here is the demo for uploading the terraform setup state to s3 bucket. I assume we have setup based on[ this document](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/docs/setup-basic-components-in-aws-account.md) since it would cost more time for the video to setup once again.

anuraaga (approved):

PR #460, opened 2021-12-28
Fix Java Hash Map Concurrent Modification Exception For EKS Container…

sethAmazon:
… Insights Log Valiator

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/validators/AbstractStructuredLogValidator.java
+++ b/validator/src/main/java/com/amazon/aoc/validators/AbstractStructuredLogValidator.java
@@ -51,6 +51,7 @@ public abstract class AbstractStructuredLogValidator implements IValidator {
   private static final int CHECK_DURATION_IN_SECONDS = 2 * 60;
 
   protected Map<String, JsonSchema> schemasToValidate = new HashMap<>();
+  protected Set<String> validatedSchema = new HashSet<>();
   protected Set<String> logStreamNames = new HashSet<>();
   protected String logGroupName;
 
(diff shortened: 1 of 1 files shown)


validator/src/main/java/com/amazon/aoc/validators/AbstractStructuredLogValidator.java
```diff
@@ -113,7 +114,7 @@ protected void fetchAndValidateLogs(Instant startTime) throws Exception {
   }
 
   private void checkResult() throws BaseException {
-    if (schemasToValidate.isEmpty()) {
+    if (schemasToValidate.isEmpty() || schemasToValidate.keySet().equals(validatedSchema)) {
```
aateeqi, line 117: do we need to keep the `isEmpty()` as i assume the empty keySet would equal the empty validatedSchema?

validator/src/main/java/com/amazon/aoc/validators/AbstractStructuredLogValidator.java
```diff
@@ -131,12 +132,15 @@ private void checkResult() throws BaseException {
...
+    if (validatedSchema.contains(key)) {
+      return;
+    }
     JsonSchema jsonSchema = schemasToValidate.get(key);
     if (jsonSchema != null) {
       ProcessingReport report = jsonSchema.validate(
               JsonLoader.fromString(logEventNode.toString()));
       if (report.isSuccess()) {
-        schemasToValidate.remove(key);
+        validatedSchema.add(key);
```
aateeqi, line 143: nice makes much more sense ✔️

validator/src/main/java/com/amazon/aoc/validators/AbstractStructuredLogValidator.java
sethAmazon: I'm not sure how this is being used in other tests. But I do think this could be removed. There are multiple validators that are inherited from this abstract.

aateeqi: okay we can try it in another PR as you suggested ✔️

aateeqi (approved):

khanhntd (approved):
LGTM.

PR #462, opened 2021-12-31
convert to public repo

khanhntd:
Description:

Currently, the CI publishes the image to private repo and we need to pull it from there for integration test. However, as we discussed with the adot team, converting from private to public ECR would give us two advantages:

More visibility for the customer and dev to pull down the image without having the credentials.
Able to use multiple regions later on.

### DIFF
--- a/terraform/common.tf
+++ b/terraform/common.tf
@@ -31,7 +31,7 @@ variable "sample_app_image" {
 }
 
 variable "aoc_image_repo" {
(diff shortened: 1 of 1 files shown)


alolita:
@khanhntd let's discuss and then take a final decision.

sethAmazon (approved):

anuraaga (approved):

PR #468, opened 2022-01-04
Add OTLP metric mock.

jefchien:
**Description:** Adds a test case for the OTLP metric mock backend. This will be used in the performance tests instead of the current `otlp_metric`.

**Testing:** Ran the mock test.

### DIFF
--- a/terraform/testcases/otlp_metric_mock/otconfig.tpl
+++ b/terraform/testcases/otlp_metric_mock/otconfig.tpl
@@ -1,0 +1,29 @@
+extensions:
+  pprof:
+    endpoint: 0.0.0.0:1777
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:${grpc_port}
+
+processors:
+  batch:
+
+exporters:
+  logging:
+    loglevel: debug
+  awsemf:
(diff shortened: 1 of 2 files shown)


terraform/testcases/otlp_metric_mock/otconfig.tpl
```diff
@@ -0,0 +1,29 @@
...
+
+processors:
+  batch:
+
+exporters:
+  logging:
+    loglevel: debug
+  awsemf:
+    region: ${region}
+    endpoint: "https://${mock_endpoint}"
```
sethAmazon, line 18: Does this cause exceptions in the collector?

sethAmazon (approved):

terraform/testcases/otlp_metric_mock/otconfig.tpl
jefchien: It does not. https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/awsemfexporter has documentation about it, but the awsemf endpoint config allows the user to override the cloudwatch service endpoint. In this case, we're overriding it so the mock backend gets the request.

aateeqi (approved):

PR #473, opened 2022-01-06
Set instance type for performance tests back to m5.2xlarge.

jefchien:
**Description:** Added variable for sidecar instance type so the performance tests can run with a different configuration.

### DIFF
--- a/terraform/ec2_setup/amis.tf
+++ b/terraform/ec2_setup/amis.tf
@@ -3,7 +3,7 @@ variable "ami_family" {
     amazon_linux = {
       login_user                         = "ec2-user"
       install_package                    = "aws-otel-collector.rpm"
-      instance_type                      = "t2.micro"
+      instance_type                      = "m5.2xlarge"
       otconfig_destination               = "/tmp/ot-default.yml"
       download_command_pattern           = "wget %s"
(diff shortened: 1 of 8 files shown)


terraform/ec2_setup/variables.tf
```diff
@@ -16,6 +16,9 @@
 variable "testing_ami" {
 }
 
+variable "sidecar_instance_type" {
```
aateeqi, line 19: would a default be useful here?

terraform/ec2_setup/amis.tf
```diff
@@ -3,7 +3,7 @@ variable "ami_family" {
     amazon_linux = {
       login_user                         = "ec2-user"
       install_package                    = "aws-otel-collector.rpm"
-      instance_type                      = "t2.micro"
+      instance_type                      = "m5.2xlarge"
```
sethAmazon, line 6: Should this default be smaller?

jefchien: As far as I can tell, this is only used by the performance tests. I don't know if we can reduce the size. Some of our performance tests take 20+ GBs of memory with our current setup. An m5.xlarge wouldn't have enough memory (16GB).

aateeqi (approved):

sethAmazon (approved):

PR #475, opened 2022-01-06
Remove X-Ray sample app from mock.

jefchien:
**Description:** The mock tests don't have the ECR credentials, so only local sample apps are available.

### DIFF
--- a/terraform/testcases/xrayreceiver_mock/parameters.tfvars
+++ b/terraform/testcases/xrayreceiver_mock/parameters.tfvars
@@ -1,8 +1,5 @@
 validation_config = "default-mocked-server-xrayreceiver-validation.yml"
 
-# src: https://github.com/aws/aws-xray-sdk-go/blob/master/sample-apps/http-server/application.go
-sample_app_image = "611364707713.dkr.ecr.us-west-2.amazonaws.com/amazon/aws-otel-goxray-sample-app:v1.1.0"
-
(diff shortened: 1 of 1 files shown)


anuraaga (approved):

khanhntd:
It's hard for me to understand the credentials since the [default value for sample app](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/basic_components/outputs.tf#L48-L50) is still the private ECR. Therefore, should we create a public ECR repo for these? In this way, it would avoid create duplicating sample app in multiple regions and we don't need credentials for the sample app.

jefchien:
> It's hard for me to understand the credentials since the [default value for sample app](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/basic_components/outputs.tf#L48-L50) is still the private ECR. Therefore, should we create a public ECR repo for these? In this way, it would avoid create duplicating sample app in multiple regions and we don't need credentials for the sample app.

Well, the mock tests don't use the `basic_components` module, so that's the wrong default value for the sample apps in this case. If a sample app is not provided, then it'll use spark.

https://github.com/aws-observability/aws-otel-test-framework/blob/c5c038b302854e5a4d1d683f1d5a2bfc300de087/terraform/common.tf#L13-L15

sethAmazon (approved):

PR #494, opened 2022-01-10
Use JDK 11 For Docker Builds

sethAmazon:
**Description:** 
Our builds use jdk 11 not the gradle 7 default of 17


khanhntd (approved):
LGTM.

PR #495, opened 2022-01-11
Update Jaeger Zipkin Sample App

sethAmazon:
**Description:** 
Update Jaeger Zipkin Sample App

**Testing:** 
Zipkin mock test

### DIFF
--- a/trace-java-client/src/main/java/com/amazon/opentelemetry/trace/client/JaegerTraceEmitClient.java
+++ b/trace-java-client/src/main/java/com/amazon/opentelemetry/trace/client/JaegerTraceEmitClient.java
@@ -23,14 +23,14 @@ import com.amazon.opentelemetry.trace.model.trace.Trace;
 import com.amazon.opentelemetry.trace.utils.OpenTracingTraceConverter;
 import com.amazon.opentelemetry.trace.utils.TraceGenerator;
 import com.amazon.opentelemetry.trace.utils.TraceTraversal;
-import io.jaegertracing.reporters.RemoteReporter;
-import io.jaegertracing.reporters.Reporter;
-import io.jaegertracing.samplers.ConstSampler;
-import io.jaegertracing.senders.HttpSender;
-import io.jaegertracing.Tracer;
-import io.jaegertracing.Tracer.Builder;
+import io.jaegertracing.internal.JaegerTracer;
+import io.jaegertracing.internal.reporters.RemoteReporter;
+import io.jaegertracing.internal.samplers.ConstSampler;
+import io.jaegertracing.spi.Reporter;
+import io.jaegertracing.thrift.internal.senders.HttpSender;
+import io.opentracing.Tracer;
 import io.opentracing.propagation.Format;
-import io.opentracing.propagation.TextMapInjectAdapter;
+import io.opentracing.propagation.TextMapAdapter;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.util.HashMap;
@@ -38,8 +38,10 @@ import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.mutable.MutableObject;
+import org.apache.thrift.transport.TTransportException;
 
 public class JaegerTraceEmitClient implements TraceClient {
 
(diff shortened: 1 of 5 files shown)


trace-java-client/src/main/java/com/amazon/opentelemetry/trace/model/trace/KeyValue.java
```diff
@@ -51,4 +51,12 @@ public static KeyValue ofBooleanType(String key, Boolean value) {
     kv.valueBool = value;
     return kv;
   }
+
+  public String getKey() {
```
jefchien: Are these getters used? I don't see a change in the clients that calls them.

terraform/testcases/jaeger_mock/parameters.tfvars
```diff
@@ -1,6 +1,6 @@
 validation_config = "jaeger-mocked-server-validation.yml"
 
-sample_app_image = "aottestbed/aws-otel-jz-sample-app:v0.1.4"
+sample_app_image = "public.ecr.aws/aws-otel-test/jaeger-zipkin-sample-app:v0.1.5"
 
 sample_app = "aws-otel-jz-sample-app"
```
jefchien: This `sample_app` name should probably be `jaeger-zipkin-sample-app` since that's the directory.

terraform/testcases/jaeger_mock/parameters.tfvars
```diff
@@ -1,6 +1,6 @@
 validation_config = "jaeger-mocked-server-validation.yml"
 
-sample_app_image = "aottestbed/aws-otel-jz-sample-app:v0.1.4"
+sample_app_image = "public.ecr.aws/aws-otel-test/jaeger-zipkin-sample-app:v0.1.5"
```
jefchien: Did you run the `jaegar_mock` test? I'm a bit surprised this works. With the mock tests, if the `sample_app_image` is provided, it'll try to pull the image. Usually, I've seen it complain about credentials because the mock module isn't set up to login to ECR.

sethAmazon: This is public ecr

sethAmazon: I did test on my mac but maybe I was still logged in when I ran this. I still think this will work since it is public ecr and you should not have to log in to pull.

trace-java-client/src/main/java/com/amazon/opentelemetry/trace/client/JaegerTraceEmitClient.java
```diff
@@ -89,13 +91,18 @@ private Tracer getTracer(Service service) {
   }
 
   private static Tracer createJaegerTracer(String collectorUrl, Service svc, int flushIntervalMillis) {
-    HttpSender sender = new HttpSender.Builder(collectorUrl + "/api/traces").build();
+    HttpSender sender = null;
+    try {
+      sender = new HttpSender.Builder(collectorUrl + "/api/traces").build();
+    } catch (TTransportException e) {
+      throw new RuntimeException("Can't create http sender");
```
aateeqi: Should we log any information about `e` here as well?

trace-java-client/src/main/java/com/amazon/opentelemetry/trace/client/ZipkinTraceEmitClient.java
```diff
@@ -94,7 +94,7 @@ public String emit(Trace trace) {
 
   private String extractTraceID(Tracer tracer, Span otSpan) {
     HashMap<String, String> baggage = new HashMap<>();
-    TextMapInjectAdapter map = new TextMapInjectAdapter(baggage);
+    TextMapAdapter map = new TextMapAdapter(baggage);
     tracer.inject(otSpan.context(), Format.Builtin.HTTP_HEADERS, map);
```
aateeqi, line 98: if we're only using `map` here, should we just pass it in directly to the tracer.inject() function?

sethAmazon: This has to do with how the java generics works for tracer.inject(v1,v2,v3) v2 and v3 must be of the same class.

terraform/testcases/jaeger_mock/parameters.tfvars
jefchien: Maybe I'm reading this wrong, but from the documentation, it looks like you need to authenticate in order to pull.
https://docs.aws.amazon.com/AmazonECR/latest/public/docker-pull-ecr-image.html

image

trace-java-client/src/main/java/com/amazon/opentelemetry/trace/client/JaegerTraceEmitClient.java
sethAmazon: Added

trace-java-client/src/main/java/com/amazon/opentelemetry/trace/client/ZipkinTraceEmitClient.java
sethAmazon: From Tracer.java
<C> void inject(SpanContext spanContext, Format<C> format, C carrier);

terraform/testcases/jaeger_mock/parameters.tfvars
anuraaga: Unfortunately "public" ECR does still effectively require login to avoid pull limits. It's a reason we can't run integration tests for instrumentation (which don't use terraform) against PR builds still - GHCR wouldn't have that problem. GHCR used to have an operations problem where images were all scoped to the organization but now they are scoped to repositories so perhaps we can give it a revisit, I notice we do seem to use it for python images already

https://github.com/orgs/aws-observability/packages

sethAmazon: Changed to using ghcr as repo

terraform/testcases/jaeger_mock/parameters.tfvars
jefchien: It looks like it was changed in the `zipkin_mock`, but not in the `jaeger_mock`.

terraform/testcases/jaeger_mock/parameters.tfvars
```diff
@@ -1,6 +1,6 @@
 validation_config = "jaeger-mocked-server-validation.yml"
 
-sample_app_image = "aottestbed/aws-otel-jz-sample-app:v0.1.4"
+sample_app_image = "ghcr.io/aws-observability/aws-otel-collector/jaeger-zipkin-sample-app:v0.1.5"
```
jefchien: Should this be in `aws-otel-test-framework`? The sample app isn't really a part of the collector.

sethAmazon: Use old image until we can make ghrc public

jefchien (approved):
LGTM

aateeqi (approved):

PR #498, opened 2022-01-13
Update Load Generator Deps

sethAmazon:
**Description:** 
After updating the load gen from dep bot. We were not able to generate proper otlp metrics and traces this fixes that. 

**Testing:** 
Ran the collector locally and ran the generator then saw traces and metrics in cloudwatch and xray.

### DIFF
--- a/load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
+++ b/load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
@@ -16,23 +16,23 @@
 package com.amazon.opentelemetry.load.generator.emitter;
 
 import com.amazon.opentelemetry.load.generator.model.Parameter;
-import io.grpc.ManagedChannelBuilder;
-import io.opentelemetry.api.metrics.GlobalMeterProvider;
-import io.opentelemetry.api.metrics.common.Labels;
+import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.metrics.LongCounter;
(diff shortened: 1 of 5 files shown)


load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
```diff
@@ -16,23 +16,24 @@
...
+import io.opentelemetry.sdk.metrics.export.PeriodicMetricReader;
+
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
 
 public class OtlpMetricEmitter extends MetricEmitter {
 
   private final static String API_NAME = "loadTest";
   LongCounter apiBytesSentCounter;
-  LongValueRecorder apiLatencyRecorder;
```
sethAmazon, line 35: Long value recorder is no longer supported. Closest I found was DoubleHistogram

sethAmazon: There is a long histogram but Meter only has double histogram

jefchien: There was a migration guide from 1.4.0 -> 1.5.0 https://github.com/open-telemetry/opentelemetry-java/discussions/3506
I'm not sure how much it still applies, but they used the `LongHistogram` to replace the `LongValueRecorder`.

sethAmazon:
Not needed

PR #499, opened 2022-01-14
Add EKS ARM64 Cluster

sethAmazon:
**Description:** 
Need a way to test arm64 docker images

**Testing:** 
terraform apply -auto-approve -var-file="../testcases/otlp_trace/parameters.tfvars" -var="aoc_image_repo=public.ecr.aws/aws-otel-test/adot-collector-integration-test" -var="testcase=../testcases/otlp_trace" -var="aoc_version=seth-multi-arc-test" -var="region=us-east-2" -var="eks_cluster_name=integ-test-arm64-cluster"

from eks terraform folder


terraform/eks_arm64_cluster_setup/main.tf
```diff
@@ -0,0 +1,114 @@
...
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+terraform {
+  required_providers {
+    kubectl = {
```
anuraaga, line 19: This provider doesn't seem to be used

terraform/eks_arm64_cluster_setup/main.tf
```diff
@@ -0,0 +1,114 @@
...
+  public_subnets  = ["10.0.101.0/24", "10.0.102.0/24", "10.0.103.0/24"]
+
+  enable_nat_gateway = true
+  enable_vpn_gateway = true
+  single_nat_gateway = true
+}
+
+module "eks" {
+  source = "terraform-aws-modules/eks/aws"
+
```
bryan-aguilar: We need to pin a version hwere for the eks module. We ran into the same issue with the `adot-op-cluster` where breaking changes made us unable to deploy. @sethAmazon  do you happen to have the version available? Would it be the same as the `adot-op-cluster` because we were dealing with the issue around the same time this PR was filed.

terraform/eks_arm64_cluster_setup/main.tf
```diff
@@ -0,0 +1,114 @@
...
+module "common" {
+  source = "../common"
+}
+
+provider "aws" {
+  region = var.region
+}
+
+module "vpc" {
+  source = "terraform-aws-modules/vpc/aws"
```
bryan-aguilar: We also need to pin a VPC version

terraform/eks_arm64_cluster_setup/variables.tf
```diff
@@ -0,0 +1,7 @@
+variable "eks_cluster_name" {
+  default = "integ-test-arm64-cluster"
+}
+
+variable "region" {
+  default = "us-east-2"
+}
```
bryan-aguilar: Is there a reason this is deployed in `us-east-2` instead of `us-west-2` with our other testing clusters?

bryan-aguilar (changes requested):

terraform/eks_arm64_cluster_setup/variables.tf
bryan-aguilar: Deployed in `us-east-2` due to limitations of number of vpcs available

bryan-aguilar: missing new line

terraform/eks_arm64_cluster_setup/main.tf
bryan-aguilar: Used on line 102

Kausik-A:
**Description**

On the suggestions in the thread, pinned the terraform,eks and vpc versions 

**Testing**

Manually tested the PR by deploying the cluster through `Terraform apply`and checked if everything is deployed correctly. Also checked if the resources are successfully destroyed.

Kausik-A:
**Description**

Added terraform state management through S3 for eks arm64 cluster

**Testing**

Manually tested the PR by deploying the cluster through `Terraform apply` and checked if everything is deployed correctly. Also checked if the resources are successfully destroyed.

terraform/eks_arm64_cluster_setup/main.tf
```diff
@@ -0,0 +1,123 @@
...
+  required_providers {
+    kubectl = {
+      source  = "gavinbunney/kubectl"
+      version = ">= 1.7.0"
+    }
+  }
+  backend "s3" {
+    bucket = "adot-op-cluster-terraform-statefile"
+    key    = "eks_arm64_cluster_setup/terraform.tfstate"
+    region = "us-west-2"
```
vasireddy99, line 27: should this be `var.region`?

vasireddy99 (changes requested):

terraform/eks_arm64_cluster_setup/main.tf
khanhntd: No. Backend does not accept variables because `backend is initiated before variables initiated` according to https://github.com/hashicorp/terraform/issues/13022 . Therefore, we can only use hard-code variables for backend.

vasireddy99 (approved):

bryan-aguilar (approved):

PR #514, opened 2022-01-18
Use Public ECR For Jaeger Zipkin Sample App

sethAmazon:
**Description:** 
Use public ecr to support arm images


jefchien (approved):

sethAmazon:
Merging with only 1 approval to unblock pipeline

Issue #497, opened 2022-01-13
Update sample-apps in the test-framework


vasireddy99:
The current sample apps in the test framework doesn't include recent changes, hence outdated.

- prometheus-sample-app
- spark

CC : @jefchien

bryan-aguilar:
CI/CD for sample apps under test is now in place. https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/.github/workflows/update-test-images.yml

PR #515, opened 2022-01-18
Remove the spark-sample-app from the otel-test-framework

vasireddy99:
This PR addresses the issue #497 

In-order to reduce the maintenance for [spark-sample app](https://github.com/aws-observability/aws-otel-java-instrumentation/tree/main/sample-apps/spark) form aws-otel-java-instrumentation in aws-otel-test-framework. Using the public ECR image `public.ecr.aws/aws-otel-test/aws-otel-java-spark:latest` as `sample-app-image` instead of dedupe.


anuraaga (approved):

khanhntd (approved):
So the whole point would be transferring all the sample apps to their corresponding aws-otel SDK. Therefore, reducing maintenance the sample app on aws-otel-test-framework correct? If that's the whole point, then it's good to me

terraform/testcases/otlp_mock/parameters.tfvars
```diff
@@ -1,2 +1,5 @@
 # data type will be emitted. Possible values: metric or trace
 soaking_data_mode = "trace"
+
+#sample app image
+sample_app_image = "public.ecr.aws/aws-otel-test/aws-otel-java-spark:latest"
```
jefchien, line 7: You'll want to change it in more places than this one. There are a lot of testcases (`logzio_exporter_trace_mock`, `newrelic_exporter_metric_mock`, etc.) that use the default sample app, which is the spark app.

jefchien (changes requested):

sethAmazon:
Does the spark latest support arm64 docker images?

vasireddy99:
> Does the spark latest support arm64 docker images?

yes, it supports them [PR](https://github.com/aws-observability/aws-otel-java-instrumentation/pull/140)

jefchien:
@vasireddy99 It looks like it's meant to, but the image on ECR https://gallery.ecr.aws/aws-otel-test/aws-otel-java-spark only has one.

vasireddy99:
Changing to draft till #526 is merged.

- `sample_app_image` takes precedence over `sample_app` couldn't take the approach of using it as default image

vasireddy99:
Closed due to rebasing issues. created [#527](https://github.com/aws-observability/aws-otel-test-framework/pull/527/files)

PR #516, opened 2022-01-18
Upgrade AWS Otel Goxray Sample App

sethAmazon:
**Description:** 
Upgrade AWS Otel Goxray Sample App

**Testing:** 
terraform apply -auto-approve -lock=false -var="aoc_image_repo=public.ecr.aws/aws-otel-test/adot-collector-integration-test" -var="aoc_version=latest" -var="testcase=../testcases/xrayreceiver" -var-file="../testcases/xrayreceiver/parameters.tfvars" -var="region=us-east-2" -var="eks_cluster_name=integ-test-arm64-cluster"


aateeqi (approved):

jefchien:
Should we make the X-Ray sample app image public? That way we would be able to include it in the xrayreceiver_mock as well.

sethAmazon:
I would say let us just unblock the pipeline for the release then we can make it public

jefchien (approved):

Issue #227, opened 2021-02-03
Allow reading expected test templates from files


anuraaga:
Currently, all expectations are in resources in this repo. But since sample apps and tested instrumentations are in different repos, really the expectations should live there too - when updating an app and something breaks, we fix the apps and tests at the same time.

wyTrivail:
@mxiamxia @vastin @pxaws

anuraaga:
@mxiamxia @wyTrivail @vastin @pxaws Is it possible to prioritize this issue? It will be quite helpful, for example for projects like the lambda repo.

wyTrivail:
@anuraaga  sorry for my late,  finally got some times to work on it. 

https://github.com/aws-observability/aws-otel-test-framework/pull/295/files

PR #522, opened 2022-01-20
Add expected trace template for ruby rails app

NathanielRN:
**Description:**

In anticipation of add a sample app showcasing OpenTelemetry Ruby for AWS X-ray, we want to provide the templates we expect the traces a simple Ruby on Rails app would generate.

In our tests, we found that the standard `/outgoing-http-call` produces the correct _default_ OTel Trace we expect, but the AWS SDK `/aws-sdk-call` does not. So, we add an option just for this ruby discrepancy and it's expected trace template as a `.mustache` file.

The trace for `/aws-sdk-call` is found here, it is only missing the `[0].http.response.status=200,` attribute.

```
{
    [0].name=aws-otel-manual-rails-sample,
    [0].id=8b6937b40eef8505,
    [0].start_time=1642644356.856313,
    [0].trace_id=1-61e8c384-fe837006488e7e62b8624d01,
    [0].end_time=1642644357.2015743,
    [0].subsegments[0].name=S3,
    [0].subsegments[0].id=feb9914f56
9361b6,
    [0].subsegments[0].start_time=1642644356.887765,
    [0].subsegments[0].end_time=1642644357.200523,
    [0].subsegments[0].namespace=aws,
    [0].subsegments[0].aws.xray.auto_instrum
entation=false,
    [0].subsegments[0].aws.xray.sdk_version=1.0.2,
    [0].subsegments[0].aws.xray.sdk=opentelemetry for ruby,
    [0].subsegments[0].aws.region=us-west-2,
    [0].subsegments[0].aws.operation=ListBuckets,
    [0].subsegments[0].metadata.default[\"rpc.service\"]=S3,
    [0].subsegments[0].metadata.default[\"rpc.system\"]=aws-api,
    [0].http.request.url=http://app:8080/aws-sdk-call,
    [0].http.request.method=GET,
    [0].http.request.user_agent=okhttp/4.9.0,
    [0].http.response.status=200,
    [0].http.response.content_length=0,
    [0].aws.xray.auto_instrumentation=false,
    [0].aws.xray.sdk_version=1.0.2,
    [0].aws.xray.sdk=opentelemetry for ruby,
    [0].metadata.default[\"otel.resource.process.command\"]=bin/rails,
    [0].metadata.default[\"otel.resource.telemetry.sdk.name\"]=opentelemetry,
    [0].metadata.default[\"otel.resource.process.pid\"]=7,
    [0].metadata.default[\"otel.resource.process.runtime.description\"]=ruby 2.7.5p203 (2021-11-24 revision f69aeb8314) [x86_64-linux],
    [0].metadata.default[\"otel.resource.process.runtime.version\"]=2.7.5,
    [0].metadata.default[\"otel.resource.process.runtime.name\"]=ruby,
    [0].metadata.default[\"otel.resource.service.name\"]=aws-otel-manual-rails-sample,
    [0].metadata.default[\"otel.resource.telemetry.sdk.language\"]=ruby,
    [0].metadata.default[\"otel.resource.telemetry.sdk.version\"]=1.0.2,
    [1].name=S3,
    [1].id=18d5983800c4d704,
    [1].parent_id=feb9914f569361b6,
    [1].start_time=1642644356.887765,
    [1].origin=AWS::S3,
    [1].trace_id=1-61e8c384-fe837006488e7e62b8624d01,
    [1].end_time=1642644357.200523,
    [1].inferred=true,
    [1].aws.xray.auto_instrumentation=false,
    [1].aws.xray.sdk_version=1.0.2,
    [1].aws.xray.sdk=opentelemetry for ruby,
    [1].aws.region=us-west-2,
    [1].aws.operation=ListBuckets
}
```

~We should create an issue on OpenTelemetry Ruby to fix this.~ Created this issue: https://github.com/open-telemetry/opentelemetry-ruby/issues/1094

**Link to tracking Issue:** N/A

**Testing:** Passes locally, it will be tested in CI once we have the ADOT Ruby repo setup.

**Documentation:** 

No documentation needed for this right now, since the trace is almost there. However we should aim for parity soon.

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/fileconfigs/PredefinedExpectedTemplate.java
+++ b/validator/src/main/java/com/amazon/aoc/fileconfigs/PredefinedExpectedTemplate.java
@@ -70,6 +70,9 @@ public enum PredefinedExpectedTemplate implements FileConfig {
   JS_SDK_AWSSDK_EXPECTED_TRACE(
     "/expected-data-template/js/jsAppExpectedAWSSDKTrace.mustache"
   ),
+  RAILS_SDK_AWSSDK_EXPECTED_TRACE(
+    "/expected-data-template/rails/railsAppExpectedAWSSDKTrace.mustache"
+  ),
 
   /**
    * Log structure template, defined in resources.
--- a/validator/src/main/resources/expected-data-template/rails/railsAppExpectedAWSSDKTrace.mustache
+++ b/validator/src/main/resources/expected-data-template/rails/railsAppExpectedAWSSDKTrace.mustache
@@ -1,0 +1,34 @@
+[{
+  "http": {
+    "request": {
+      "url": "{{endpoint}}/aws-sdk-call",
+      "method": "GET"
+    },
+    "response": {
+      "status": 200
+    }
+  },
+  "metadata": {
+    "default": {
+      "otel.resource.telemetry.sdk.name": "opentelemetry",
+      "otel.resource.telemetry.sdk.language": "ruby"
+    }
+  },
+  "subsegments": [
+    {
+      "name": "S3",
+      "aws": {
+        "operation": "ListBuckets"
+      },
+      "namespace": "aws"
+    }
+  ]
+},
+{
+  "name": "S3",
+  "inferred":true,
+  "aws": {
+      "operation": "ListBuckets"
+  },
+  "origin": "AWS::S3"
+}]
--- a/validator/src/main/resources/validations/rails-otel-trace-validation.yml
+++ b/validator/src/main/resources/validations/rails-otel-trace-validation.yml
@@ -1,0 +1,13 @@
+-
+  validationType: "trace"
+  httpPath: "/outgoing-http-call"
+  httpMethod: "get"
+  callingType: "http"
+  expectedTraceTemplate: "OTEL_SDK_HTTP_EXPECTED_TRACE"
+-
+  validationType: "trace"
+  httpPath: "/aws-sdk-call"
+  httpMethod: "get"
+  callingType: "http"
+  expectedTraceTemplate: "RAILS_SDK_AWSSDK_EXPECTED_TRACE"
+


sethAmazon:
Can you add a test case in here?

sethAmazon:
Can we hold off until they fix the ruby trace to add a status code?

NathanielRN:
> Can we hold off until they fix the ruby trace to add a status code?

We probably shouldn't because it's blocking our work to test everything else on the ADOT Ruby side which is pretty much ready to go. I added the issue we are tracking to fix it in the README above, but that would also have to wait for PR review, merge, and eventual release.

With this solution we're not making any long lasting changes that we won't be able to easily change later, so I think it should be good as a temporary solution?

> Can you add a test case in here?

Our downstream apps in the `aws-otel-<language>` repos consistently test these files, but I can add one here too! 🙂 Could you point me to an example? The closest I see to tests in the validator is https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/validator/src/test/java/com/amazon/aoc/validators/CWMetricValidatorTest.java but this seems to be testing the behavior of the Test Framework itself which makes sense. One way to test this would be to have a ruby app within this repo but that would probably be overkill.

Let me know what kind of test you have in mind? Otherwise I think we should leave the testing of specific validator templates to the language repos.

anuraaga:
@NathanielRN A long time ago #227 was fixed to allow templates to be read from a file. We never refactored our repos, but maybe this is the time to put the template in the downstream repo instead of here, what do you think? Especially since it sounds like we are already expecting some churn on the template as code changes upstream.

NathanielRN:
@anuraaga

Oh I didn't realize that! That's really good to know! ~Actually that would be super useful in the Lambda repos where [we currently copy and paste our local template into one of the Test Framework templates](
https://github.com/aws-observability/aws-otel-lambda/blob/486911eb8e05f4512d4abafaca0b9430914394e5/.github/workflows/main-build.yml#L161) 😅~

~I'll open an issue on the repo to do this.~ Never mind I see you already know about the application for Lambda 🙂 But I'm not sure if what we have currently in the link I provided is what you had in mind? My guess is it's not because this repo still has `validator/bin/main/validations/default-lambda-validation.yml`? I imagine if we used it we could delete it from the Test Framework repo and have it only live locally on the Lambda repo? If that's right I can create an issue there, let me know!

In terms of adding it to every downstream repo, I definitely think that would have its pros! Having them all on this separate repo puts a large amount of distance between the actual sample app/language where we test and the "expected" template we want to test against. Having them next to each other might also be good for "documentation" nice-ity because the user can see exactly what kind of trace they will get if they follow the Sample App setup.

However, I liked having the templates in this centralized place because it reinforces the idea that they _should_ all really eventually produce the `default-otel-trace-validation.yml` attributes _at minimum_, and that these extra files are just temporary while we move them to all be consistent. I guess I'm worried that putting them in the downstream repo will make it easier to forget that we need to make them all match?

I think the goal of consolidating is worth keeping them here until we have time to update all repos to have parity in the spans they produce because we hope to one day delete them.  Let me know if I understood you correctly? If I missed any benefit then happy to move this downstream! 😄

anuraaga:
It looks like we currently essentially have per language templates in this repo (go, js, java, and default (python?)) but if it seems practical to consolidate as the instrumentation has stabilized over time (one problem last year was that almost every release required editing the validation template 😅 ), then keeping things centralized is good. I think we need to do a sweep to see if that's possible, and we wouldn't need to move the file to every downstream repo.

Separately, for Ruby if there is already churn that we know is coming, it's probably better to keep that in the downstream repo until it settles and then switch to the default template when that'll work.

NathanielRN:
> but if it seems practical to consolidate as the instrumentation has stabilized over time

Lol OTel moves so fast that I feel like practicality is a luxury 😆 But yeah I agree with all you said! I think the biggest plus side is that our team and yours seem to agree that they want consistency among all the languages? But prioritizing the work is another issue so it probably won't be done soon 🙂 

> Separately, for Ruby if there is already churn that we know is coming, it's probably better to keep that in the downstream repo until it settles and then switch to the default template when that'll work.

Yeah that's fair, although I would say there's no church that is uncommon to the other languages (instrumentation being pre 1.0 for example).

I actually decided to proactively go ahead and make the change I want in Ruby which should give us the change we need in https://github.com/open-telemetry/opentelemetry-ruby/pull/1095! And the maintainers seem to make sense of my change so if that gets released sooner than later we won't need this PR at all 🙂

Will monitor it and keep this PR open. If it gets delayed for any reason then sure I can put it downstream. I guess all in all I should just close this PR if we want to err on the side of the instrumentations changing soon. Thanks for the discussion!

PR #590, opened 2022-03-09
Update `winrm` config settings for the canary_windows workflow

vasireddy99:
**Description:** 
This PR updates the `amis.tf` to avoid the issue in canary tests on `canary_windows` platform. reference [here](https://github.com/hashicorp/packer/issues/4567#issuecomment-286509007)

### DIFF
--- a/terraform/setup/setup.tf
+++ b/terraform/setup/setup.tf
@@ -16,6 +16,17 @@
 # in this module, we create the necessary common resources for the integ-tests, this setup module will only need to be executed once.
 # vpc, iam role, security group, the number of those resources could be limited, creating them concurrently for every pr would trigger throttling issue.
 
+#(TODO) pvasir - The terraform major version 4.0.0 is causing an versioning issue for s3 bucket, the version shall be updated once the fix is availble
+# reference - https://github.com/hashicorp/terraform-provider-aws/issues/23106
+terraform {
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+      version = "~> 3.74"
+    }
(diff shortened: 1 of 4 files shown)


terraform/setup/setup.tf
```diff
@@ -16,6 +16,19 @@
...
 # vpc, iam role, security group, the number of those resources could be limited, creating them concurrently for every pr would trigger throttling issue.
 
+#(TODO) pvasir - The terraform major version 4.0.0 is causing an versioning issue for s3 bucket, the version shall be updated once the fix is availble
+# reference - https://github.com/hashicorp/terraform-provider-aws/issues/23106
+/*
+terraform {
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+      version = "~> 3.74"
```
bryan-aguilar: Why are we commenting this out rather than pinning to a version prior to 4.0.0? When did our canary tests start to pick up version 4.0.0?

bryan-aguilar: Also why is this commented out rather than use the solution provided in your link [here ](https://github.com/hashicorp/terraform-provider-aws/issues/23125#issuecomment-1038424883)

vasireddy99: sure, probably, can create a seperate PR if this is required. since there is a workaround shall remove the comment. thank you

bryan-aguilar: Yea, if it is not relevant to this PR it should probably be removed

bryan-aguilar (approved):

PR #604, opened 2022-03-18
Saving terraform state file to a s3 bucket

Kausik-A:
**Description:** 
This Saves the terraform state file to a s3 bucket rather than local machine

**Testing:** Attached in the document

**Documentation:**

### DIFF
--- a/terraform/eks_adot_operator_cluster_setup/main.tf
+++ b/terraform/eks_adot_operator_cluster_setup/main.tf
@@ -20,11 +20,10 @@ terraform {
       version = ">= 1.7.0"
     }
   }
-
   ## comment out the backend block if you want the terraform state file to be locally saved on the computer (or) change the name of the s3 bucket if you are testing out on your personal AWS account
-  backend "s3"{
+  backend "s3" {
     bucket = "adot-op-cluster-terraform-statefile"
-    key = "global/s3/terraform.tfstate"
+    key    = "global/s3/terraform.tfstate"
     region = "us-west-2"
   }
 }
--- a/terraform/eks_adot_operator_cluster_setup/main.tf
+++ b/terraform/eks_adot_operator_cluster_setup/main.tf
@@ -23,7 +23,7 @@ terraform {
 
   ## comment out the backend block if you want the terraform state file to be locally saved on the computer (or) change the name of the s3 bucket if you are testing out on your personal AWS account
   backend "s3"{
-    bucket = "adot-cluster-terraform-statefile"
+    bucket = "adot-op-cluster-terraform-statefile"
     key = "global/s3/terraform.tfstate"
     region = "us-west-2"
   }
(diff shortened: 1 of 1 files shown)


terraform/eks_adot_operator_cluster_setup/main.tf
```diff
@@ -20,6 +20,12 @@ terraform {
       version = ">= 1.7.0"
     }
   }
+  ## comment out the backend block if you want the terraform state file to be locally saved on the computer (or) change the name of the s3 bucket if you are testing out on your personal AWS account
```
bryan-aguilar: Is this the only way to save the state file to a local machine after this change is merged? I am very much opposed to leaving in instructions to `comment out` code. This will greatly increase changes that we will have commented code merged back into the codebase. 

Shouldn't every user be saving the state file remotely even if it to a personal account? There are cost associations but so are the other infra assets that are being deployed.

terraform/eks_adot_operator_cluster_setup/main.tf
```diff
@@ -20,6 +20,12 @@ terraform {
...
+  backend "s3" {
```
bryan-aguilar, line 24: Does the backend module have a version we can pin? I think it would be a good idea to move away from using latest in these modules especially for sensitive items such as state files.

terraform/eks_adot_operator_cluster_setup/main.tf
bryan-aguilar: Are any other solutions possible? Toggle the saving of the file through a user provided variable? Maybe just always saving the file remotely but instructing users on how to pull the file from s3?

Kausik-A: I think its recommended to always have the state file saved on the s3 bucket. I don't see a way to disable the back-end block through a variable. We can give instructions to the user on curl the latest version of the state file from s3 bucket or reverting back to an older version.

bryan-aguilar: Sure, I would agree with that. We should remove this comment then.

terraform/eks_adot_operator_cluster_setup/main.tf
Kausik-A: The backend s3 module is part of the main terraform release. https://github.com/hashicorp/terraform/blob/main/website/docs/language/settings/backends/s3.mdx. So we can lock down the terraform version. But I am not sure if we want to do that as the tf script before the backend upgrade didn't lock down to a specific version. Lmk if its okay to do that

bryan-aguilar: I think it should be safe for this version but it may make sense to open an issue to see if it should be done as a whole for all modules in this repository.

Kausik-A: okay, pushed that into the new commit

Kausik-A: @bryan-aguilar is it okay we push this PR up? As per anuraag suggestion, I can lock the versions of others tf scripts on another PR

bryan-aguilar (approved):

Issue #605, opened 2022-03-19
Unable to Test `otlp_grpc_exporter_cw_amp_xray_ecs` in ECS Environment


erichsueh3:
I am trying to run the test `otlp_grpc_exporter_cw_amp_xray_ecs`, but have run into multiple errors. The first issue is an issue with setting up the s3 buckets.

Steps to reproduce this error:
1. run `cd terraform/setup && terraform init && terraform apply -auto-approve` from [here](https://github.com/open-o11y/aws-otel-test-framework/blob/terraform/docs/run-testing-framework.md#212-run-setup). The error is
```
Error: Value for unconfigurable attribute

with aws_s3_bucket.example,
on main.tf line 1, in resource "aws_s3_bucket" "example":
1: resource "aws_s3_bucket" "example" {

Can't configure a value for "acl": its value will be decided automatically based on the result of applying this configuration.

```

This is due to the [terraform version 4 upgrade](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-4-upgrade#acl-argument) and new resources need to be made for `acl`, `server_side_encryption_configuration `, and `versioning`.

After this first issue is resolved, `setup` and `imagebuild` work fine, though `imagebuild` gives an error but still builds and pushes all images to the ECR except the `jaeger-zipkin`sample-app`. It also gives a `local-exec provisioner error`, that seems to be the cause of the errors that follow as well.

Steps to reproduce this error:
1. run `cd terraform/setup && terraform init && terraform apply -auto-approve`
2. run `cd terraform/imagebuild && terraform init && terraform apply -auto-approve`
```
│ Error: local-exec provisioner error
│ 
│   with null_resource.build_and_push_sample_apps[0],
│   on main.tf line 58, in resource "null_resource" "build_and_push_sample_apps":
│   58:   provisioner "local-exec" {
│ 
│ Error running command 'docker build -t
│ 195181405982.dkr.ecr.us-west-2.amazonaws.com/otel-test/sample-apps:jaeger-zipkin-sample-app-latest
│ ../../sample-apps/jaeger-zipkin-sample-app/': exit status 1. Output: #1 [internal] load build definition
│ from Dockerfile
...
...
```

Then, when running the test, the test setup eventually hangs on retrying to create a validator. 

Steps to reproduce this error:
1. run `cd terraform/setup && terraform init && terraform apply -auto-approve`
[...]

PR #606, opened 2022-03-19
Changed usage of `awsprometheusremotewriteexporter` to `prometheusremotewriteexporter` and `sigv4authextension`

erichsueh3:
**Description:**
This PR is aligned with the [PR](https://github.com/aws-observability/aws-otel-collector/pull/1066) to add the sigv4 extension to the ADOT Collector. 

**Testing:** 
Note that testing has NOT been able to be performed; see #605 for more information

### DIFF
--- a/terraform/testcases/otlp_grpc_exporter_cw_amp_xray_ecs/otconfig.tpl
+++ b/terraform/testcases/otlp_grpc_exporter_cw_amp_xray_ecs/otconfig.tpl
@@ -1,6 +1,10 @@
 extensions:
   pprof:
     endpoint: 0.0.0.0:1777
+  sigv4auth:
+    region: ${region}
+    service: "aps"
+
 receivers:
   awsecscontainermetrics:
   otlp:
@@ -108,13 +112,10 @@ exporters:
     region: '${region}'
     resource_to_telemetry_conversion:
       enabled: true
-  awsprometheusremotewrite:
+  prometheusremotewrite:
     endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
     resource_to_telemetry_conversion:
       enabled: true
-    aws_auth:
-      region: ${region}
-      service: "aps"
   awsxray:
     local_mode: true
     region: '${region}'
@@ -130,7 +131,7 @@ service:
     metrics/container/amp:
       receivers: [ awsecscontainermetrics ]
       processors: [ filter, metricstransform, resource, batch ]
-      exporters: [ awsprometheusremotewrite,logging ]
+      exporters: [ prometheusremotewrite,logging ]
     metrics/application/cw:
       receivers: [ otlp ]
       processors: [ resourcedetection, batch ]
@@ -138,12 +139,12 @@ service:
     metrics/application/amp:
       receivers: [ otlp ]
       processors: [ resourcedetection, batch ]
-      exporters: [ awsprometheusremotewrite,logging]
+      exporters: [ prometheusremotewrite,logging]
     traces/application/xray:
       receivers: [ otlp ]
       processors: [ resourcedetection, batch ]
       exporters: [ awsxray ]
-  extensions: [pprof]
+  extensions: [pprof, sigv4auth]
   telemetry:
     logs:
       level: debug
--- a/terraform/testcases/otlp_grpc_exporter_cw_amp_xray_ecs/otconfig.tpl
+++ b/terraform/testcases/otlp_grpc_exporter_cw_amp_xray_ecs/otconfig.tpl
@@ -123,6 +123,8 @@ exporters:
     endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
     resource_to_telemetry_conversion:
       enabled: true
+    auth:
+      authenticator: sigv4auth
   awsxray:
     local_mode: true
     region: '${region}'
--- a/terraform/testcases/otlp_grpc_exporter_cw_amp_xray_ecs/otconfig.tpl
+++ b/terraform/testcases/otlp_grpc_exporter_cw_amp_xray_ecs/otconfig.tpl
@@ -112,6 +112,13 @@ exporters:
     region: '${region}'
     resource_to_telemetry_conversion:
       enabled: true
+  awsprometheusremotewrite:
+    endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
+    resource_to_telemetry_conversion:
+      enabled: true
+    aws_auth:
+      region: ${region}
+      service: "aps"
   prometheusremotewrite:
     endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
     resource_to_telemetry_conversion:
@@ -131,7 +138,7 @@ service:
     metrics/container/amp:
       receivers: [ awsecscontainermetrics ]
       processors: [ filter, metricstransform, resource, batch ]
-      exporters: [ prometheusremotewrite,logging ]
+      exporters: [ awsprometheusremotewrite,prometheusremotewrite,logging ]
     metrics/application/cw:
       receivers: [ otlp ]
       processors: [ resourcedetection, batch ]
@@ -139,7 +146,7 @@ service:
     metrics/application/amp:
       receivers: [ otlp ]
       processors: [ resourcedetection, batch ]
-      exporters: [ prometheusremotewrite,logging]
+      exporters: [ awsprometheusremotewrite,prometheusremotewrite,logging]
     traces/application/xray:
       receivers: [ otlp ]
       processors: [ resourcedetection, batch ]
--- a/terraform/testcases/otlp_grpc_exporter_cw_amp_xray_ecs_awsprw/otconfig.tpl
+++ b/terraform/testcases/otlp_grpc_exporter_cw_amp_xray_ecs_awsprw/otconfig.tpl
@@ -1,10 +1,6 @@
 extensions:
   pprof:
     endpoint: 0.0.0.0:1777
-  sigv4auth:
-    region: ${region}
-    service: "aps"
-
 receivers:
   awsecscontainermetrics:
   otlp:
@@ -112,12 +108,13 @@ exporters:
     region: '${region}'
     resource_to_telemetry_conversion:
       enabled: true
-  prometheusremotewrite:
+  awsprometheusremotewrite:
     endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
     resource_to_telemetry_conversion:
       enabled: true
-    auth:
-      authenticator: sigv4auth
+    aws_auth:
+      region: ${region}
+      service: "aps"
   awsxray:
     local_mode: true
     region: '${region}'
@@ -133,7 +130,7 @@ service:
     metrics/container/amp:
       receivers: [ awsecscontainermetrics ]
       processors: [ filter, metricstransform, resource, batch ]
-      exporters: [ prometheusremotewrite, logging ]
+      exporters: [ awsprometheusremotewrite,logging ]
     metrics/application/cw:
       receivers: [ otlp ]
       processors: [ resourcedetection, batch ]
@@ -141,12 +138,12 @@ service:
     metrics/application/amp:
       receivers: [ otlp ]
       processors: [ resourcedetection, batch ]
-      exporters: [ prometheusremotewrite, logging ]
+      exporters: [ awsprometheusremotewrite,logging]
     traces/application/xray:
       receivers: [ otlp ]
       processors: [ resourcedetection, batch ]
       exporters: [ awsxray ]
-  extensions: [pprof, sigv4auth]
+  extensions: [pprof]
   telemetry:
     logs:
       level: debug
--- a/terraform/testcases/otlp_metric_amp/otconfig.tpl
+++ b/terraform/testcases/otlp_metric_amp/otconfig.tpl
@@ -1,6 +1,9 @@
 extensions:
   pprof:
     endpoint: 0.0.0.0:1777
+  sigv4auth:
+    region: ${region}
+    service: "aps"
 receivers:
   otlp:
     protocols:
@@ -13,11 +16,8 @@ processors:
 exporters:
   logging:
     loglevel: debug
-  awsprometheusremotewrite:
+  prometheusremotewrite:
     endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
-    aws_auth:
-      region: ${region}
-      service: "aps"
     timeout: 10s
 
 service:
@@ -25,8 +25,8 @@ service:
     metrics:
       receivers: [otlp]
       processors: [batch]
-      exporters: [awsprometheusremotewrite,logging]
-  extensions: [pprof]
+      exporters: [prometheusremotewrite,logging]
+  extensions: [pprof, sigv4auth]
   telemetry:
     logs:
       level: debug
--- a/terraform/testcases/otlp_metric_amp/otconfig.tpl
+++ b/terraform/testcases/otlp_metric_amp/otconfig.tpl
@@ -16,6 +16,12 @@ processors:
 exporters:
   logging:
     loglevel: debug
+  awsprometheusremotewrite:
+    endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
+    aws_auth:
+      region: ${region}
+      service: "aps"
+    timeout: 10s
   prometheusremotewrite:
     endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
     timeout: 10s
@@ -25,7 +31,7 @@ service:
     metrics:
       receivers: [otlp]
       processors: [batch]
-      exporters: [prometheusremotewrite,logging]
+      exporters: [awsprometheusremotewrite,prometheusremotewrite,logging]
   extensions: [pprof, sigv4auth]
   telemetry:
     logs:
--- a/terraform/testcases/otlp_metric_amp/otconfig.tpl
+++ b/terraform/testcases/otlp_metric_amp/otconfig.tpl
@@ -25,6 +25,8 @@ exporters:
   prometheusremotewrite:
     endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
     timeout: 10s
+    auth:
+      authenticator: sigv4auth
 
 service:
   pipelines:
--- a/terraform/testcases/otlp_metric_amp_awsprw/otconfig.tpl
+++ b/terraform/testcases/otlp_metric_amp_awsprw/otconfig.tpl
@@ -1,9 +1,6 @@
 extensions:
   pprof:
     endpoint: 0.0.0.0:1777
-  sigv4auth:
-    region: ${region}
-    service: "aps"
 receivers:
   otlp:
     protocols:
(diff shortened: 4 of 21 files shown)


Aneurysm9:
Does this leave the `awsprometheusremotewrite` exporter untested?  Can we ensure that both paths are tested?

erichsueh3:
> Does this leave the `awsprometheusremotewrite` exporter untested? Can we ensure that both paths are tested?

Added back in the `awsprometheusremotewrite` exporter tests

erichsueh3:
I have cloned the tests that include the AWS PRW Exporter. Tests that include the AWS PRW Exporter have been renamed to `<original_directory_name>_awsprw`. I have also added in tests for the PRW Exporter + Sigv4 Extension, and those now reside in the original directories that the AWS PRW Exporter tests once lived in. This was all done so the two Exporters were being tested separately.

We have tested the `otlp_grpc_exporter_cw_amp_xray_ecs` test, the `prometheus_static` test, and the `prometheus_mock` test in ECS (`prometheus_mock` in local as well). The `otlp_metric_amp` test is supposed to be tested in EC2, but even with a default Collector image + unchanged tests, it is unable to be tested. The `prometheus_sd`, `prometheus_sd_adot_operator`, and `prometheus_static_adot_operator` tests are supposed to be tested in EKS, but even with a default Collector image + unchanged tests, it is unable to be tested. However, even though there are some environments that we still cannot test in (the above tests can only be tested in either EC2 or EKS), there is high confidence that this suite of tests all pass.

erichsueh3:
The `otlp_metric_amp` test fails in EC2, with this command:
```
cd terraform/ec2 && terraform init && terraform apply -auto-approve \
    -var="aoc_image_repo=public.ecr.aws/aws-observability/aws-otel-collector" \
    -var="aoc_version=latest" \
    -var="testcase=../testcases/otlp_metric_amp_awsprw" \
    -var-file="../testcases/otlp_metric_amp_awsprw/parameters.tfvars"
```
This is the error:
```
Error: remote-exec provisioner error
│ 
│   with null_resource.start_collector[0],
│   on main.tf line 277, in resource "null_resource" "start_collector":
│  277:   provisioner "remote-exec" {
│ 
│ error executing "/tmp/terraform_485573440.sh": Process exited with status 1
```
which stems from this
```
null_resource.start_collector[0] (remote-exec): error: aws-otel-collector.rpm: not an rpm package (or package manifest):
```

The tests in EKS fail with a different error than above, but they all fail the same. With this command:
```
cd terraform/eks && terraform init && terraform apply -auto-approve \
    -var="aoc_image_repo=public.ecr.aws/aws-observability/aws-otel-collector" \
    -var="aoc_version=latest" \
    -var="testcase=../testcases/prometheus_sd" \
    -var-file="../testcases/prometheus_sd/parameters.tfvars" \
    -var="eks_cluster_name=<MY_CLUSTER_NAME>r"
```
This is the error:
```
│ Error: Unauthorized
│ 
│   with kubernetes_namespace.aoc_ns,
│   on main.tf line 99, in resource "kubernetes_namespace" "aoc_ns":
│   99: resource "kubernetes_namespace" "aoc_ns" {
│ 


│ Error: Unauthorized
│ 
│   with kubernetes_namespace.aoc_fargate_ns,
│   on main.tf line 106, in resource "kubernetes_namespace" "aoc_fargate_ns":
│  106: resource "kubernetes_namespace" "aoc_fargate_ns" {
│ 
```

terraform/testcases/otlp_grpc_exporter_cw_amp_xray_ecs/otconfig.tpl
```diff
@@ -1,6 +1,10 @@
 extensions:
   pprof:
     endpoint: 0.0.0.0:1777
+  sigv4auth:
+    region: ${region}
```
bryan-aguilar, line 5: does this need to be? I see it in this format in the exporters section.
```suggestion
    region: '${region}'
```

terraform/testcases/otlp_grpc_exporter_cw_amp_xray_ecs_awsprw/otconfig.tpl
```diff
@@ -0,0 +1,149 @@
...
+    namespace: '${otel_service_namespace}/${otel_service_name}'
+    region: '${region}'
+    resource_to_telemetry_conversion:
+      enabled: true
+  awsprometheusremotewrite:
+    endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
+    resource_to_telemetry_conversion:
+      enabled: true
+    aws_auth:
+      region: ${region}
```
bryan-aguilar, line 116:
```suggestion
      region: '${region}'
```
`awsemf` exporter uses this format. Does it need to be used here?

terraform/testcases/otlp_grpc_exporter_cw_amp_xray_ecs_awsprw/otconfig.tpl
```diff
@@ -0,0 +1,149 @@
...
+        action: delete
+
+exporters:
+  awsemf:
+    namespace: '${otel_service_namespace}/${otel_service_name}'
+    region: '${region}'
+    resource_to_telemetry_conversion:
+      enabled: true
+  awsprometheusremotewrite:
+    endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
```
bryan-aguilar, line 112:
```suggestion
    endpoint: '${cortex_instance_endpoint}/api/v1/remote_write'
```

terraform/testcases/otlp_metric_amp/otconfig.tpl
```diff
@@ -1,6 +1,9 @@
 extensions:
   pprof:
     endpoint: 0.0.0.0:1777
+  sigv4auth:
+    region: ${region}
```
bryan-aguilar, line 5:
```suggestion
    region: '${region}'
```

terraform/testcases/otlp_metric_amp_awsprw/otconfig.tpl
```diff
@@ -0,0 +1,32 @@
...
+
+exporters:
+  logging:
+    loglevel: debug
+  awsprometheusremotewrite:
+    endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
+    aws_auth:
+      region: ${region}
+      service: "aps"
+    timeout: 10s
```
bryan-aguilar, line 21: Do we need `''` ticks?
```suggestion
  awsprometheusremotewrite:
    endpoint: '${cortex_instance_endpoint}/api/v1/remote_write'
    aws_auth:
      region: '${region}'
      service: "aps"
    timeout: 10s
```

terraform/testcases/prometheus_mock/otconfig.tpl
```diff
@@ -1,3 +1,6 @@
+extensions:
+  sigv4auth:
+    region: "us-west-2"
```
bryan-aguilar, line 3: Why is region fixed here but passed in on other configs?

terraform/testcases/prometheus_mock_awsprw/otconfig.tpl
```diff
@@ -0,0 +1,22 @@
...
+        scrape_interval: 15s
+      scrape_configs:
+      - job_name: "test-pipeline-job"
+        static_configs:
+        - targets: [ ${sample_app_listen_address_host}:${sample_app_listen_address_port} ]
+exporters:
+  awsprometheusremotewrite:
+    endpoint: "https://${mock_endpoint}"
+    aws_auth:
+      region: "us-west-2"
```
bryan-aguilar, line 14: Same question on why this region is fixed.

terraform/testcases/prometheus_sd/otconfig.tpl
```diff
@@ -1,3 +1,7 @@
+extensions:
+  sigv4auth:
+    region: ${region}
```
bryan-aguilar, line 3:
```suggestion
    region: '${region}'
```

terraform/testcases/prometheus_sd_adot_operator/otconfig.tpl
```diff
@@ -1,3 +1,7 @@
+extensions:
+  sigv4auth:
+    region: ${region}
```
bryan-aguilar, line 3:
```suggestion
    region: '${region}'
```

terraform/testcases/prometheus_sd_adot_operator_awsprw/otconfig.tpl
```diff
@@ -0,0 +1,31 @@
+receivers:
+      prometheus:
```
bryan-aguilar, line 2: Is this tab spacing correct?

terraform/testcases/prometheus_sd_adot_operator_awsprw/otconfig.tpl
```diff
@@ -0,0 +1,31 @@
...
+              action: keep
+              regex: "aoc-ns-${testing_id}"
+            - source_labels: [ __meta_kubernetes_pod_label_app ]
+              action: keep
+              regex: "sample-app"
+    exporters:
+      awsprometheusremotewrite:
+        endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
+        aws_auth:
+          region: ${region}
```
bryan-aguilar, line 19:
```suggestion
      awsprometheusremotewrite:
        endpoint: '${cortex_instance_endpoint}/api/v1/remote_write'
        aws_auth:
          region: '${region}'
```

terraform/testcases/prometheus_sd_awsprw/otconfig.tpl
```diff
@@ -0,0 +1,31 @@
...
+          regex: "aoc-ns-${testing_id}"
+        - source_labels: [ __meta_kubernetes_pod_label_app ]
+          action: keep
+          regex: "sample-app"
+exporters:
+  awsprometheusremotewrite:
+    endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
+    aws_auth:
+      region: ${region}
+      service: "aps"
```
bryan-aguilar, line 20: Should this use prometheus remote write and sigv4 instead? Or is this test specifically for `awsprw`?

terraform/testcases/prometheus_static/otconfig.tpl
```diff
@@ -1,3 +1,7 @@
+extensions:
+  sigv4auth:
+    region: ${region}
```
bryan-aguilar, line 3:
```suggestion
    region: '${region}'
```

terraform/testcases/prometheus_static_adot_operator/otconfig.tpl
```diff
@@ -1,3 +1,7 @@
+extensions:
+  sigv4auth:
+    region: ${region}
```
bryan-aguilar, line 3:
```suggestion
    region: '${region}'
```

terraform/testcases/prometheus_static_adot_operator_awsprw/otconfig.tpl
```diff
@@ -0,0 +1,24 @@
+receivers:
+      prometheus:
+        config:
```
bryan-aguilar, line 3: Is this tab spacing correct?

terraform/testcases/prometheus_static_adot_operator_awsprw/otconfig.tpl
```diff
@@ -0,0 +1,24 @@
...
+            scrape_interval: 15s
+          scrape_configs:
+          - job_name: "test-prometheus-sample-app"
+            static_configs:
+            - targets: [ ${sample_app_listen_address_host}:${sample_app_listen_address_port} ]
+    exporters:
+      awsprometheusremotewrite:
+        endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
+        aws_auth:
+          region: ${region}
```
bryan-aguilar, line 14:
```suggestion
      awsprometheusremotewrite:
        endpoint: '${cortex_instance_endpoint}/api/v1/remote_write'
        aws_auth:
          region: '${region}'
```

terraform/testcases/prometheus_static_awsprw/otconfig.tpl
```diff
@@ -0,0 +1,24 @@
...
+      scrape_configs:
+      - job_name: "test-prometheus-sample-app"
+        static_configs:
+        - targets: [ ${sample_app_listen_address_host}:${sample_app_listen_address_port} ]
+exporters:
+  awsprometheusremotewrite:
+    endpoint: ${cortex_instance_endpoint}/api/v1/remote_write
+    aws_auth:
+      region: ${region}
+      service: "aps"
```
bryan-aguilar, line 15: Same question as before, should this be using `awsprw`?

terraform/testcases/otlp_metric_amp_awsprw/otconfig.tpl
erichsueh3: We don't use ticks, here or above/below; any of the tests with the `_awsprw` suffix are the same as tests that already exist in the repository (and thus have been passing already)

terraform/testcases/prometheus_mock_awsprw/otconfig.tpl
erichsueh3: Good question, I didn't write the original tests so I'm not sure why it's fixed, and same for above.

terraform/testcases/prometheus_sd_adot_operator_awsprw/otconfig.tpl
erichsueh3: It does look off, but again I did not write these and are exactly what they are in the current repository

terraform/testcases/prometheus_sd_awsprw/otconfig.tpl
erichsueh3: any of the tests with the `_awsprw` suffix are the same as tests that already exist in the repository (for example, the `prometheus_sd_awsprw` testcase is the same as the `prometheus_sd` testcase that lives in the repository already. I've modified the `prometheus_sd` testcase to use prw+sigv4, and moved usage of the aws prw to the `prometheus_sd_awsprw` testcase

terraform/testcases/prometheus_static_adot_operator_awsprw/otconfig.tpl
erichsueh3: See this [comment](https://github.com/aws-observability/aws-otel-test-framework/pull/606#discussion_r840787649) above

terraform/testcases/prometheus_static_awsprw/otconfig.tpl
erichsueh3: See [here](https://github.com/aws-observability/aws-otel-test-framework/pull/606#discussion_r840788037)

vasireddy99:
hey @erichsueh3, I would prefer to just add the tests specific to `prometheusremotewriteexporter+sigv4 extension` and leave the current tests uninterrupted. which would be a vice-versa in this case.

vasireddy99 (commented):
hey @erichsueh3, I would prefer to just add the tests specific to prometheusremotewriteexporter+sigv4 extension and leave the current tests uninterrupted. which would be a vice-versa in this case :-), if that sounds good.

erichsueh3:
> hey @erichsueh3, I would prefer to just add the tests specific to `prometheusremotewriteexporter+sigv4 extension` and leave the current tests uninterrupted. which would be a vice-versa in this case :-), if that sounds good.

The reason why I did it the way it is is because we recommend use of the PRW Exporter + Sigv4 Extension over the AWS PRW Exporter and thought that this was better reflected the way I structured the tests; if anyone else feels strongly about changing it to be the other way around, I can change it

bryan-aguilar:
Do we lose any test cases with the current implementation? As long as we are confident that all the existing `aws_pwr` tests still exist, even if they are renamed, then I think it is just a semantics issue. If the current PR has removed test cases from `aws_pwr` then those need to be added back. 

It does make sense that if we are recomending the use of `prw + sigv4` then our prometheus test cases should use that. I just want to make sure that no coverage was lost from `aws_prw`.

vasireddy99:
> > hey @erichsueh3, I would prefer to just add the tests specific to `prometheusremotewriteexporter+sigv4 extension` and leave the current tests uninterrupted. which would be a vice-versa in this case :-), if that sounds good.
> 
> The reason why I did it the way it is is because we recommend use of the PRW Exporter + Sigv4 Extension over the AWS PRW Exporter and thought that this was better reflected the way I structured the tests; if anyone else feels strongly about changing it to be the other way around, I can change it

I totally agree, but the recommendations has nothing to do with the existing tests. so by added them as seperate tests, it would be better to track `prw+sigv4` tests in our CI too, also considering the scenario that where we were not able to reproduce all the `test-cases` locally. I strongly recommend for these above reasons.

bryan-aguilar:
> > > hey @erichsueh3, I would prefer to just add the tests specific to `prometheusremotewriteexporter+sigv4 extension` and leave the current tests uninterrupted. which would be a vice-versa in this case :-), if that sounds good.
> > 
> > 
> > The reason why I did it the way it is is because we recommend use of the PRW Exporter + Sigv4 Extension over the AWS PRW Exporter and thought that this was better reflected the way I structured the tests; if anyone else feels strongly about changing it to be the other way around, I can change it
> 
> I totally agree, but the recommendations has nothing to do with the existing tests. so it would be better to track `prw+sigv4` tests in our CI too, also considering the scenario that where we were not able to reproduce the tests locally. I strongly recommend for these above reasons.

In that case shouldn't this test case `terraform/testcases/prometheus_sd/otconfig.tpl` be renamed to `prometheus_sd_awsprw` since it uses `awsprw` instead of `prw+sigv4`. Which is the change that was already made. I think leaving prometheus test names without differentiation between exporters would be a bad choice. 

Would a compromise be to change all prometheus tests to indicate which exporter they are using? So we can have a differentiation between `sigv4` and `awsprw` implementations? In that case the above mentioned test would need some indicator that it was using `sigv4` in the name. 

The downside to this is that the additional context of `sigv4` vs `awsprw` could probably be removed from test names when `awsprw` is removed from the ADOT Collector. Would it just make sense to leave test cases that use `prw+sigv4` without any additional tag and instead tag `awsprw` tests?

bryan-aguilar:
@erichsueh3  could you possibly provide some more context on how you renamed/moved tests in the PR description?

erichsueh3:
> @erichsueh3 could you possibly provide some more context on how you renamed/moved tests in the PR description?

Yeah, for the `_awsprw` tests I created a new folder with the suffix for that test, went to the corresponding test in the repo, and just did a "Copy raw contents" for the `.tpl` and `tf.vars` files to make sure that those were the same. When I did this, the changes to the original test cases were already made since the recommendation to have the tests split weren't until after the PR was made

bryan-aguilar:
That makes a lot of sense. I think this is an okay way forward. I understand wanting to be cautious about adjusting existing tests. This should be a nonissue as long the test case matrix is properly updated in the `aws-otel-collector` repository. This way makes it easier to simply delete the `_awspr` tests when `awspwr` is removed from `aws-otel-collector`.

Aneurysm9 (approved):

bryan-aguilar (approved):

vasireddy99 (approved):

PR #607, opened 2022-03-19
Terraform version 4 upgrade - s3 resource changes

erichsueh3:
**Description:** 
See first error in #605, made changes as suggested by [terraform version 4 upgrade](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-4-upgrade#acl-argument).

### DIFF
--- a/terraform/setup/setup.tf
+++ b/terraform/setup/setup.tf
@@ -239,7 +239,7 @@ resource "aws_s3_bucket_server_side_encryption_configuration" "terrafrom-state-e
   #To encrypt the content
   rule {
     apply_server_side_encryption_by_default {
-      sse_algorithm     = "AES256"
+      sse_algorithm = "AES256"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


vasireddy99 (approved):

terraform/setup/setup.tf
```diff
@@ -224,19 +224,22 @@ resource "aws_ecr_repository" "mocked_server_ecr_repo" {
...
-
-  #Allow multiple upload
-  versioning {
-    enabled = true
+resource "aws_s3_bucket_versioning" "terrafrom-state-versioning" {
+  bucket = aws_s3_bucket.terrafrom-state.id
+  versioning_configuration {
+    status = "Enabled"
   }
+}
```
bryan-aguilar, line 234: Is the purpose of this bucket to house the terraform state file? If so there is a better way to implement this using the terraform backend module. https://www.terraform.io/language/settings/backends

bryan-aguilar: I now see the commented code out on top. I am unsure why that's not being used.

bryan-aguilar:
Will this affect existing infrastructure at all in the testing account?

erichsueh3:
See [here](https://github.com/hashicorp/terraform-provider-aws/issues/23125#issuecomment-1038424883) for another solution. Both that solution and the one listed in this PR are for local testing, but should not affect the testing account as the `setup` is already done. As long as `terraform destroy` is not ran in `terraform/setup` it should be okay. However I do think because of that this PR doesn't need to be merged; this should be added to documentation in some form though

erichsueh3:
Closing this PR, but still note that this needs to be documented

vasireddy99:
> Closing this PR, but still note that this needs to be documented

Please create an relevant issue, so we keep track of it, if that sounds good :)

PR #622, opened 2022-04-06
Update terraform setup

bryan-aguilar:
**Description:** The setup module was broken due to changes in the s3 bucket definitions post version `4.0.0`. This PR makes changes so that bucket definitions are now compliant with the aws version `4.8.0`. See documentation [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket) for s3 reference. 

This script also failed due to the hard coded bucket names. S3 bucket names need to be globally unique. I added an env var which appends a `bucketId` which can be set by the user. 

**Testing:** Deployed setup to fresh AWS account

**Documentation:** Updated setup docs

### DIFF
--- a/docs/run-testing-framework.md
+++ b/docs/run-testing-framework.md
@@ -38,7 +38,10 @@ Refer to: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.h
 
 #### 2.1.2 Setup unique bucket ID
 
-First create a unique S3 bucket identifier. This needs to be globally unique. Change example to an UUID of your choice
+First create a unique S3 bucket identifier that will be appened to your S3 bucket names. This will
+ensure that the S3 bucket name is globally unique. The UUID can be generated with any method of your
+choosing.
+See [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html) for S3 bucket naming rules.
 ```shell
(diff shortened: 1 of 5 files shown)


docs/run-testing-framework.md
```diff
@@ -35,8 +35,16 @@ In the case that you want to debug for a certain platform, you can also use this
  
 #### 2.1.1 Setup your aws credentials
 Refer to: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html
+
+#### 2.1.2 Setup unique bucket ID
+
+First create a unique S3 bucket identifier. This needs to be globally unique. Change example to an UUID of your choice
```
Aneurysm9: Should the docs here make clear that this is appended to a fixed portion of a bucket name and doesn't need to be a whole bucket name?  Maybe provide an example that generates a random value rather than just `example`?

```shell
export TF_VAR_bucketId=$(dd if=/dev/urandom bs=1k count=1k | shasum | cut -b 1-8)
```

bryan-aguilar: Yea your suggestion makes more sense.
Would changing the variable name away from `bucketId` help also? What could we call it instead? `bucketUUID`?

bryan-aguilar:
I also removed the dash to not conflict with existing bucket deployments.

Aneurysm9 (approved):

PR #634, opened 2022-04-13
Add batch execution of tests

bryan-aguilar:
**Description:** 
**Please do not merge this PR**
This will be a multi PR change to add the ability to generate and execute a batch of test cases for the ADOT Collector. I would like to merge them all at once in case we have to cherry pick this into previous release branches. Having a single merge commit with all changes will reduce the chance that anything is missed. We could also merge this into a dev branch if it is easier that way.

This is Part 1 of N.

This PR adds a make target to execute the tests along with a shell script to parse the `test-case-batch` file 
and execute the `terraform apply`. 

A tool to automatically generate the `test-case-batch` file will be added in subsequent PRs.

### DIFF
--- a/docs/run-testing-framework.md
+++ b/docs/run-testing-framework.md
@@ -213,6 +213,44 @@ Don't forget to clean up your resources:
 terraform destroy -auto-approve
 ````
 
+### 2.7 Run batch tests
+
+Batch testing allows a set of tests to be ran synchronously. To do this,
+a `test-case-batch` file is required in the `./terraform` directory. 
+The format of the `test-case-batch` file is as such. The `test-case-batch` 
+file can contain an infinite amount of test cases as
+long as the file follows this format.
+
+```
+serviceName1 testCase1 additionalValues1
+serviceName2 testCase2 additionalValues2
+serviceName3 testCase3 additionalValues3
+serviceNameN testCaseN additionalValuesN
+```
+
+The values for these fiels are as follows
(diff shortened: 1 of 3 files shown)


docs/run-testing-framework.md
```diff
@@ -213,6 +213,44 @@ Don't forget to clean up your resources:
 terraform destroy -auto-approve
 ````
 
+### 2.7 Run batch tests
+
+Batch testing allows a set of tests to be ran synchronously. To do this,
+a `test-case-batch` file is required in the `./terraform` directory. 
+The format of the `test-case-batch` file is as such. The `test-case-batch` 
+file can contain an infinite amount of test cases as
```
Aneurysm9: Infinite?  Surely it's at least bounded by available disk space.

terraform/Makefile
```diff
@@ -1,5 +1,22 @@
+ifeq ($(origin .RECIPEPREFIX), undefined)
+  $(error This Make does not support .RECIPEPREFIX. Please use GNU Make 4.0 or later)
```
Aneurysm9: Why?  This seems unnecessary and impedes portability.  I wouldn't be able to run this locally as I have `GNU Make 3.81`.

bryan-aguilar: The argument was found in this [post](https://tech.davis-hansson.com/p/make/) as I was doing some `make` research. The gist is that the recipe prefix can avoid any errors brought by tabs or spaces.

docs/run-testing-framework.md
bryan-aguilar: Updated

PR #670, opened 2022-05-05
Add timeout and retry to terraform apply

### LINKED ISSUE #668: Add timeout and retry to executeTerraformTest.sh
`executeTerraformTest.sh` is the backbone of the `aws-otel-collector` CI workflow now. It is responsible for setting up the env and executing each testcase given to it. 

During CI runs there are instances where certain test cases may go on for far too long due to hanging on certain resource deployments. [Timeout](https://man7.org/linux/man-pages/man1/timeout.1.html) should be added to `terraform apply` command which can cap each test case at a certain time. This will stop the long running tests that can sometimes stall out completion. Timeouts could also be added to specific modules that are failing, such as collector deployments, but it would require more effort to target each resource used individually. 

A retry mechanic should also be added so that if a test case fails it will be retried `N` amount of items. Adding it inside of the shell script is a better solution than implementing the github `retry` action because it severely reduced coupling to GitHub Actions.

bryan-aguilar:
**Description:** Adds timeout and retry functionality to `terraform apply` steps. By default this will only retry once but that can be tweaked if necssary. 

Timeout was displaying a weird behavior where two interrupt signals were sent despite only logging one. This can be seen through this log during testing.
```
timeout: sending signal INT to command ‘terraform’
Stopping operation...

Interrupt received.
Please wait for Terraform to exit or data loss may occur.
Gracefully shutting down...


Two interrupts received. Exiting immediately. Note that data loss may have occurred.

╷
│ Error: operation canceled
│ 
│ 
```
I have found correspondence that this may be due to the version I am using for `coreutils` [here](https://stackoverflow.com/questions/66518048/double-signal-send-with-command-timeout). This could cause an issue since the ubuntu vm uses `coreutils` version `8.30-3ubuntu2` but it is unknown without testing on the workflow itself. 

**Testing:** Tested on local machine with tests that were guaranteed to fail. Set timeout very low to test timeout and retry mechanic at the same time. 

Fixes #668

### DIFF
--- a/terraform/executeTerraformTest.sh
+++ b/terraform/executeTerraformTest.sh
@@ -64,14 +64,17 @@ case "$service" in
     ;;
 esac
 
+checkCache(){
+    CACHE_HIT=$(aws dynamodb get-item --region=us-west-2 --table-name ${DDB_TABLE_NAME} --key {\"TestId\":{\"S\":\"$1$2$3\"}\,\"aoc_version\":{\"S\":\"${TF_VAR_aoc_version}\"}})
+}
 
-CACHE_HIT=$(aws dynamodb get-item --region=us-west-2 --table-name ${DDB_TABLE_NAME} --key {\"TestId\":{\"S\":\"$1$2$3\"}\,\"aoc_version\":{\"S\":\"${TF_VAR_aoc_version}\"}})
-
-
-if [ -z "${CACHE_HIT}" ]; then
-    cd ${TEST_FOLDER};
+checkCache $1 $2 $3
+# Used as a retry mechanic.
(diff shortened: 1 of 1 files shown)


bryan-aguilar:
Based on further research it looks like if a workflow is canceled it will not allow time for terraform to shut down gracefully. This can be seen [here](https://github.com/aws-observability/aws-otel-collector/runs/6265576144?check_suite_focus=true#step:8:10253). The timeout is implemented so that workflows do not have to be canceled when tests hang as linked earlier.

Aneurysm9 (approved):

PR #706, opened 2022-06-01
Downgrade Operator image and helm chart

Kausik-A:
**Description:** 

Downgraded the operator ECR image to `v0.45.0` and pinned `opentelemetry-operator` helm chart to `0.6.4`


bryan-aguilar:
I think there is a better way to do this instead of pinning to a specific version within the module. Instead of forcing an explicit version we should consider using a variable that can be changed at test runtime.

Kausik-A:
> I think there is a better way to do this instead of pinning to a specific version within the module. Instead of forcing an explicit version we should consider using a variable that can be changed at test runtime.

Good Point. Dynamically changing the yaml file would require us to pass it as an environment variable. Can we first check the  `otlp_metric_adot_operator` with this PR and if indeed it passes; I will make the necessary changes.

bryan-aguilar:
> > I think there is a better way to do this instead of pinning to a specific version within the module. Instead of forcing an explicit version we should consider using a variable that can be changed at test runtime.
> 
> Good Point. Dynamically changing the yaml file would require us to pass it as an environment variable. Can we first check the `otlp_metric_adot_operator` with this PR and if indeed it passes; I will make the necessary changes.

This is intentionally adding tech debt. While I understand the desire to validate the findings first there is a tradeoff. Have you evaluated how much work it would be to use a variable instead this go around? If it will take too long at a bare minimum I would like an issue to be created to track this.

Kausik-A:
Made the necessary changes. I put in the default values as the versions I had pinned earlier but we can change it up on CLI or to latest

terraform/eks/adot-operator/adot_operator.tf
```diff
@@ -23,16 +23,29 @@ variable "kubeconfig" {
   default = "kubeconfig"
 }
 
+module "variables" {
+  source = "./variables"
+
+  operator_tag           = var.operator_tag
+  operator_helm_version  = var.operator_helm_version
+}
+
```
bryan-aguilar: Is it necessary to have an extra module for this? Can the variables just be referenced using `var.operator_tag` in the subsequent modules?

terraform/eks/variables.tf
```diff
@@ -54,4 +54,12 @@ variable "rollup" {
...
 variable "fargate_sample_app_lb_port" {
   default = "80"
+}
+
+variable "operator_tag" {
+  default = "v0.45.0"
+}
+
+variable "operator_helm_version" {
+  default = "0.6.4"
```
bryan-aguilar, line 64: Are these sensible defaults? Would we rather point them to latest? If we do not want to point them to latest then should we have no default and error out if none is provided? 

Also missing a newline at EOF

terraform/eks/adot-operator/adot_operator.tf
Kausik-A: In `eks/main.tf` there was a similar implementation ( [here](https://github.com/aws-observability/aws-otel-test-framework/blob/6f0f8929cdf602244750da9728b621dec8232429/terraform/eks/main.tf#L36) ) so followed off it.

bryan-aguilar: Is there any reason these cannot be scoped to the specific file like `testing_id` and `kubeconfig`?

Kausik-A: Implemented the suggestion

terraform/eks/variables.tf
Kausik-A: For now, Yeah I think so. Considering we are running this branch of `17.0.1`, the operator and helm chart would always be the same version. This cuts down our tasks of editing the workflow to pass this variable from CLI.
If we decide to port this over to the `terraform` branch, we can change the defaults to latest and none respectively

bryan-aguilar (approved):

PR #718, opened 2022-06-13
Use latest ghcr image for tests

erichsueh3:
Change Operator tests to use the ghcr Operator image instead of the ECR Operator image

### DIFF
--- a/terraform/eks/adot-operator/adot-operator-values.yaml
+++ b/terraform/eks/adot-operator/adot-operator-values.yaml
@@ -1,6 +1,6 @@
 manager:
   image:
-    repository: public.ecr.aws/aws-observability/adot-operator
+    repository: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
     tag: "latest"
 
 kubeRBACProxy:


bryan-aguilar:
Should we really be using the latest image though? Wouldn't we want to target the same version that is used in the ADOT Collector? What happens when the collector version drifts behind the upstream operator version?

erichsueh3:
> Should we really be using the latest image though? Wouldn't we want to target the same version that is used in the ADOT Collector? What happens when the collector version drifts behind the upstream operator version?

Yes, it is true that the ADOT Collector version will drift behind the upstream Operator version in between releases. But, the Operator tests would only be ran during Operator CI, which happens when we want to release the ADOT Operator, which happens alongside ADOT Collector releases

EDIT: minor correction to this statement, but Operator CI will be ran continuously, similar to Collector CI.

bryan-aguilar:
Does there need to be a process established for pinning versions in a release branch then?

erichsueh3:
> Does there need to be a process established for pinning versions in a release branch then?

Since Operator Ci will be ran continuously (see edit above) then the problem of having the ADOT Collector version drift behind the upstream Operator version becomes an issue again. Does the test framework allow for a way to use the correct ADOT Operator image during non-release CI, but also use the latest ghcr image during a release CI? Yes though, we need to be able to use the right image/pin the right version during release.

erichsueh3:
Marking as draft for now; opting to use correct ADOT Operator image during Operator CI to prevent drifting of versions, and researching a way to set Operator image in test framework during Operator CI for a release specifically

bryan-aguilar:
Food for thought coming up.
1. We know that the collector will always have a dependency on `go.opentelemetry.io/collector` in the root `go.mod` file
2. We can pull the version out of that in the github workflow
3. Turn the operator version into a variable and set the variable in the calling workflow

There would be an edge case where the collector version gets updated before the corresponding upstream operator version is available. In this case we would probably need to check to see if the image exists before running the tests. If the image doesn't exist the workflow should probably fail.

erichsueh3:
Closing as #750 addresses this issue

PR #725, opened 2022-06-14
Update spark sample app target

bryan-aguilar:
Reverts aws-observability/aws-otel-test-framework#647

Spark sample app needs to match with other test cases. Currently Spark sample app is published on merge to main in `aws-otel-java-instrumentation`. 

While this will update the sample app to  use the `latest` tag this may not be the best `tag` to target going forward. When a new image is published a `latest` tag and a tag for the `java-instrumentation` commit sha is published. We should instead be targetting the commit sha to avoid version drift in release branches.


vasireddy99:
Merging this PR.

vasireddy99 (approved):

PR #750, opened 2022-07-08
Change Operator repository and tag to variables

erichsueh3:
This PR changes the way operator images are set during the helm chart deployment and allows for it to be set as a variable.

### DIFF
--- a/terraform/eks/adot-operator/adot_operator.tf
+++ b/terraform/eks/adot-operator/adot_operator.tf
@@ -42,13 +42,13 @@ resource "helm_release" "adot-operator" {
   ]
 
   set {
-    name = "manager.image.repository"
-    value = "${var.operator_repository}"
+    name  = "manager.image.repository"
+    value = var.operator_repository
   }
 
   set {
-    name = "manager.image.tag"
-    value = "${var.operator_tag}"
+    name  = "manager.image.tag"
+    value = var.operator_tag
   }
 
   provisioner "local-exec" {
--- a/terraform/eks/adot-operator/adot_operator.tf
+++ b/terraform/eks/adot-operator/adot_operator.tf
@@ -23,6 +23,14 @@ variable "kubeconfig" {
   default = "kubeconfig"
 }
 
+variable "operator_repository" {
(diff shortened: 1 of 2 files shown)


erichsueh3:
One note: will need to make a PR in `aws-otel-collector` to set the operator version, as pulled from the `go.mod` `go.opentelemetry.io/collector` dependency, in the Operator CI github workflow (`export TF_VAR_xxx=...`). This will also likely include a change in the [execute terraform script](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/executeTerraformTest.sh)

EDIT: no change in the script is necessary

terraform/eks/adot-operator/adot_operator.tf
```diff
@@ -23,6 +23,14 @@ variable "kubeconfig" {
   default = "kubeconfig"
 }
 
+variable "operator_repository" {
+  default = "ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator"
+}
+
+variable "operator_tag" {
+  default = "latest"
```
bryan-aguilar, line 31:
```suggestion
  default = "latest"
```

Should we remove the default for this variable? This would make it required and thus would error out of this was not provided. 

On the other hand I am not certain that if doing would cause non `adot-operator` tests to error if they do not supply an `operator_tag` variable. We would need to test this to confirm.

erichsueh3: I think we should keep the default in as it would follow the logic that we had before where the operator repo/tag were hardcoded in `adot-operator-values.yaml`, just adding in more flexibility in it being a variable. Also, like you mentioned, removing it might cause issues elsewhere

erichsueh3:
This PR has a follow up in the `aws-otel-collector` repo, [1344](https://github.com/aws-observability/aws-otel-collector/pull/1344). This must be merged first before 1344. 

Also small note, not mentioned above, is that we use the GHCR Operator image now instead of the ECR ADOT Operator image

Aneurysm9 (approved):

PR #755, opened 2022-07-14
Referencing AMI ID directly in aws_instance

Kausik-A:
**Description:** 

This PR is fix the issue on EC2 AMI getting depreciated after 2 years. The test-framework uses a lot of the AMI images which are on the verge of getting depreciated soon. The EC2 documentation mentions that `Users of a deprecated AMI can continue to launch instances and describe the deprecated AMI using its ID` so the current fix is to references the`ami_id` directly in the `aws_instance` rather than searching it up using the API. 

**Testing:** 

ran `terraform apply` locally 

### DIFF
--- a/terraform/ec2/amis.tf
+++ b/terraform/ec2/amis.tf
@@ -399,43 +399,6 @@ locals {
   arch = var.amis[var.testing_ami]["arch"] == "amd64" ? "x86_64" : var.amis[var.testing_ami]["arch"]
 }
 
-# data "aws_ami" "selected" {
-#   most_recent = true
-
-#   filter {
-#     name = "name"
-#     values = [
-#     var.amis[var.testing_ami]["ami_search_pattern"]]
-#   }
-
-#   filter {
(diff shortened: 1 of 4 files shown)


terraform/ec2/amis.tf
```diff
@@ -380,41 +399,41 @@ locals {
...
+#       ]
+#     }
+#   }
 
-  owners = [
-  var.amis[var.testing_ami]["ami_owner"]]
-}
+#   owners = [
+#   var.amis[var.testing_ami]["ami_owner"]]
+# }
```
bryan-aguilar: Can we remove this instead of leaving a commented out block?

bryan-aguilar (approved):

terraform/ec2/amis.tf
Aneurysm9: +1. It'll be there in the git history if we need it later.

Kausik-A: Changed it

PR #767, opened 2022-07-19
Add operator arm64 tests

erichsueh3:
This PR adds testcases to be tested by the Operator with arm64

### DIFF
--- a/docs/run-testing-framework.md
+++ b/docs/run-testing-framework.md
@@ -238,7 +238,7 @@ serviceNameN testCaseN additionalValuesN
 ```
 
 The values for these fields are as follows
-`serviceName`: `EKS` `EKS-arm64` `EKS-fargate` `EKS-operator` `EKS-operator-arm64` `ECS` `EC2`
+`serviceName`: `EKS`, `EKS_ARM64`, `EKS_FARGATE`, `EKS_ADOT_OPERATOR`, `EKS_ADOT_OPERATOR_ARM_64`, `ECS`, `EC2`
 `testCase`: Must be an applicable test case in the `terraform/testcases` directory
 `additionalValues`: For `EC2` tests it is expected that the `testing_ami` value is provided.
 For ECS tests the `launch_type` variable is expected. For `EKS-arm64` tests it is expected that
--- a/docs/run-testing-framework.md
+++ b/docs/run-testing-framework.md
@@ -241,7 +241,7 @@ The values for these fields are as follows
 `serviceName`: `EKS` `EKS-arm64` `EKS-fargate` `EKS-operator` `EKS-operator-arm64` `ECS` `EC2`
 `testCase`: Must be an applicable test case in the `terraform/testcases` directory
 `additionalValues`: For `EC2` tests it is expected that the `testing_ami` value is provided.
-For ECS tests the `launch_type` variable is expected. For `EKS-arm64` and `EKS-operator-arm64` tests it is expected that
+For ECS tests the `launch_type` variable is expected. For `EKS-arm64` tests it is expected that
 a pipe delimited string of `region|clustername|amp_endoint` is provided.
 
 It is also expected that`TF_VAR_aoc_version` and `TF_VAR_aoc_image_repo` are set to valid values
--- a/docs/run-testing-framework.md
+++ b/docs/run-testing-framework.md
@@ -238,10 +238,10 @@ serviceNameN testCaseN additionalValuesN
 ```
 
 The values for these fields are as follows
-`serviceName`: `EKS` `EKS-arm64` `EKS-fargate` `EKS-operator` `ECS` `EC2`
+`serviceName`: `EKS` `EKS-arm64` `EKS-fargate` `EKS-operator` `EKS-operator-arm64` `ECS` `EC2`
 `testCase`: Must be an applicable test case in the `terraform/testcases` directory
 `additionalValues`: For `EC2` tests it is expected that the `testing_ami` value is provided.
-For ECS tests the `launch_type` variable is expected. For `EKS-arm64` tests it is expected that
+For ECS tests the `launch_type` variable is expected. For `EKS-arm64` and `EKS-operator-arm64` tests it is expected that
 a pipe delimited string of `region|clustername|amp_endoint` is provided.
 
 It is also expected that`TF_VAR_aoc_version` and `TF_VAR_aoc_image_repo` are set to valid values
--- a/terraform/executeTerraformCleanup.sh
+++ b/terraform/executeTerraformCleanup.sh
@@ -14,7 +14,7 @@
 # $1: aws_service
 # $2: testcase 
 # $3: ECS/EC2 only - ami/ecs launch type 
-# $3: For EKS-arm64 we expect region|clustername|amp_endoint
+# $3: For EKS-arm64 and EKS-operator-arm64 we expect region|clustername|amp_endoint
 
 ##########################################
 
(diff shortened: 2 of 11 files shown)


tools/batchTestGenerator/internal/testCaseBuilder.go
```diff
@@ -85,7 +85,7 @@ func buildTestCases(runConfig RunConfig) ([]TestCaseInfo, error) {
 						}
 						newTests = append(newTests, newTest)
 					}
-				case "EKS_ARM64":
+				case "EKS_ARM64", "EKS_ADOT_OPERATOR_ARM64":
```
erichsueh3: One question here is should `EKS_ADOT_OPERATOR_ARM64` be added in this case, with `additionalVar: runConfig.EksARM64Vars` or should it be in the next case, with `additionalVar: runConfig.EksVars`? They're the same variables from my understanding (`ampEndpoint`, `clusterName`, and `region`) but will have different values depending on if it is arm64 or not

erichsueh3: This has changed, since operator arm 64 tests do not need the arm64 endpoint/cluster/region variables as it uses the same endpoint/region as the EKS tests, and the cluster is set as a `-var=` flag during `executeTerraformTest.sh` since that takes [precedence](https://www.terraform.io/language/values/variables#variable-definition-precedence) over the [`parameters.tfvars`](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/testcases/otlp_metric_adot_operator/parameters.tfvars#L7) used to set the EKS cluster name in a non-arm64 operator test case

EDIT: `parameters.tfvars` is provided as a `-var-file` option on the command line which has the same precedence as a `-var` flag, but if the `-var` option is provided after the `-var-file` option it will have the correct oreder of precedence

docs/run-testing-framework.md
```diff
@@ -271,4 +271,4 @@ cd terraform/add_on/remote_state && terraform init && terraform apply \
    -var="testing_id={{test case unique id}}" \
    -var="folder_name={{folder name when uploading to s3}} \
    -var="platform={{platform running (ec2, ecs, eks, canary,...)"\
-````
\ No newline at end of file
+````
```
bryan-aguilar, line 274:
```suggestion
```
```

File: terraform/executeTerraformCleanup.sh [hunk 3]
```diff
@@ -59,9 +61,9 @@ esac
 
 
 case "$service" in
-    "EKS_FARGATE" | "EKS_ADOT_OPERATOR") terraform destroy --auto-approve $opts;
+    "EKS_FARGATE" | "EKS_ADOT_OPERATOR" | "EKS_ADOT_OPERATOR_ARM64") terraform destroy --auto-approve $opts;
     ;;
 *)
     terraform destroy --auto-approve;
 ;;
-esac
```
bryan-aguilar: Missing newline

terraform/executeTerraformTest.sh
```diff
@@ -54,6 +54,9 @@ case "$service" in
     ;;
     "EKS_ADOT_OPERATOR") TEST_FOLDER="./eks/";
     ;;
+    "EKS_ADOT_OPERATOR_ARM64") TEST_FOLDER="./eks/"
+        opts+=" -var=eks_cluster_name=arm64-adot-op-cluster";
+    ;;
```
bryan-aguilar, line 60:
```suggestion
    "EKS_ADOT_OPERATOR_ARM64") TEST_FOLDER="./eks/"
        export TF_VAR_eks_cluster_name=arm64-adot-op-cluster;
    ;;
```
I think we should stay consistent with how we set the `eks_cluster_name` variable in this file.

bryan-aguilar (commented):
Half way through reviewing, will come back to it later today

terraform/executeTerraformTest.sh
erichsueh3: The issue here is that for operator tests, `eks_cluster_name` is passed in through a `.tfvars` file (see [here](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/executeTerraformTest.sh#L31) or [this example](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/testcases/prometheus_static_adot_operator/parameters.tfvars#L10)). This takes [precedence](https://www.terraform.io/language/values/variables#variable-definition-precedence) over environment variables, so we cannot do an `export TF_VAR_...` for setting the arm64 operator cluster name

erichsueh3: To add on to this, the `-var-file` option that provides `parameters.tfvars` would be provided before the `-var=eks_cluster_name` option, which would follow the correct order of precedence for us to set `arm64-adot-op-cluster` as the name

terraform/executeTerraformCleanup.sh
erichsueh3: Seems like it's added in the line below

docs/run-testing-framework.md
erichsueh3: Not sure what this is supposed to be - if it's to add a newline then that should be added if not already present

terraform/executeTerraformTest.sh
bryan-aguilar: I feel a little conflicted on this choice. On one hand I understand why the env var won't work, but on the other it's a little confusing to look at from an outsiders perspective. I would be wondering "why is it done one way here and a different way below". 

I think we have two options
1. Add some documentation explaining this choice, a comment will probably do
2. Remove the cluster name from the existing parameters files so that we can uniformly use env vars. 

I think another downside to this is that we have a hard dependency for our arm-64 cluster names to use this specific name. It's not very portable between accounts but I am not sure if we are going for portability with this script. To be fair, all EKS tests other than the `EKS_ARM64` tests have this issue because we are using default values from the terraform modules.

terraform/executeTerraformCleanup.sh
```diff
@@ -48,6 +48,8 @@ case "$service" in
     ;;
     "EKS_ADOT_OPERATOR") TEST_FOLDER="./eks/";
     ;;
+    "EKS_ADOT_OPERATOR_ARM64") TEST_FOLDER="./eks/";
```
bryan-aguilar, line 52:
```suggestion
    "EKS_ADOT_OPERATOR_ARM64") TEST_FOLDER="./eks/";
```
Don't we also need to set the cluster name here?

docs/run-testing-framework.md
```diff
@@ -238,7 +238,7 @@ serviceNameN testCaseN additionalValuesN
 ```
 
 The values for these fields are as follows
-`serviceName`: `EKS` `EKS-arm64` `EKS-fargate` `EKS-operator` `ECS` `EC2`
+`serviceName`: `EKS` `EKS-arm64` `EKS-fargate` `EKS-operator` `EKS-operator-arm64` `ECS` `EC2`
```
bryan-aguilar:
```suggestion
`serviceName`: `EKS`, `EKS_ARM64`, `EKS_FARGATE`, `EKS_ADOT_OPERATOR`, `EKS_ADOT_OPERATOR_ARM_64`, `ECS`, `EC2`
```

bryan-aguilar: I think this was a mistake in the original documentation. These values should match the `switch` statement in the bash script right?

terraform/executeTerraformCleanup.sh
erichsueh3: Unsure - if yes, does cluster name also need to be set in for `EKS_ARM64` test cleanup? Does it need to be set in `EKS_ADOT_OPERATOR` test cleanup too?

EDIT: answer is yes for operator, and likely no for arm64, since it currently doesn't set cluster name in cleanup. Adding on to this, the operator cluster name will be included in `opts`, so that should be taken care of already

docs/run-testing-framework.md
erichsueh3: Fixed

terraform/executeTerraformTest.sh
erichsueh3: Update here, I've filed #784 that changes the usage of environment variables to set terraform variables in the script to use the `-var` option instead. It also addresses a few other things as well. The hard dependency for operator/arm64 operator tests to use these cluster names still exists, but it's slightly better than having `eks_cluster_name` set in multiple `parameters.tfvars` files.

bryan-aguilar (approved):

PR #793, opened 2022-08-15
Adding ECS_HEALTH_CHECK test case

PaurushGarg:
Adding ECS_HEALTH_CHECK test case. Remaining work:

1. Create a PR to Link testcase to ADOT Collector Repo- Dependent upon approval of another PR (https://github.com/aws-observability/aws-otel-collector/pull/1285) in the ADOT Collector repo.
2. Command used to run this testcase (includes new collector image): 
```
terraform apply -auto-approve \
    -var="aoc_image_repo=public.ecr.aws/q1w6y6d0/adotcrossprom" \
    -var="aoc_version=latest" \
    -var="testcase=../testcases/ecshealthcheck" \
    -var-file="../testcases/ecshealthcheck/parameters.tfvars"
```


validator/src/main/java/com/amazon/aoc/services/TaskService.java
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * Method to call Amazon AWS listTask API.
+   * @param clusterArn Cluster Arn
+   * @return
+   */
+  public String listTasks(String clusterArn) {
+    ListTasksRequest listTasksRequest = new ListTasksRequest().withCluster(clusterArn);
+    ListTasksResult result =  client.listTasks(listTasksRequest);
+    if (result != null && !result.getTaskArns().isEmpty()) {
+      return result.getTaskArns().get(0);
```
Aneurysm9: This seems to assume that this cluster will only ever have a single task.  Is this a valid assumption?  What gives us confidence it is valid?

validator/src/main/java/com/amazon/aoc/validators/ECSHealthCheckValidator.java
```diff
@@ -0,0 +1,80 @@
...
+  public void init(Context context, ValidationConfig validationConfig,
+                   ICaller caller, FileConfig expectedDataTemplate) throws Exception {
+    this.context = context;
+    this.expectedMetric = expectedDataTemplate;
+  }
+
+  @Override
+  public void validate() throws Exception {
+    log.info("allow sample app load balancer to start");
+    TimeUnit.SECONDS.sleep(15);
```
Aneurysm9: Do other validation methods do this?  It feels wrong.  The validation shouldn't start until all necessary resources are in place.

rapphil:
>Create a PR to Link testcase to ADOT Collector Repo- Dependent upon approval of another PR in ADOT Collector Repo.


What PR? can you provide link in the description?

validator/src/main/resources/validations/healthcheck-metric-validation.yml
```diff
@@ -0,0 +1,3 @@
+-
+  validationType: "ecs-describe-task"
+  callingType: "none"
```
Aneurysm9: Ensure all files end with a newline.  Many are not.

validator/src/main/java/com/amazon/aoc/validators/ECSHealthCheckValidator.java
bryan-aguilar: Other validations methods implement lots of retries to wait for resources to initialize.

terraform/testcases/ecshealthcheck/otconfig.tpl
```diff
@@ -0,0 +1,30 @@
+extensions:
+  health_check:
+  health_check/1:
+    endpoint: "localhost:13133"
+    path: "/health/status"
+    check_collector_pipeline:
+      enabled: true
+      interval: "5m"
+      exporter_failure_threshold: 5
```
bryan-aguilar: Why do we need multiple health check extensions? Especially since `healthcheck/1` isn't defined in the service.

validator/src/main/java/com/amazon/aoc/validators/ECSHealthCheckValidator.java
```diff
@@ -0,0 +1,80 @@
...
+    log.info("[ECSHealthCheckValidator] start validating ECS Health Check");
+    AtomicBoolean validationSuccessFlag = new AtomicBoolean(false);
+    RetryHelper.retry((retryCount == 0 ? DEFAULT_MAX_RETRY_COUNT : retryCount),
+        CHECK_INTERVAL_IN_MILLI, true, () -> {
+        if (context != null && context.getEcsContext() != null
```
rapphil: can we replace all this nesting by early returns? 

e.g.: 
```java
if (context == null || context.getEcsContext() == null || ontext.getEcsContext().getEcsClusterArn() == null) {
          throw new BaseException(
                  ExceptionCode.ECS_RESOURCES_NOT_READY,
                  "[ECSHealthCheckValidator] Awaiting on ECS Resources to be ready");
}
DescribeTasksResult result =
                  taskService.describeTask(context.getEcsContext().getEcsClusterArn());
if (result == null || result.getTasks() == null || result.getTasks().isEmpty()) {
          throw new BaseException(
                  ExceptionCode.ECS_RESOURCES_NOT_READY,
                  "[ECSHealthCheckValidator] Awaiting on ECS Resources to be ready");
}
Task task = result.getTasks().get(0);
if (task == null  || task.getContainers().isEmpty()) {
          throw new BaseException(
                  ExceptionCode.ECS_RESOURCES_NOT_READY,
                  "[ECSHealthCheckValidator] Awaiting on ECS Resources to be ready");
}
List<Container> containers = task.getContainers().stream()
                      .filter(container -> container.getName().equalsIgnoreCase("aoc-collector"))
                      .collect(Collectors.toList());
.....

```

You can transform 
```java
throw new BaseException(
                  ExceptionCode.ECS_RESOURCES_NOT_READY,
                  "[ECSHealthCheckValidator] Awaiting on ECS Resources to be ready");

``` 
in a method if you need to.

PaurushGarg:
> > Create a PR to Link testcase to ADOT Collector Repo- Dependent upon approval of another PR in ADOT Collector Repo.
> 
> What PR? can you provide link in the description?

Thanks. I have now updated the description.

validator/src/main/java/com/amazon/aoc/services/TaskService.java
PaurushGarg: Thanks. I have updated the code now to not take this assumption.

validator/src/main/java/com/amazon/aoc/validators/ECSHealthCheckValidator.java
PaurushGarg: Thanks. I have updated the code. I left the `log.info` and `time.sleep` by mistake in the code. I have removed it now. ECS resources are created before the validator.

validator/src/main/resources/validations/healthcheck-metric-validation.yml
PaurushGarg: Thanks. I have updated the code now.

terraform/testcases/ecshealthcheck/otconfig.tpl
PaurushGarg: Thanks. I have updated the code now.

validator/src/main/java/com/amazon/aoc/validators/ECSHealthCheckValidator.java
PaurushGarg: Thanks. Yes, we can do that. Actually, its easier to write the passing case, and then the complement of that case goes for exception (retries). With early return (as you described) we need to catch all the negative cases so that the complement (single case- positive) passes the test. I have updated the code as per this comment, as it becomes more readable, debug-able. I tried to incorporate all the negative cases and catch all exceptions.

rapphil (approved):

terraform/testcases/ecshealthcheck/ecs_taskdef.tpl
```diff
@@ -0,0 +1,148 @@
...
+            "value": "${region}"
+        },
+        {
+          "name": "INSTANCE_ID",
+          "value": "${testing_id}"
+        },
+        {
+        "name": "OTEL_RESOURCE_ATTRIBUTES",
+        "value": "service.namespace=${otel_service_namespace},service.name=${otel_service_name}"
+        },
```
Aneurysm9:
```suggestion
        {
            "name": "OTEL_RESOURCE_ATTRIBUTES",
            "value": "service.namespace=${otel_service_namespace},service.name=${otel_service_name}"
        },
```

Check the alignment of these and other values.  Doesn't affect the functionality but it is easier to scan as a human when they're consistent.

terraform/testcases/ecshealthcheck/ecs_taskdef.tpl
```diff
@@ -0,0 +1,148 @@
...
+            "value": "${sample_app_listen_address}"
+        },
+        {
+        "name": "JAEGER_RECEIVER_ENDPOINT",
+        "value": "127.0.0.1:${http_port}"
+        },
+        {
+        "name": "ZIPKIN_RECEIVER_ENDPOINT",
+        "value": "127.0.0.1:${http_port}"
+        },
```
Aneurysm9: Are these needed?  If so, they probably need to have different ports.

terraform/testcases/ecshealthcheck/otconfig.tpl
```diff
@@ -0,0 +1,25 @@
...
+      loglevel: debug
+service:
+  pipelines:
+    metrics:
+     receivers: [prometheus]
+     exporters: [logging]
+  extensions: [health_check]
+  telemetry:
+    logs:
+      level: debug
```
Aneurysm9:
```suggestion
      level: ${log_level
```

A new variable is available to control this behavior and should be used here.

PaurushGarg: Thanks. I have updated code accordingly.

terraform/testcases/ecshealthcheck/ecs_taskdef.tpl
PaurushGarg: Thanks. I have updated code accordingly.

terraform/testcases/ecshealthcheck/ecs_taskdef.tpl
PaurushGarg: Thanks. I have updated code accordingly.

terraform/testcases/ecshealthcheck/otconfig.tpl
```diff
@@ -0,0 +1,25 @@
...
+    config:
+      global:
+        scrape_interval: 15s
+      scrape_configs:
+      - job_name: "test-prometheus-sample-app"
+        static_configs:
+        - targets: [ ${sample_app_listen_address_host}:${sample_app_listen_address_port} ]
+exporters:
+  logging:
+      loglevel: debug
```
bryan-aguilar, line 16: Does this need to use the log level variable also?

PaurushGarg: In my opinion, it's not required here. for ref: https://github.com/aws-observability/aws-otel-test-framework/blob/812a4eef108a878c28f76f8493b4e1b38f590370/terraform/testcases/otlp_metric_mock/otconfig.tpl

bryan-aguilar (approved):

PR #801, opened 2022-08-23
Adding code to allow multi-arch imagebuild

PaurushGarg:
This PR allows multi-arch docker image build for sample apps and mocked servers. This is required to run this testsuite from new darwin/ARM64 systems.

### DIFF
--- a/terraform/imagebuild/main.tf
+++ b/terraform/imagebuild/main.tf
@@ -56,7 +56,7 @@ resource "null_resource" "build_and_push_sample_apps" {
   count = length(local.sample_apps_dockerfile_list)
 
   provisioner "local-exec" {
-    command = "docker build -t ${data.aws_ecr_repository.sample_app.repository_url}:${dirname(element(local.sample_apps_dockerfile_list, count.index))}-latest ../../sample-apps/${dirname(element(local.sample_apps_dockerfile_list, count.index))}/"
+    command = "docker buildx build --platform ${var.platform_arch} -t ${data.aws_ecr_repository.sample_app.repository_url}:${dirname(element(local.sample_apps_dockerfile_list, count.index))}-latest ../../sample-apps/${dirname(element(local.sample_apps_dockerfile_list, count.index))}/"
   }
 
   provisioner "local-exec" {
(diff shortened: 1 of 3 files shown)


vasireddy99 (approved):

bryan-aguilar:
Why do we not use `buildx` to build [multiplatform](https://docs.docker.com/build/buildx/multiplatform-images/) images? 

edit/rephrase: Should use buildx to build multiplatform images instead of targeting a single platform?

Aneurysm9:
> Why do we not use `buildx` to build [multiplatform](https://docs.docker.com/build/buildx/multiplatform-images/) images?
> 
> edit/rephrase: Should use buildx to build multiplatform images instead of targeting a single platform?

Better yet, how about we separate sample app build and deployment from test execution.  Are we even using the sample apps deployed in this manner anymore?

bryan-aguilar:
> > Why do we not use `buildx` to build [multiplatform](https://docs.docker.com/build/buildx/multiplatform-images/) images?
> > edit/rephrase: Should use buildx to build multiplatform images instead of targeting a single platform?
> 
> Better yet, how about we separate sample app build and deployment from test execution. Are we even using the sample apps deployed in this manner anymore?

This is not done at test execution time. The `imagebuild` module is [executed when setting up a new account to run tests](https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/docs/setup-basic-components-in-aws-account.md#3-build-sample-app-images).

PaurushGarg:
> Why do we not use `buildx` to build [multiplatform](https://docs.docker.com/build/buildx/multiplatform-images/) images?
> 
> edit/rephrase: Should use buildx to build multiplatform images instead of targeting a single platform?

Hi. Thanks. I have now updated the code to build multi-platform images (rather than just cross-platform). Please review.

terraform/imagebuild/main.tf
```diff
@@ -56,11 +56,11 @@ resource "null_resource" "build_and_push_sample_apps" {
   count = length(local.sample_apps_dockerfile_list)
 
   provisioner "local-exec" {
-    command = "docker build -t ${data.aws_ecr_repository.sample_app.repository_url}:${dirname(element(local.sample_apps_dockerfile_list, count.index))}-latest ../../sample-apps/${dirname(element(local.sample_apps_dockerfile_list, count.index))}/"
-  }
-
-  provisioner "local-exec" {
-    command = "docker push ${data.aws_ecr_repository.sample_app.repository_url}:${dirname(element(local.sample_apps_dockerfile_list, count.index))}-latest"
+    command = <<-EOT
+      docker buildx create --use --name sapbuilder.${count.index} --platform=linux/arm64,linux/amd64 --driver  docker-container --bootstrap
```
bryan-aguilar, line 60:
```suggestion
      docker buildx create --use --name sapbuilder.${count.index} --platform=linux/arm64,linux/amd64 --driver  docker-container --bootstrap
```
What is the reason for having to create a new builder instance for each sample app? Is it because these builds are done in parallel and thus we require a builder for each one?

PaurushGarg: Yes. that's correct. Had to create distinct builders for this reason.

vasireddy99 (approved):

bryan-aguilar (approved):

PR #804, opened 2022-08-24
Added standard validation files

alexperez52:
Added validation files for standard sample app validations.
Validations added:
- Metric validations
- HTTP call trace validations
- AWS SDK call trace validations


terraform/testcases/standard_otlp_metric_trace/otconfig.tpl
```diff
@@ -0,0 +1,31 @@
+extensions:
+  pprof:
+    endpoint: 0.0.0.0:1777
```
bryan-aguilar, line 3: Why do we need pprof?

terraform/testcases/standard_otlp_metric_trace/otconfig.tpl
```diff
@@ -0,0 +1,31 @@
...
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:${grpc_port}
+
+processors:
+  batch:
+
+exporters:
+  logging:
+    loglevel: debug
```
bryan-aguilar, line 15: There is a new flag in the test framework that will control this. see other configs for an example.

terraform/testcases/standard_otlp_metric_trace/parameters.tfvars
```diff
@@ -0,0 +1,6 @@
+validation_config = "standard-otel-trace-metric-validation.yml"
+
+soaking_data_mode = "trace"
+
+# Replace with ecr standard go sample app image when available
+sample_app_image = ""
```
bryan-aguilar: Do we expect the images to be available on ECR or can we build them locally in the test environment?

bryan-aguilar: Also, this standard template should be used with a variety of sample apps right? Should this value be set at test runtime and not explicitly defined in the parameters file?

validator/src/main/resources/expected-data-template/manual-instrumentation/standardExpectedAWSSDKTrace.mustache
```diff
@@ -0,0 +1,24 @@
...
+        "default": {
+          "port": {{port}},
+          "host": {{host}},
+          "language": {{language}},
+          "signal": "trace"
+        }
+      }
+    }
+  ]
+}]
```
bryan-aguilar: nit: newline needs to be fixed at EOF (end of file).

terraform/testcases/standard_otlp_metric_trace/otconfig.tpl
humivo: I believe pprof was added because Alex was building the otconfig based on the otlp_trace and otlp_metric testcases which include pprof.  pprof can be used to collector performance profiles and investigate issues with the Collector but if we decide it is not necessary then I can remove it.

terraform/testcases/standard_otlp_metric_trace/otconfig.tpl
humivo: Can you point me to this new flag in the test framework?  I can't seem to find it in the other configs

terraform/testcases/standard_otlp_metric_trace/parameters.tfvars
humivo: We can explore both options when it comes to ECR vs. building images locally, and I can edit this behavior depending on what we decide.  If the images were available on ECR, I believe it would be easier to just pull the different images in the workflows and allow people to easily grab the sample app images if needed.  However, it would require more maintenance in ECR and takes up resources.  If the images were built locally in the test environment then that maintenance would not be required and resources are saved, but the workflow would require more work to build each different image and people would have to learn how to build each respective image on their own if needed.  What do you think would be the best path considering this?

humivo: Yes this standard template should be used with a variety of sample apps, so this value should be set at test runtime

terraform/testcases/standard_otlp_metric_trace/otconfig.tpl
alexperez52: My apologies for the late reply. But yes, Huy is correct.

terraform/testcases/standard_otlp_metric_trace/parameters.tfvars
alexperez52: The original idea was to pull the images from ECR where they would be maintained. But it makes sense to also build them locally in order to know that an image passes all testing to then being updated or uploaded to ECR.

terraform/testcases/standard_otlp_metric_trace/otconfig.tpl
bryan-aguilar: Is it necessary?

terraform/testcases/standard_otlp_metric_trace/otconfig.tpl
bryan-aguilar: https://github.com/aws-observability/aws-otel-test-framework/blob/23f605c6125b4465c5961b292f01f8bccf228e7d/terraform/testcases/otlp_trace/otconfig.tpl#L29

bryan-aguilar: Although that is being used to control the telemetry log level and not the logging exporter. I am not sure if we should be using it for both. So I may be off base here.

terraform/testcases/standard_otlp_metric_trace/parameters.tfvars
```diff
@@ -0,0 +1,3 @@
+validation_config = "standard-otel-trace-metric-validation.yml"
+
+soaking_data_mode = "trace"
```
bryan-aguilar, line 3: What is this used for? Are we planning on doing soaking tests for this test case?

bryan-aguilar (approved):

terraform/testcases/standard_otlp_metric_trace/otconfig.tpl
humivo: I found this [note](https://github.com/humivo/aws-otel-test-framework#:~:text=Note%3A%20Please%20enable%20pprof%20extension%20in%20your%20otconfig.tpl%20for%20the%20debugging%20purpose.) in the test framework README advising to enable pprof extension

PR #856, opened 2022-09-30
Update distribution support for Testing Framework

Kausik-A:
**Description:** 

This PR introduces various updates for the `EC2 instance distributions` that the ADOT Testing Framework supports.The changes include:

- A simplified regex pattern to search for new AMIs.
- Remove AMI Id hardcoding.
- Include new versions of the Image distributions and deprecate older ones ( Ex Ubuntu 22 ).
- Add ARM images for some distributions.
- `Canary` and `Performance tests` now use `Windows 2022 `and `Amazon Linux 2`.
- Deprecate CentOS.

**Testing:** 

Tested locally by cross checking the terraform output AMI Id with AWS-EC2-AMI Console results.

### DIFF
--- a/terraform/canary/amis.tf
+++ b/terraform/canary/amis.tf
@@ -60,7 +60,7 @@ variable "amis" {
     }
     canary_linux = {
       os_family          = "amazon_linux"
-      ami_search_pattern = "amzn2-ami-kernel*"
+      ami_search_pattern = "amzn2-ami-kernel.5*"
       ami_owner          = "amazon"
       ami_product_code   = []
       family             = "amazon_linux"
--- a/terraform/canary/amis.tf
+++ b/terraform/canary/amis.tf
@@ -52,21 +52,19 @@ variable "amis" {
   default = {
     canary_windows = {
       os_family          = "windows"
-      ami_search_pattern = "Windows_Server-2019-English-Full-Base-*"
+      ami_search_pattern = "Windows_Server-2022-English-Full-Base*"
       ami_owner          = "amazon"
-      ami_id             = "ami-01f14dc60171d8d7b"
       ami_product_code   = []
       family             = "windows"
-      arch               = "amd64"
+      arch               = "x86_64"
     }
     canary_linux = {
       os_family          = "amazon_linux"
-      ami_search_pattern = "amzn2-ami-hvm-2.0.????????.?-x86_64-gp2"
+      ami_search_pattern = "amzn2-ami-kernel*"
       ami_owner          = "amazon"
-      ami_id             = "ami-0c2ab3b8efb09f272"
       ami_product_code   = []
       family             = "amazon_linux"
-      arch               = "amd64"
+      arch               = "x86_64"
     }
   }
 }
--- a/terraform/ec2/amis.tf
+++ b/terraform/ec2/amis.tf
@@ -92,13 +92,26 @@ variable "amis" {
     # Ubuntu Distribution
     ubuntu18 = {
       os_family          = "ubuntu"
-      ami_search_pattern = "ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-*"
-      ami_owner          = "099720109477"
-      ami_id             = "ami-0688ba7eeeeefe3cd"
+      ami_search_pattern = "ubuntu/images/hvm-ssd/ubuntu-bionic-*"
+      owners             = "aws-marketplace"
       ami_product_code   = []
       family             = "debian"
-      arch               = "amd64"
+      arch               = "x86_64"
       login_user         = "ubuntu"
+      user_data          = <<EOF
+#! /bin/bash
+sudo snap refresh amazon-ssm-agent
+EOF
+    }
+    arm_ubuntu18 = {
+      os_family          = "ubuntu"
+      ami_search_pattern = "ubuntu/images/hvm-ssd/ubuntu-bionic-*"
+      owners             = "aws-marketplace"
+      ami_product_code   = []
+      family             = "debian"
+      arch               = "arm64"
+      login_user         = "ubuntu"
+      instance_type      = "t4g.nano"
       user_data          = <<EOF
 #! /bin/bash
 sudo snap refresh amazon-ssm-agent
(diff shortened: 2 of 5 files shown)


terraform/canary/amis.tf
```diff
@@ -52,21 +52,19 @@ variable "amis" {
...
-      ami_id             = "ami-01f14dc60171d8d7b"
       ami_product_code   = []
       family             = "windows"
-      arch               = "amd64"
+      arch               = "x86_64"
     }
     canary_linux = {
       os_family          = "amazon_linux"
-      ami_search_pattern = "amzn2-ami-hvm-2.0.????????.?-x86_64-gp2"
+      ami_search_pattern = "amzn2-ami-kernel*"
```
Aneurysm9: Do we need to choose a kernel versions here, or does this pattern only catch the kernel 5.x AMIs?

Kausik-A: It catches the latest kernel version. Currently it is 5.x; but it should ideally pick 6.x if `amzn2-ami-kernel.6.x` is out

terraform/ec2/amis.tf
```diff
@@ -89,46 +89,137 @@ EOF
...
       os_family          = "ubuntu"
-      ami_search_pattern = "ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-*"
-      ami_owner          = "099720109477"
-      ami_id             = "ami-0688ba7eeeeefe3cd"
+      ami_search_pattern = "ubuntu/images/hvm-ssd/ubuntu-bionic-*"
+      owners             = "aws-marketplace"
       ami_product_code   = []
       family             = "debian"
-      arch               = "amd64"
+      arch               = "x86_64"
```
bryan-aguilar, line 99: The `arm_ubuntu18` object declares an instance type. Why doesn't `ubunutu18`? What is the default?

terraform/ec2/amis.tf
```diff
@@ -89,46 +89,137 @@ EOF
...
-      # NOTE: we need product code to pick the right debian 10.
-      ami_product_code = [
-      "auhljmclkudu651zy27rih2x2"]
-      family     = "debian"
-      arch       = "amd64"
-      login_user = "admin"
-      user_data  = <<EOF
+      ami_search_pattern = "debian-10-*"
+      ami_owner          = "aws-marketplace"
+      ami_product_code   = []
```
bryan-aguilar, line 218: The previous version had a comment about the product code. In your testing did the "correct" debian10 get selected?

bryan-aguilar (commented):
I see that we change `amd64` to `x86_64` in several places. It looks like this removes the need for the local variable translator. Does any other part of the `ec2` module rely on the `amd64` value?

terraform/ec2/amis.tf
Kausik-A: The default values come from [here](https://github.com/aws-observability/aws-otel-test-framework/blob/45ee692bac2ca73baa0132c6fcc4e122dc8cdd3f/terraform/ec2/amis.tf#L21) based on the `family` attribute. For ARM images we override to `t4g.nano `because`t2.micro` is not supported to run ARM binaries

terraform/ec2/amis.tf
Kausik-A: As we use `aws-marketplace` as the `ami_owner`, they have the supported debian images. SO product code is not required now

Kausik-A: Screen Shot 2022-10-04 at 11 33 33 AM

Screen Shot 2022-10-04 at 11 35 58 AM

Kausik-A:
> I see that we change `amd64` to `x86_64` in several places. It looks like this removes the need for the local variable translator. Does any other part of the `ec2` module rely on the `amd64` value?

Searching through the codebase, it looks like no other part of the ec2 module rely on `amd64` anymore.

terraform/canary/amis.tf
Aneurysm9: Does it catch the latest kernel version, or the latest AMI?  If there are both `kernel.5.x` and `kernel.6.x` AMIs available would it consistently get the 6.x version or could it get 5.x one time and 6.x another depending on which was published most recently?

Do we still support the 4.x AL2 releases?  I believe those are the ones with no `kernel` in the AMI name.  Should we have two entries, one for `hvm-*` and one for `kernel.5*`?  We can always add another for `kernel.6*` if it becomes necessary, though I expect that won't happen on AL2.

terraform/ec2/amis.tf
Aneurysm9: We should probably be consistent with `micro` or `nano`.  If we're using `t2.micro` let's also use `t4g.micro` so that RAM and burst performance are equivalent.

terraform/canary/amis.tf
Kausik-A: >Does it catch the latest kernel version, or the latest AMI? If there are both kernel.5.x and kernel.6.x AMIs available would it consistently get the 6.x version or could it get 5.x one time and 6.x another depending on which was published most recently?

That is a good point, I did not think about that edge case. Even-though idk how to exactly test out this behavior ; the terraform documentation mentions that it pulls the` recent results`. So it will probably pull the `kernel.5.1.x` if it is updated after `kernel.6.x`. I think it is better to update the search patter to `amzn2-ami-kernel.5*` and update to `amzn2-ami-kernel.6*` when its out.

>Do we still support the 4.x AL2 releases? I believe those are the ones with no kernel in the AMI name. Should we have two entries, one for hvm-* and one for kernel.5*? We can always add another for kernel.6* if it becomes necessary, though I expect that won't happen on AL2.

Should we ? The initial plan was just to test out the latest version of AL2. The [release notes ](https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-linux-2-ami-kernel-5-10/)for kernal 5 mention various security improvements. So IMO we should suggest users to upgrade.

terraform/ec2/amis.tf
Kausik-A: Updated in [6e23b18](https://github.com/aws-observability/aws-otel-test-framework/pull/856/commits/6e23b182746786a94cd7da63fe6418cc2eee05a6)

terraform/canary/amis.tf
Kausik-A: Updated the search pattern in [ef1a2a0](https://github.com/aws-observability/aws-otel-test-framework/pull/856/commits/ef1a2a0a0d6c13acfb8d84f54fbc0bfa62f41e4c)

Aneurysm9 (approved):

bryan-aguilar (approved):

PR #863, opened 2022-10-10
Change arch name from x86_64 to amd64

Kausik-A:
**Description:** 

After #856 the canary tests inside aws-otel-collector started failing so reverting the default `arch` variable name to `amd64`. This is because the s3 buckets use this naming terminology, which we use to pull the latest collector image from 

### DIFF
--- a/terraform/ec2/amis.tf
+++ b/terraform/ec2/amis.tf
@@ -96,7 +96,7 @@ variable "amis" {
       owners             = "aws-marketplace"
       ami_product_code   = []
       family             = "debian"
-      arch               = "x86_64"
+      arch               = "amd64"
       login_user         = "ubuntu"
       user_data          = <<EOF
 #! /bin/bash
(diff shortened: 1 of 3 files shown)


vasireddy99 (approved):

terraform/ec2/amis.tf
```diff
@@ -380,6 +379,10 @@ EOF
   }
 }
 
+# Local variables only apply to aws_ami for the customized filter input
+locals {
+  arch = var.amis[var.testing_ami]["arch"] == "amd64" ? "x86_64" : var.amis[var.testing_ami]["arch"]
```
vasireddy99, line 384: Do we still need 'x86_64' here in this case

Kausik-A: yes, the terraform ami data filter only use x86_64 as parameter instead of amd64

PR #864, opened 2022-10-10
Use local Arch variable 

Kausik-A:
**Description:** 

This is an extension to #863 to let `aws_ami` data module to use `local arch` variable instead of the `default`

### DIFF
--- a/terraform/ec2/amis.tf
+++ b/terraform/ec2/amis.tf
@@ -398,7 +398,7 @@ data "aws_ami" "selected" {
 
   filter {
     name   = "architecture"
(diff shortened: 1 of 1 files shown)


vasireddy99 (approved):

terraform/ec2/amis.tf
```diff
@@ -398,7 +398,7 @@ data "aws_ami" "selected" {
 
   filter {
     name   = "architecture"
-    values = [var.amis[var.testing_ami]["arch"]]
```
vasireddy99, line 401: Are there any other places that need this change?

Kausik-A: no

Issue #771, opened 2022-07-25
Issue with performance tests for metrics


povilasv:
If you try to run `otlp_grpc_exporter_metric` performance test:

```
terraform apply -var="data_rate=5000" -var="debug=true" -var="testcase=../testcases/otlp_grpc_exporter_metric_mock" -var="install_package_source=local" -var-file="../testcases/otlp_grpc_exporter_metric_mock/parameters.tfvar
```

It will give you:

```
{"testcase":"otlp_grpc_exporter_metric_mock","instanceType":"m5.2xlarge","receivers":["otlp"],"processors":["batch"],"exporters":["otlp"],"dataType":"otlp","dataMode":"metric","dataRate":5000,"avgCpu":0.03666620071652944,"avgMem":62.920635733333334,"maxCpu":0.19999619791228065,"maxMem":63.229952,"commitId":"dummy_commit","collectionPeriod":10,"testingAmi":"soaking_linux"}
```

Which is virtually no usage (numbers don't look correct)

I tried running load-generator and Opentelemetry Collector locally and load-generator doesn't seem to produce any data nor log anny errors:

```
./app/bin/load-generator metric -r=5000 -d=otlp -u=127.0.0.1:4317
```

Additionally, I made a sample Go app using otel sdk, which succesfully managed to send the grpc metrics to `127.0.0.1:4317`, so the issues seems to be with the load generator.

Issue #736, opened 2022-06-20
Java package imports broken in load generator 


bryan-aguilar:
PRs #721, #715, #714, #691 are failing due to a breaking change introduced in the upstream java release [`v1.15.0`](https://github.com/open-telemetry/opentelemetry-java/releases/tag/v1.15.0).

```
BREAKING: merge all stable OTLP exporters into opentelemetry-exporter-otlp. opentelemetry-exporter-otlp-trace, opentelemetry-exporter-otlp-metrics, opentelemetry-exporter-otlp-http-trace, and opentelemetry-exporter-otlp-http-metrics are no longer published and their contents have been merged into a single artifact.
BREAKING: merge log OTLP exporters into opentelemetry-exporter-otlp-logs. opentelemetry-exporter-otlp-http-logs is no longer published and its contents have been merged into a single artifact.
```

Additionally:

- [x] Fix issue: https://github.com/aws-observability/aws-otel-test-framework/issues/771
- [x] Fix issue: `--dataRate` parameter is currently not generating output as described in `readme`.
- [x] Add flexibility to create multiple `metrics` as well multiple `datapoints`.
- [x] Update Docker Image in Public ECR.

PaurushGarg:
Closing this issue, as relevant PR for this issue has been merged now.

PR #874, opened 2022-10-11
Changes to generate suitable load in otlp and statsd metric

PaurushGarg:
Fixes issues described in issue #771 and issue #736 

Todo: 
- [x] Analyse the  impact of this change on `StatsdMetricEmitter`
- [ ] Update dependent projects where OTLP metric of `load-generator` is used (not part of this PR).


load-generator/README.md
```diff
@@ -15,8 +15,16 @@
     1. --dataFormat -d (eg, otlp, prometheus, xray)
 
 ### OTLP Metrics Load Test Sample Command,
+
+The following is a list of optional command line flags for configuration of OTLP metrics:
+* `--flushInterval, -f` : (default=1000)the metric collection export interval
```
Aneurysm9: What is the unit here? `ms`?

load-generator/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM gradle:7.3.3-jdk11 as builder
+FROM gradle:7.5.1-jdk18 as builder
```
Aneurysm9: Does this need to increase to JDK18?

load-generator/README.md
```diff
@@ -15,8 +15,16 @@
...
+* `--metric_count, -m`: (default=1) the number of metrics that should be created for each metric type
+* `--datapointCount, -dp`: (default=1) the number of datapoints that should be created for each metric.
+* `--observationInterval, -o`: (default=1000) the interval at which metric observations/values are updated.
```
Aneurysm9: Also needs units.

load-generator/README.md
```diff
@@ -15,8 +15,16 @@
     1. --dataFormat -d (eg, otlp, prometheus, xray)
 
 ### OTLP Metrics Load Test Sample Command,
+
+The following is a list of optional command line flags for configuration of OTLP metrics:
+* `--flushInterval, -f` : (default=1000)the metric collection export interval
+* `--metric_count, -m`: (default=1) the number of metrics that should be created for each metric type
+* `--datapointCount, -dp`: (default=1) the number of datapoints that should be created for each metric.
```
Aneurysm9: Are these per-period?  How do these values relate to the other values?

load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
```diff
@@ -53,13 +56,37 @@ public void emitDataLoad() throws Exception {
 
   @Override
   public void nextDataPoint() {
-    apiBytesSentCounter
-            .add(10, METRIC_ATTRIBUTES);
-    mutateApiLatency();
+    log.info("Updating metrics for : {}" , param.getMetricType());
```
Aneurysm9: Should this be debug rather than info?  Seems like it could generate a lot of log noise.

load-generator/README.md
PaurushGarg: Thanks. Made changes accordingly.

load-generator/Dockerfile
PaurushGarg: Thanks. Revert the change.

load-generator/README.md
PaurushGarg: Thanks. Updated read me accordingly. 
Also revised the `statsdmetricemitter` to fix the similar issue of load-generation, and grouped parameters in readme accordingly.

load-generator/README.md
PaurushGarg: Thanks. Updated readme with this information.

load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
PaurushGarg: Thanks. Made changes accordingly.

load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
```diff
@@ -53,13 +56,37 @@ public void emitDataLoad() throws Exception {
...
+          for(int id=0 ; id < param.getDatapointCount(); id++) {
+            Attributes datapointAttributes = getDataPointAttributes(id);
+            counter.add(ThreadLocalRandom.current().nextInt(1,10), datapointAttributes);
+          }
+        }
+        break;
+      case "gauge":
+        for(int i=0 ; i < param.getMetricCount(); i++) {
+          for (int id = 0; id < param.getDatapointCount(); id++) {
+            this.gaugeValues.add(id, ThreadLocalRandom.current().nextLong(-100, 100));
```
rapphil: I think this will have unintended effects no? https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#add-int-E-

>Inserts the specified element at the specified position in this list. Shifts the element currently at that position (if any) and any subsequent elements to the right (adds one to their indices).

Every time that you call nextDatapoint, more data will be added to the array. if this gets called enough, you might become out of memory.

maybe use plain array `new long[param.getDataPointCount]`?

load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
```diff
@@ -53,13 +56,37 @@ public void emitDataLoad() throws Exception {
...
+          }
+        }
+        break;
```
rapphil: nit: I think you could use a Template class pattern or a strategy pattern in order to make the code cleaner instead of having to deal with these two different cases. The idea is that you instantiate a different implementation based on the metric type.

load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
```diff
@@ -53,13 +56,37 @@ public void emitDataLoad() throws Exception {
...
-    apiBytesSentCounter
-            .add(10, METRIC_ATTRIBUTES);
-    mutateApiLatency();
+    log.debug("Updating metrics for : {}" , param.getMetricType());
+
+    switch (param.getMetricType()) {
+      case "counter":
+        for(LongCounter counter: counters) {
+          for(int id=0 ; id < param.getDatapointCount(); id++) {
+            Attributes datapointAttributes = getDataPointAttributes(id);
```
bryan-aguilar, line 65: Can you attach attributes to counters at creation time or does it have to be done here?

load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
```diff
@@ -77,22 +104,52 @@ public void setupProvider() {
             GlobalOpenTelemetry.meterBuilder("aws-otel-load-generator-metric")
```
bryan-aguilar:
```suggestion
            GlobalOpenTelemetry.meterBuilder("adot-load-generator-metric")
```

load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/StatsdMetricEmitter.java
```diff
@@ -43,24 +45,37 @@ public void emitDataLoad() throws Exception {
...
+        log.debug("Updating metrics...");
+        for (int i = 0; i < this.param.getMetricCount(); i++) {
+            for (int id = 0; id < this.param.getDatapointCount(); id++) {
+                payload.append("statsdTestMetric").append(i).append(":")
+                        .append(ThreadLocalRandom.current().nextInt(-100,100))
+                        .append("|g|").append(attributes).append(id).append("\n");
+            }
+        }
+        buf = payload.toString().getBytes();
         DatagramPacket packet = new DatagramPacket(buf, buf.length, ipAddress, portAddress);
```
rapphil, line 83: Just fyi, udp packets have size limitation of 64k. I think it is ok to let the exception bubble up, but maybe we can proactively check what is the size of `buf` before calling `send` and raise an `RuntimeException` asking the user to reduce the number of metrics/datapoints.

PaurushGarg: Thanks updated.

load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
PaurushGarg: Thanks, Updated.

load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
PaurushGarg: as each datapoint has a differentiating(at least one distinct) label, so I think it has to be done here (at the time of datapoint creation).

load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
PaurushGarg: Thanks. Updated.

load-generator/src/main/java/com/amazon/opentelemetry/load/generator/emitter/OtlpMetricEmitter.java
PaurushGarg: I have not updated code with this change, as I thought it is just a simple switch case in sample-app. But do let me know if it has to be done, then I will make the necessary changes.

rapphil (approved):

PR #876, opened 2022-10-14
[batchTestGenerator] Add multi cluster support

bryan-aguilar:
**Description:** Adds the ability to dispatch a single test case to multiple EKS clusters. 

Blocked by: #873 


tools/batchTestGenerator/cmd/root.go
```diff
@@ -2,14 +2,11 @@ package cmd
 
 import (
 	"fmt"
+	batch "github.com/aws-observability/aws-otel-test-framework/batchTestGenerator/internal"
+	"github.com/spf13/cobra"
```
Aneurysm9: These shouldn't be interleaved with stdlib imports.  `goimports` can fix this, I think.

tools/batchTestGenerator/internal/config.go
```diff
@@ -1,26 +1,109 @@
...
+		for _, platform := range testCase.Platforms {
+			if _, ok := allPlatformsSet[platform]; !ok {
+				return fmt.Errorf("not a valid platform: %s", platform)
+			}
+		}
+	}
+	return nil
+}
+
+func stringArrayToMap(input []string) map[string]struct{} {
```
Aneurysm9:
```suggestion
func stringArrayToSet(input []string) map[string]struct{} {
```

Specifically here this is making a set for presence detection and not a map to any usable type.

tools/batchTestGenerator/internal/config.go
```diff
@@ -1,26 +1,109 @@
...
 	rc := RunConfig{
 		IncludedServices: ism,
 		MaxBatches:       40,
 	}
 	return rc
 }
+
+func (r *RunConfig) ValidateTestCaseInput() error {
+	// validate clustertargets array
+	awsRegionRegex, err := regexp.Compile("^(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\\d$")
```
Aneurysm9: This is a static regex, so we should be able to move it out to a package-scoped variable and create it once with `regexp.MustCompile()`.

Also, raw strings are easier to work with when making regexes:

```suggestion
	awsRegionRegex, err := regexp.Compile(`^(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d$`)
```

tools/batchTestGenerator/internal/config.go
```diff
@@ -1,26 +1,109 @@
...
+	allPlatformsSet := stringArrayToMap(allPlatforms)
+	if err != nil {
+		return fmt.Errorf("failed to build regex: %w", err)
+	}
+	for _, clusterTarget := range r.TestCaseInput.ClusterTargets {
+		if _, ok := allPlatformsSet[clusterTarget.Type]; !ok || !strings.HasPrefix(clusterTarget.Type, "EKS") {
```
Aneurysm9: This predicate could use an explanatory comment or extraction into a well-named helper function.

tools/batchTestGenerator/cmd/root.go
bryan-aguilar: First time hearing about `goimports`. Thanks!

tools/batchTestGenerator/internal/config.go
bryan-aguilar: Refactored to a function that I hope is easier to grok.

tools/batchTestGenerator/internal/config.go
```diff
@@ -1,26 +1,123 @@
...
+	"EKS_ARM64":               {},
+	"EKS_ADOT_OPERATOR":       {},
+	"EKS_ADOT_OPERATOR_ARM64": {},
+	"EKS_FARGATE":             {},
+	"CANARY":                  {},
+	"PERF":                    {},
+	"LOCAL":                   {},
+}
+
+var awsRegionRegex = regexp.MustCompile(`^(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d$`)
```
Aneurysm9:
```suggestion
var awsRegionRegex = regexp.MustCompile(`^(?:us(?:-gov)?|ap|ca|cn|eu|sa)-(?:central|(?:north|south)?(?:east|west)?)-\d$`)
```

Micro-optimization, but since you're not using the contents of these groups they can be made non-capturing.

Aneurysm9 (approved):

tools/batchTestGenerator/internal/mock_test_data/eksarm64test.json
```diff
@@ -1,24 +1,61 @@
 {
-    "tests": [
```
rapphil, line 2: why this file is not inside config? This seems to be a complete config.

tools/batchTestGenerator/internal/config.go
```diff
@@ -1,26 +1,123 @@
...
 	MaxBatches       int
 	TestCaseFilePath string
-	EksARM64Vars     string
-	EksVars          string
+	TestCaseInput    Tests
 }
 
-func NewDefaultRunConfig() RunConfig {
-	defaultServices := []string{"EKS", "ECS", "EC2", "EKS_ARM64", "EKS_ADOT_OPERATOR", "EKS_ADOT_OPERATOR_ARM64", "EKS_FARGATE"}
+type Tests struct {
```
rapphil: nit:  should we add comments and maybe a more meaningful name? Maybe `TestSuite`?

tools/batchTestGenerator/internal/config.go
```diff
@@ -1,26 +1,123 @@
...
+	Tests          []Test          `json:"tests"`
+	ClusterTargets []ClusterTarget `json:"clustertargets"`
+}
+
+type Test struct {
+	CaseName  string   `json:"case_name"`
+	Platforms []string `json:"platforms"`
+}
+
+type ClusterTarget struct {
```
rapphil, line 53: Let me see if I got the idea here. When a test has platform EKS for example, there can be multiple targets for the EKS platform and the same test will be executed multiple times?

tools/batchTestGenerator/inputFileSyntax.md
```diff
@@ -0,0 +1,64 @@
...
+        "CANARY"
+      ]
+    }
+  ]
+}
+```
+
+### clustertargets
+Define a set of clusters to dispatch EKS tests to. Each EKS platform that is included in your
+test case platforms must have cluster targets defined. Defining a test case with `EKS_FARGATE` tests
```
rapphil: can we clarify if the test is executed for all the targets?

tools/batchTestGenerator/inputFileSyntax.md
```diff
@@ -0,0 +1,64 @@
+# Input File Syntax
```
rapphil, line 1: nit: add an overview about what this file represents. E.g.: a set of tests that must be run and can be broken into smaller batches. We create combinations for each testcase and platform. Please add more details about the targets in the clustertargets section. Do we create combinations for each target and test case?

tools/batchTestGenerator/cmd/root.go
```diff
@@ -69,20 +67,18 @@ func newCommandConfig() *commandConfig {
...
+			return fmt.Errorf("error parsing input file: %w", err)
 		}
-		c.runConfig.EksVars = eksVars
 
-		if c.runConfig.MaxBatches < 1 {
-			return fmt.Errorf("max batches must be greater than 0")
+		err = c.runConfig.ValidateTestCaseInput()
+		if err != nil {
+			return fmt.Errorf("input file validation failed: %w", err)
 		}
```
rapphil: nit: maybe transform  these two into a private methods and call them in a single `Initialize` method. I mean, they have to be called in this order always right, so why leak this detail to a code that rely on it.

tools/batchTestGenerator/internal/config.go
bryan-aguilar: Yes. Each target is synonymous with an EKS cluster to run the test on.

tools/batchTestGenerator/internal/mock_test_data/eksarm64test.json
bryan-aguilar: These files are mocks for the `testCaseBuilder` tests. I have refactored them into their own directory. I hope this provides a bit more clarity.

tools/batchTestGenerator/internal/config.go
bryan-aguilar: Added extra info and renamed

tools/batchTestGenerator/inputFileSyntax.md
bryan-aguilar: Added additional context

tools/batchTestGenerator/inputFileSyntax.md
bryan-aguilar: Added some additional context

tools/batchTestGenerator/cmd/root.go
bryan-aguilar: Implemented, I think this ends up looking a bit cleaner in the test cases also.

rapphil (approved):

PR #906, opened 2022-11-03
[dev/cdk] Add ability to target k8s versions greater than v1.21.0

bryan-aguilar:
**Description:** This PR uses new CDK assets provided by the CDK library to assign the correct version of kubectl layers to EKS Clusters.


Aneurysm9:
> If this PR is just too big to look at with the prettier changes then I can revert those and submit a separate PR.

Yes, please!  As a general rule, it's always best to do large, automated rewrites/refactorings in a separate PR to make it easier for reviewers to determine whether a given change is intentional.

bryan-aguilar:
@Aneurysm9 Gotchya! I have removed the prettier changes and will submit those after this PR has been merged.

cdk_framework/EKS/lib/utils/kubectlLayer.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Construct } from "constructs/lib/construct";
+
+export function GetLayer(scope: Construct, k8sVersion : aws_eks.KubernetesVersion):  ILayerVersion{
+    switch (k8sVersion){
+      case aws_eks.KubernetesVersion.V1_22:
+        return new KubectlV22Layer(scope, 'v22Layer');
+      case aws_eks.KubernetesVersion.V1_23:
+        return  new KubectlV23Layer(scope,'v23Layer');
+      default:
+        return new KubectlLayer(scope, 'v20Layer');
```
Aneurysm9: No `v21Layer`?

cdk_framework/EKS/lib/utils/validate-interface-schema.ts
```diff
@@ -4,7 +4,7 @@ import { ec2ClusterInterface } from "../interfaces/ec2cluster-interface"
 const validateSchema = require('yaml-schema-validator')
 
 const supportedLaunchTypes = new Set(['fargate', 'ec2'])
-const supportedVersions = new Set(['1.18', '1.19', '1.20', '1.21']);
+const supportedVersions = new Set(['1.18', '1.19', '1.20', '1.21','1.22','1.23']);
```
Aneurysm9:
```suggestion
const supportedVersions = new Set(['1.21','1.22','1.23']);
```

Aren't all versions prior to 1.21 deprecated and unable to be used for new clusters?

cdk_framework/EKS/lib/utils/kubectlLayer.ts
bryan-aguilar: The v20 layer is used by default for `v1.21` and `v1.20` according to the [docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_eks-readme.html#runtime). A `v21` layer was made available last release though and I have updated the code to use that instead.

cdk_framework/EKS/lib/utils/validate-interface-schema.ts
bryan-aguilar: Good catch, thanks.

bryan-aguilar:
The readme was updated to better reflect steps required. I have added the necessary packaged to dev dependencies for user convenience.

Aneurysm9 (approved):

PR #908, opened 2022-11-07
Use latest ami id for arm_suse15 in suse OS family

vasireddy99:
**Description:** 

This PR is to use latest arm_suse15 available for the tests, using latest ami availble in community closest to the pattern since `suse-sles-15-sp2-v20200721-hvm-ssd-arm64*` doesn't any updated matching ami's available.

image

### DIFF
--- a/terraform/ec2/amis.tf
+++ b/terraform/ec2/amis.tf
@@ -300,13 +300,13 @@ EOF
 
     arm_suse15 = {
       os_family          = "suse"
-      ami_search_pattern = "suse-sles-15*"
-      ami_owner          = "amazon"
+      ami_search_pattern = "suse-sles-15-sp2-v20210212-hvm-ssd-arm64*"
+      ami_owner          = "013907871322"
+      ami_id             = "ami-0b704e615e7364682"
       ami_product_code   = []
       family             = "linux"
-      login_user         = "ec2-user"
       arch               = "arm64"
-      instance_type      = "t4g.micro"
+      instance_type      = "t4g.nano"
       user_data          = <<EOF
 #! /bin/bash
 cd /tmp
--- a/terraform/ec2/amis.tf
+++ b/terraform/ec2/amis.tf
@@ -300,13 +300,13 @@ EOF
 
     arm_suse15 = {
       os_family          = "suse"
-      ami_search_pattern = "suse-sles-15-sp2-v20200721-hvm-ssd-arm64*"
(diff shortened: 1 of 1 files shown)


bryan-aguilar (changes requested):
Is the search pattern used in this branch? I thought we had changed to using a fixed `ami_id` and did not revert back to a search pattern until post `v0.22.x`. I believe this PR as is would cause an error because we use the fixed [`ami_id`](https://github.com/aws-observability/aws-otel-test-framework/blob/6ef782319001f5ac82cb760f820a530db60c7acf/terraform/ec2/main.tf#L67) rather than the search pattern. See [this](https://github.com/aws-observability/aws-otel-test-framework/commit/e6a61633894e9706bb621498d3e74acf6f5767a3#diff-74d4b9e2f0f97a5797874df730dacab3d04096afa0b81a7cccdc630802dde5d8) commit for when that change was made for additional context.

terraform/ec2/amis.tf
```diff
@@ -300,13 +300,13 @@ EOF
...
-      ami_owner          = "amazon"
+      ami_search_pattern = "suse-sles-15-sp2-v20210212-hvm-ssd-arm64*"
+      ami_owner          = "013907871322"
+      ami_id             = "ami-0b704e615e7364682"
       ami_product_code   = []
       family             = "linux"
-      login_user         = "ec2-user"
       arch               = "arm64"
-      instance_type      = "t4g.micro"
+      instance_type      = "t4g.nano"
```
bryan-aguilar, line 309: Why was this change made?

terraform/ec2/amis.tf
```diff
@@ -300,13 +300,13 @@ EOF
 
     arm_suse15 = {
       os_family          = "suse"
-      ami_search_pattern = "suse-sles-15*"
-      ami_owner          = "amazon"
+      ami_search_pattern = "suse-sles-15-sp2-v20210212-hvm-ssd-arm64*"
+      ami_owner          = "013907871322"
```
bryan-aguilar, line 304: What owner does this reference?

bryan-aguilar: Saw the updated PR description.

terraform/ec2/amis.tf
vasireddy99: reverted this change

humivo (approved):

PR #922, opened 2022-11-14
Add workflow to update test images

rapphil:
Signed-off-by: Raphael Silva <rapphil@gmail.com>

**Description:** Add workflow to update docker images used during tests.

### DIFF
--- a/.github/workflows/update-test-images.yml
+++ b/.github/workflows/update-test-images.yml
@@ -1,0 +1,23 @@
+name: Update Test Images
+on:
+  push:
+    branches: ['terraform']
+    
+    # List of paths that should be considered when triggering the workflow
+    paths: ['mocked_servers/**']
+jobs:
+  run-image-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          role-to-assume: ${{ secrets.COLLECTOR_ASSUMABLE_ROLE_ARN }}
+          aws-region: us-west-2
+      - name: Set up terraform
+        uses: hashicorp/setup-terraform@v2
+      - name: run imagebuild
+        working-directory: terraform/imagebuild
+        run: |
+          terraform init && terraform apply --auto-approve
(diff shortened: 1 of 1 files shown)


.github/workflows/update-test-images.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          role-to-assume: ${{ secrets.COLLECTOR_ASSUMABLE_ROLE_ARN }}
+          aws-region: us-west-2
+      - name: Set up terraform
+        uses: hashicorp/setup-terraform@v2
+      - name: run imagebuild
```
bryan-aguilar, line 29: I believe this will fail since without a `gradle build` being ran first in the [`jaeger-zipkin-sample-app`](https://github.com/aws-observability/aws-otel-test-framework/tree/terraform/sample-apps/jaeger-zipkin-sample-app). Can we confirm this? I thought at one point the docker file relied on the `build` directory to exist.

.github/workflows/update-test-images.yml
```diff
@@ -0,0 +1,23 @@
+name: Update Test Images
+on:
+  push:
+    branches: ['terraform']
+    
+    # List of paths that should be considered when triggering the workflow
+    paths: ['mocked_servers/**']
```
bryan-aguilar: Should we also publish for `sample_apps`also since `imagebuild` module does both?

.github/workflows/update-test-images.yml
rapphil: good catch. I will add a command to build it.

.github/workflows/update-test-images.yml
rapphil: yes, I updated the list of watched directories.

.github/workflows/update-test-images.yml
```diff
@@ -0,0 +1,31 @@
...
+    paths: ['mocked_servers/**', 'sample-apps/**']
+jobs:
+  run-image-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          role-to-assume: ${{ secrets.TEST_FW_ASSUMABLE_ROLE_ARN }}
```
bryan-aguilar, line 16: When using an OIDC provider the default session length is 1 hour. Are we going to need to set it to be longer for this? 

[ref](https://github.com/aws-actions/configure-aws-credentials#assuming-a-role)

bryan-aguilar (approved):

PR #973, opened 2022-11-28
Update the usage of set-output command in GH actions

vasireddy99:
**Description:** 
This PR is created to update the usage of set-output command in GH workflows as per the [recommendation](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

**Testing:** 
Targeted change: https://github.com/vasireddy99/aws-otel-test-framework/blob/test-setOutput/tools/batchTestGenerator/internal/githubGenerator.go#L48-L72

Test Workflow: https://github.com/vasireddy99/aws-otel-test-framework/blob/test-setOutput/.github/workflows/test.yml#L33-L43

Test Result: https://github.com/vasireddy99/aws-otel-test-framework/actions/runs/3884938865/jobs/6628160676#step:5:1

**Documentation:**  N/A


.github/workflows/pr-build.yml
```diff
@@ -44,7 +44,7 @@ jobs:
     - name: Get npm cache directory
       id: npm-cache-dir
       run: |
-        echo "::set-output name=dir::$(npm config get cache)"
+        echo "name=dir::$(npm config get cache)" >> $GITHUB_OUTPUT
```
bryan-aguilar:
```suggestion
        echo "dir=$(npm config get cache)" >> $GITHUB_OUTPUT
```

tools/batchTestGenerator/internal/githubGenerator.go
```diff
@@ -42,9 +42,9 @@ func GithubGenerator(config RunConfig) error {
 		return fmt.Errorf("failed to marshal batch values object: %w", err)
 	}
 
-	fmt.Printf(`::set-output name=batch-keys::%s`, githubBatchKeys)
+	fmt.Printf(`"name=batch-keys::%s" >> $GITHUB_OUTPUT`, githubBatchKeys)
```
bryan-aguilar:
```suggestion
	fmt.Printf(`batch-keys=%s >> $GITHUB_OUTPUT`, githubBatchKeys)
```

tools/batchTestGenerator/internal/githubGenerator.go
```diff
@@ -42,9 +42,9 @@ func GithubGenerator(config RunConfig) error {
...
 	fmt.Printf("\n")
-	fmt.Printf(`::set-output name=batch-values::%s`, githubBatchValues)
+	fmt.Printf(`"name=batch-values::%s" >> $GITHUB_OUTPUT`, githubBatchValues)
```
bryan-aguilar:
```suggestion
	fmt.Printf(`batch-values=%s >> $GITHUB_OUTPUT`, githubBatchValues)
```

tools/batchTestGenerator/internal/validateCache.go
```diff
@@ -54,7 +54,7 @@ func ValidateCache(rc RunConfig, ddbTableName string, aocVersion string) error {
 		fmt.Println(strings.Join(cacheMisses, "\n"))
 		succcess = "false"
 	}
-	fmt.Printf(`::set-output name=release-candidate-ready::%s`, succcess)
+	fmt.Printf(`"name=release-candidate-ready::%s" >> $GITHUB_OUTPUT`, succcess)
```
bryan-aguilar:
```suggestion
	fmt.Printf(`release-candidate-ready=%s >> $GITHUB_OUTPUT`, succcess)
```

bryan-aguilar (commented):
Have you tested writing to the output file from a go binary on a test workflow? I have a suspicion that we may have to actually write to the env var instead of just printing to std out. I believe that in the previous method, GHA looked for `::set-output` on std out and would update the env var for us. The new method requires a user to write to the env var directly.

PaurushGarg (approved):
LGTM

tools/batchTestGenerator/internal/githubGenerator.go
```diff
@@ -42,10 +44,23 @@ func GithubGenerator(config RunConfig) error {
...
 	}
 
-	fmt.Printf(`::set-output name=batch-keys::%s`, githubBatchKeys)
-	fmt.Printf("\n")
-	fmt.Printf(`::set-output name=batch-values::%s`, githubBatchValues)
-	fmt.Printf("\n")
+	ghOutputFile := os.Getenv("GITHUB_OUTPUT")
+	ghEnv, err := os.OpenFile(ghOutputFile, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0600)
+	if err != nil {
+		log.Fatal(err)
```
bryan-aguilar: Should this error be returned instead? Also does calling `Fatal` here prevent the file from being closed correctly? I am not sure how the `defer` is handled in this case. My concern is that the flow goes something like `log -> os.exit(1)` and then the file is never correctly closed.

tools/batchTestGenerator/internal/githubGenerator.go
```diff
@@ -42,10 +44,23 @@ func GithubGenerator(config RunConfig) error {
...
+	}
+	defer ghEnv.Close()
+
+	//Writing into the gh env fie
+	_, err = ghEnv.WriteString(fmt.Sprintf(`batch-keys=%s`, githubBatchKeys))
+	if err != nil {
+		fmt.Printf("error writing githubBatchKeys in GITHUB_OUTPUT env: %v", err)
```
bryan-aguilar: This error should be returned from the function and not just printed.

tools/batchTestGenerator/internal/githubGenerator.go
```diff
@@ -42,10 +44,23 @@ func GithubGenerator(config RunConfig) error {
...
+	}
+
+	_, err = ghEnv.WriteString(fmt.Sprintf(`batch-values=%s`, githubBatchValues))
+	if err != nil {
+		fmt.Printf("error writing githubBatchValues in GITHUB_OUTPUT env: %v", err)
```
bryan-aguilar: This error should also be returned.

tools/batchTestGenerator/internal/githubGenerator.go
```diff
@@ -42,10 +44,23 @@ func GithubGenerator(config RunConfig) error {
...
-	fmt.Printf("\n")
+	ghOutputFile := os.Getenv("GITHUB_OUTPUT")
+	ghEnv, err := os.OpenFile(ghOutputFile, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0600)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer ghEnv.Close()
+
+	//Writing into the gh env fie
+	_, err = ghEnv.WriteString(fmt.Sprintf(`batch-keys=%s`, githubBatchKeys))
```
bryan-aguilar: How does the `ghEnv` file handle delimiters? Does this need a newline delimter at the end of each string?

tools/batchTestGenerator/internal/githubGenerator.go
```diff
@@ -42,10 +44,23 @@ func GithubGenerator(config RunConfig) error {
...
+	if err != nil {
+		fmt.Printf("error writing githubBatchKeys in GITHUB_OUTPUT env: %v", err)
+	}
+
+	_, err = ghEnv.WriteString(fmt.Sprintf(`batch-values=%s`, githubBatchValues))
```
bryan-aguilar: Same question here. Does this require a newline delimiter?

tools/batchTestGenerator/internal/validateCache.go
```diff
@@ -54,8 +56,19 @@ func ValidateCache(rc RunConfig, ddbTableName string, aocVersion string) error {
 		fmt.Println(strings.Join(cacheMisses, "\n"))
 		succcess = "false"
 	}
-	fmt.Printf(`::set-output name=release-candidate-ready::%s`, succcess)
-	fmt.Printf("\n")
+
+	ghOutputFile := os.Getenv("GITHUB_OUTPUT")
+	ghEnv, err := os.OpenFile(ghOutputFile, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0600)
+	if err != nil {
+		log.Fatal(err)
```
bryan-aguilar: Same questions as above for how this error should be handled.

tools/batchTestGenerator/internal/validateCache.go
```diff
@@ -54,8 +56,19 @@ func ValidateCache(rc RunConfig, ddbTableName string, aocVersion string) error {
...
+	}
+	defer ghEnv.Close()
+
+	//Writing into the gh env fie
+	_, err = ghEnv.WriteString(fmt.Sprintf(`release-candidate-ready=%s`, succcess))
+	if err != nil {
+		fmt.Printf("error writing githubBatchKeys in GITHUB_OUTPUT env: %v", err)
```
bryan-aguilar: This error should be returned

bryan-aguilar (commented):
The changes here seem reasonable but I still had a few questions. Can you link to a test workflow that you ran where the GitHub Env file was written to by a go program and then read from a subsequent step/job? It does not have to be the batchtest generator but could be a small go program demonstrating the required functionality.

PR #979, opened 2022-11-30
Adding code to build and push load-generator docker image

PaurushGarg:
- Changes docker file of load-gen to allow building of docker image while referencing local project without manual movements of files.
- Creates new GHA workflow to build project and build & push docker image to Amazon public ECR (at same location of current load-generator docker image) upon trigger of 'new push to `load-generator` in `terraform` branch'.
- Currently, unsure that used credentials are sufficient to build & push the image.

TODO:

- [ ] Followup PR required to migrate from current usage of docker image of load-generator to new image with latest tag.

#736 
#771

 Tested manual on a separate branch.


.github/workflows/update-load-generator-image.yml
```diff
@@ -0,0 +1,56 @@
...
+          registry-type: public
+      - name: Set up Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+      - name: Build with Gradle
+        run: ./gradlew :load-generator:build
+      - name: Build, tag, and push docker image to Amazon ECR Public
```
bryan-aguilar: In this case I suggest using github marketplace actions to ensure that all docker dependencies are setup correctly. The collector CI has an [example](https://github.com/aws-observability/aws-otel-collector/blob/6f6a27d672b5e266729cfb6da15ec8d106a2826d/.github/workflows/CI.yml#L510) of this.

.github/workflows/update-load-generator-image.yml
```diff
@@ -0,0 +1,56 @@
...
+        env:
+          REGISTRY: ${{ steps.login-ecr-public.outputs.registry }}
+          REGISTRY_ALIAS: aws-otel-test
+          REPOSITORY: aws-otel-load-generator
+          IMAGE_TAG: latest
```
bryan-aguilar: We should publish more than just the latest tag. What about also tagging the image with the git short sha? 
Also, what image tag are dependencies of this project referencing? If they are referencing `latest` does pushing a new image break them?

load-generator/Dockerfile
```diff
@@ -10,7 +10,7 @@ COPY ./build/ ./build/
 # default 1.0
 ARG VER=1.0
 
-RUN gradle build
+# assume app is already built since it will not build in docker container
```
bryan-aguilar: Why?

.github/workflows/update-load-generator-image.yml
```diff
@@ -0,0 +1,56 @@
...
+  workflow_dispatch:
+  push:
+    branches: ['terraform']
+
+    # List of paths that should be considered when triggering the workflow
+    paths: ['load-generator/**']
+
+permissions:
+  id-token: write
+  contents: write
```
bryan-aguilar:
```suggestion
  contents: read
```

I don't think this requires write permissions to the repository does it?

.github/workflows/update-load-generator-image.yml
```diff
@@ -0,0 +1,56 @@
...
+  create-and-push-load-generator-image-on-push:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
bryan-aguilar: AWS Credentials should be acquired using an IAM Role which utilizes a GitHub OIDC provider.

load-generator/build.gradle
```diff
@@ -51,9 +51,6 @@ dependencies {
     implementation group: 'org.apache.logging.log4j', name: 'log4j-api', version: '2.19.0'
     implementation group: 'org.apache.logging.log4j', name: 'log4j-core', version: '2.19.0'
 
-    // local project can't be referenced in Dockerfile
-    // the workaround is build fat jar and copy it in Dockerfile for building docker image
-    //  implementation files('build/libs/trace-java-client-1.0-all.jar')
```
Aneurysm9, line 56: What did this comment mean and why is it no longer applicable?

load-generator/Dockerfile
Aneurysm9: Yes, why?

Also, if not building the application in the Dockerfile then there's more extensive surgery to do.  This probably doesn't even need to be a multi-stage build if that's the case.

PaurushGarg: [trace-java-client](https://github.com/aws-observability/aws-otel-test-framework/tree/terraform/trace-java-client) is referenced as local project by two gradle projects: load-generator and [jaeger-zipkin-sample-app](jaeger-zipkin-sample-app). 
`Load-generator` build by `Gradle` works by referencing to `trace-java-client` local project. But to build `load-generator` inside `dockerfile` requires commenting out [line 57](https://github.com/aws-observability/aws-otel-test-framework/blob/06496c5e679991d839b628afd737dcf28ce35214/load-generator/build.gradle#L57) and uncommenting [line 54 to 56](https://github.com/aws-observability/aws-otel-test-framework/blob/06496c5e679991d839b628afd737dcf28ce35214/load-generator/build.gradle#L54). - This is the current setup. Dockerfile and Gradle project are not able to build sameway b/c of difference in referencing a local project.

So, I noticed that `jaeger-zipkin-sample-app` resolves this thing by [simply not building project inside dockerfile]( https://github.com/aws-observability/aws-otel-test-framework/blob/06496c5e679991d839b628afd737dcf28ce35214/sample-apps/jaeger-zipkin-sample-app/Dockerfile#L12) and rather requires that the project is built before the dockerfile is built so that Docker can simply copy the executable over.

Now with this, the first stage of copying the source code is indeed meaningless as we are expecting the binary to be available in the local filesystem before docker build command. I avoided to have this change in this PR, but if it's approved then I can edit the dockerfile of both these projects in this PR.

.github/workflows/update-load-generator-image.yml
PaurushGarg: Ok. Thanks. I will update this.

.github/workflows/update-load-generator-image.yml
PaurushGarg: Ok. Thanks. I will update this.

.github/workflows/update-load-generator-image.yml
PaurushGarg: From what I noticed, All the current sample-apps are publishing [image with latest tag only](https://github.com/aws-observability/aws-otel-test-framework/blob/06496c5e679991d839b628afd737dcf28ce35214/terraform/imagebuild/main.tf#L61).
A git short sha can be added as a tag through the GHA workflow, but I couldn't find a solve on how to get that tag to become part of this [dockerfile path](https://github.com/aws-observability/aws-otel-test-framework/blob/06496c5e679991d839b628afd737dcf28ce35214/terraform/ec2_setup/variables.tf#L24) automatically. It will become a manual process to ensure that the latest/recent image is used. 
I couldn't find a existing setup in test-framework to reference this as well.

bryan-aguilar: Here is a similar example in [collector CI](https://github.com/aws-observability/aws-otel-collector/blob/daf6a4942f318cc31df927bf7e5cc92ba4cfe2dd/.github/workflows/CI.yml#L452). 

Is it possible to run `git rev-parse --short HEAD` in a subsequent step, store it as an output, and then reference that output in the list of tags?

load-generator/Dockerfile
rapphil: Have you taken a look into the Java sample apps? They use jib to build and push the container images.

https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/sample-apps/spark/build.gradle.kts#L27


If possible we should avoid assuming things and just build everything.

There is an example here of how to use job in a multi project environment
https://github.com/GoogleContainerTools/jib/blob/master/examples/multi-module/name-service/build.gradle#L19

.github/workflows/update-load-generator-image.yml
PaurushGarg: Hi Bryan, my query is how to include this dynamic `short-sha` as part of static path of `load-generator` [docker image](https://github.com/aws-observability/aws-otel-test-framework/blob/06496c5e679991d839b628afd737dcf28ce35214/terraform/ec2_setup/variables.tf#L24)- in the code? This is required to pull this image that we are publishing.

bryan-aguilar: Could we split the image URL and tag? Then provide the tag value at test execution time?

load-generator/Dockerfile
PaurushGarg: Thanks @rapphil for this suggestion. I was not aware of this tool, but now I see it quite a lot in test-framework. I have updated the code to incorporate this. Since the `dockerfile` is not used anymore so I have removed it.

.github/workflows/update-load-generator-image.yml
PaurushGarg: Thanks I have updated this.

.github/workflows/update-load-generator-image.yml
PaurushGarg: Thanks @bryan-aguilar for guiding on this. I have used the role utilizing OIDC provider.

.github/workflows/update-load-generator-image.yml
PaurushGarg: Thanks. I have updated this. added multiple tags. In the follow up PR, I will pull this image using latest tag.

.github/workflows/update-load-generator-image.yml
PaurushGarg: Thanks. This section has now been modified to use google jib instead.

.github/workflows/update-load-generator-image.yml
```diff
@@ -0,0 +1,49 @@
...
+        id: login-ecr-public
+        uses: aws-actions/amazon-ecr-login@v1
+        with:
+          registry-type: public
+      - name: Set up Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Build image with jib
+        uses: burrunan/gradle-cache-action@v1
```
bryan-aguilar: I'm not familiar with this action. It also does not appear to be actively maintained. Have you considered using the marketplace action that the java-instrumentation uses? can be seen [here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/22af6a0e21b980aeacef711e2fe4db74ef5d1a6b/.github/workflows/main-build.yml#L49).

.github/workflows/update-load-generator-image.yml
```diff
@@ -0,0 +1,49 @@
+# This workflow uses GHA to build a Java project with Gradle 'load-generator', and build and push multi-arch docker image of this project to the Amazon ECR.
+
+name: Update Load-Generator Image
+
+# Should execute only one workflow run at a time
+concurrency:
+  group: build-and-push-load-generator-image
```
bryan-aguilar, line 7: Let's add the [cancel in progress](https://github.com/aws-observability/aws-otel-test-framework/blob/aec36308faa0178ee9a65b4bc4fd6bcbd15caea6/.github/workflows/cdk-deploy.yml#L15) field here also.

bryan-aguilar: This will ensure that we don't rely on the default behavior.

PaurushGarg: Updated. Done.

.github/workflows/update-load-generator-image.yml
PaurushGarg: Updated.

bryan-aguilar (approved):

rapphil (approved):

PR #999, opened 2022-12-13
Choose subnet based on random id. 

bryan-aguilar:
**Description:** Previously all ec2 instances created were going into a single subnet. This would cause the subnet to run out of IPs. This change tries to balance this selection across the three subnets to avoid IP resource limitations. 

Testing: Tested this locally by running multiple ec2 tests to ensure different subnets would be selected. 

### DIFF
--- a/terraform/basic_components/main.tf
+++ b/terraform/basic_components/main.tf
@@ -20,6 +20,16 @@ module "common" {
   security_group_name = var.aoc_vpc_security_group
 }
 
+locals {
+  otconfig_path = fileexists("${var.testcase}/otconfig.tpl") ? "${var.testcase}/otconfig.tpl" : module.common.default_otconfig_path
+
+  subnet_ids_list = tolist(data.aws_subnet_ids.aoc_public_subnet_ids.ids)
+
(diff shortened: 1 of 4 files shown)


bryan-aguilar:
This will also need to be PRed into the dev branch.

terraform/basic_components/main.tf
```diff
@@ -20,6 +20,16 @@ module "common" {
   security_group_name = var.aoc_vpc_security_group
 }
 
+locals {
+  otconfig_path = fileexists("${var.testcase}/otconfig.tpl") ? "${var.testcase}/otconfig.tpl" : module.common.default_otconfig_path
+
+  subnet_ids_list = tolist(data.aws_subnet_ids.aoc_public_subnet_ids.ids)
+
+  subnet_ids_random_index = random_id.subnetSelector.dec % length(data.aws_subnet_ids.aoc_public_subnet_ids.ids)
```
Aneurysm9:
```suggestion
  subnet_ids_random_index = random_id.subnetSelector.dec % length(subnet_ids_list)
```

Might as well use the list you just made.

Aneurysm9 (approved):

PR #1041, opened 2023-01-10
Add addons integ test clusters to `clusters.yml`

erichsueh3:
This PR modifies `clusters.yml` to add in new clusters to be used for ADOT EKS add-on integration testing.

### DIFF
--- a/cdk_infra/lib/config/cluster-config/clusters.yml
+++ b/cdk_infra/lib/config/cluster-config/clusters.yml
@@ -71,59 +71,8 @@ clusters:
     version: "1.24"
     launch_type: ec2
     instance_type: "m5.large"
-  - name: eks-addon-amd64-1-21
-    version: "1.21"
-    launch_type: ec2
-    instance_type: "m5.large"
-    cert_manager: true
-  - name: eks-addon-arm64-1-21
-    version: "1.21"
-    launch_type: ec2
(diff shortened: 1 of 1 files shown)


Aneurysm9:
Do these overlap with the existing "operator" clusters?  Can these be combined in any way, such as by testing both fargate and non-fargate deployments on the same cluster in appropriately configured namespaces?  Should this go into a new account?

erichsueh3:
> Do these overlap with the existing "operator" clusters? Can these be combined in any way, such as by testing both fargate and non-fargate deployments on the same cluster in appropriately configured namespaces? Should this go into a new account?

These will be setup similar to the existing operator clusters, but should be separate to avoid any potential issues with failures due to concurrently running tests. As for combining fargate/non-fargate deployments, I think they could be, though I think that the addition of 4 additional clusters is relatively inexpensive, and having them in separate clusters may be useful when trying to debug tests. As for the new account, the infrastructure already exists in the test account, and I think ADOT EKS add-on integration testing would fit into that category as well, but would be open to a different perspective on that

erichsueh3:
Update on this PR, we will add only one cluster initially to satisfy a POC for ADOT EKS add-on integration testing. Further research will be done on ways to minimize the number of clusters used. cc @Aneurysm9

Aneurysm9 (approved):

PR #1050, opened 2023-01-19
Use cluster name to choose AMI type for EKS MNG launch template

Aneurysm9:
Signed-off-by: Anthony J Mirabella <a9@aneurysm9.com>

**Description:** Because we're specifying a launch template for the EKS managed nodegroups we're creating we need to tell it which AMI to use and that AMI needs to match the architecture for the instance type selected.  Because all of our clusters include the architecture in the name I have chosen to extract that information from there rather than add a new field to the configuration.

### DIFF
--- a/cdk_infra/lib/cluster-deployment.ts
+++ b/cdk_infra/lib/cluster-deployment.ts
@@ -1,18 +1,17 @@
 #!/usr/bin/env node
 import 'source-map-support/register';
 import * as cdk from 'aws-cdk-lib';
-import { aws_eks as eks, StackProps } from 'aws-cdk-lib';
-import { readFileSync } from 'fs';
-import { EC2Stack } from './stacks/eks/ec2-cluster-stack';
-import { FargateStack } from './stacks/eks/fargate-cluster-stack';
-import { validateFileSchema } from './utils/eks/validate-config-schema';
-import { ClusterInterface } from './interfaces/eks/cluster-interface';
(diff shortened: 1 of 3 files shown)


humivo (approved):

cdk_infra/lib/cluster-deployment.ts
```diff
@@ -57,7 +56,8 @@ export function deployClusters(
         name: ec2Cluster.name,
         vpc: vpc,
         version: versionKubernetes,
-        instance_type: ec2Cluster.instance_type,
+        instanceTypes: [new InstanceType(ec2Cluster.instance_type.toLowerCase())],
+        amiType: ec2Cluster.name.match('\-arm64\-') ? NodegroupAmiType.AL2_ARM_64 : NodegroupAmiType.AL2_X86_64,
```
rapphil, line 60: nit: We could pass the AMI type from the test_cluster.yaml file as well, so that we don't need to rely on the name.

https://github.com/aws/aws-cdk/blob/v1.189.0/packages/@aws-cdk/aws-eks/lib/managed-nodegroup.ts#L24

rapphil (approved):

cdk_infra/lib/cluster-deployment.ts
Aneurysm9: We could, but we already had the convention of including the architecture in the cluster name and matching on that is likely to be less error-prone than correctly copying the magic AMI type string.  This also ensures that we only use one of these two AMI types.

PR #1067, opened 2023-01-31
[Hostmetrics Receiver]: e2e testing- CW validator, Linux expectation template, Testcase, and ECS script

PaurushGarg:
Added `Hostmetrics` Receiver e2e testing 

- CW validator
- Linux expectation template
- testcase
- and ECS script

https://github.com/aws-observability/aws-otel-ops/issues/137
PR is long but can be spit into sections. Kept it long so that whole PR can be tested by running as one testcase with below command.
Manually testing on local box by running:
```
cd terraform/ecs
terraform init
terraform apply -auto-approve \
    -var="aoc_image_repo=public.ecr.aws/q1w6y6d0/hostmetricsamd" \
    -var="aoc_version=latest" \
    -var="testcase=../testcases/hostmetrics_receiver" \
    -var-file="../testcases/hostmetrics_receiver/parameters.tfvars"
```

Note - Requires Collector with `hostmetrics_reciever` component

- [ ] Tracking errors that came up during execution of these tests: https://quip-amazon.com/nXJSAzstMBsW/Issues-discovered-during-hostmetrics-e2e-testing


terraform/ec2/main.tf
```diff
@@ -129,11 +128,6 @@ resource "aws_instance" "aoc" {
     TestID    = module.common.testing_id
     ephemeral = "true"
   }
-
-  metadata_options {
-    http_endpoint = "enabled"
-    http_tokens   = "required"
-  }
```
bryan-aguilar: Why was this removed? I believe this is required to disable IMDSv1

terraform/testcases/hostmetrics_receiver/parameters.tfvars
```diff
@@ -0,0 +1,7 @@
+# this file is defined in validator/src/main/resources/validations
+validation_config     = "hostmetrics-metric-linux-validation.yml" #incase windows ami_family is used in ec2 then this property will be overridden in EC2
```
bryan-aguilar, line 2: Where does the responsibility of choosing the correct `validation_config` lie? In what cases would a test case require different `validation_config` files? To me it seems like if a test case requires a different `validation_config` then it should be it's own test case. 

Can you walk through the use case or constraints where it made it necessary for the terraform ec2 module to ensure the correction `validation_config` value was chosen?

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
```diff
@@ -20,57 +20,130 @@
...
+    List<Metric> expectedMetricList = null;
     ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
-    List<Metric> expectedMetricList = mapper.readValue(
-          yamlExpectedMetrics.getBytes(StandardCharsets.UTF_8),
-          new TypeReference<List<Metric>>() {});
+
+    if (context != null
+        && context.getTestcase() != null
+        && context.getTestcase().contains("hostmetrics_receiver")) {
+      if (context.getHostmetricsContext() != null) {
```
bryan-aguilar, line 61: Do we need all these null checks or can we just check if `context.getHostmetricsContext()!=null`? Is there a reason where this would be true but we would not want use the host metrics mapper? Or would we at minimum check that the test case name contains `hostmetrics_receiver` and the context is not null?

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
```diff
@@ -20,57 +20,130 @@
...
-      ICaller caller
-  ) throws Exception {
+      Context context, FileConfig expectedMetric, ICaller caller) throws Exception {
     // call endpoint
     if (caller != null) {
       caller.callSampleApp();
     }
 
     // get expected metrics as yaml from config
-    String yamlExpectedMetrics = mustacheHelper.render(expectedMetric, context);
```
bryan-aguilar, line 59: Why does this change need to be made? It is repeated in both blocks of the `if/else` below.

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
```diff
@@ -20,57 +20,130 @@
...
+    }
 
     if (context.getIsRollup()) {
       return this.rollupMetric(expectedMetricList);
     }
-
     return expectedMetricList;
   }
 
+  private List<Metric> parseCWModel(
```
bryan-aguilar, line 83:
```suggestion
  private List<Metric> convertHostMetricModeltoMetricModel(
```
I believe a more descriptive name is helpful here.

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
```diff
@@ -20,57 +20,130 @@
...
+      ObjectMapper mapper, List<com.amazon.aoc.models.cwmodel.Metric> expectedMetricsForHostMetrics)
+      throws JsonProcessingException {
+    List<Metric> metrics = new ArrayList<>();
+    Stack<com.amazon.aoc.models.cwmodel.Metric> dpParentStack = new Stack<>();
+    Stack<com.amazon.aoc.models.cwmodel.Metric> dpChildStack = new Stack<>();
+
+    for (com.amazon.aoc.models.cwmodel.Metric hostMetricDataPoint : expectedMetricsForHostMetrics) {
+      dpParentStack.push(hostMetricDataPoint);
+      if (hostMetricDataPoint.getDimensions()
+          != null) {
```
bryan-aguilar, line 93:
```suggestion
      if (hostMetricDataPoint.getDimensions() != null) {
```

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
```diff
@@ -20,57 +20,130 @@
...
+        int dimensionsSize = hostMetricDataPoint.getDimensions().size();
+        if (dimensionsSize > 0) {
+          for (int i = 0; i < dimensionsSize; i++) { // iterate over each dimension
```
bryan-aguilar, line 96:
```suggestion
          for (int i = 0; i < hostMetricDataPoint.getDimensions().size(); i++) { // iterate over each dimension
```

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
```diff
@@ -20,57 +20,130 @@
...
-
     return expectedMetricList;
   }
 
+  private List<Metric> parseCWModel(
+      ObjectMapper mapper, List<com.amazon.aoc.models.cwmodel.Metric> expectedMetricsForHostMetrics)
+      throws JsonProcessingException {
+    List<Metric> metrics = new ArrayList<>();
+    Stack<com.amazon.aoc.models.cwmodel.Metric> dpParentStack = new Stack<>();
+    Stack<com.amazon.aoc.models.cwmodel.Metric> dpChildStack = new Stack<>();
```
bryan-aguilar, line 88: What does dp stand for here?

bryan-aguilar (commented):
I made it through the terrafom setup code but hit a time block when I started to review `parseCWModel` function. As a note, I am not a java expert and building a mental model of the translation algorithm is taking some time. I think others may be helpful to add some comments on why the conversion function exists and what it's purpose is. Also please ensure that descriptive variable names are used where appropriate. This makes it easy for reviewers to build context around what is happening.

terraform/ecs/main.tf
```diff
@@ -312,6 +329,29 @@ resource "aws_ecs_service" "aoc_without_sample_app_for_bridge" {
...
+  }
+  depends_on = [aws_ecs_task_definition.aoc]
+}
+
+# wait for task to get deployed and metrics have been received by cloudwatch
+resource "time_sleep" "wait_60_seconds" {
+  count      = endswith(var.testcase, "hostmetrics_receiver") ? 1 : 0
+  depends_on = [aws_ecs_service.aoc_without_sample_app_daemon_scheduling]
+
+  create_duration = "60s"
```
erichsueh3, line 353: Where did 60s come from, is this just an arbitrary number chosen? Just curious

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
```diff
@@ -20,57 +20,130 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
+import java.util.*;
```
erichsueh3, line 29: Any specific reason why we opt to use the wildcard import here? Small nit which I don't think matters in this context, but for example I believe this would cause a spotbugs warning when compiling. I think this applies to a few other flies' code imports as well

validator/src/main/java/com/amazon/aoc/helpers/CWMetricHelper.java
erichsueh3: Isn't this line for removing this change, and adding it to the `if/else` blocks? + the line under initializes `yamlExpectedMetrics` so that it will compile correctly as it won't compile if it is declared only in the `if/else` block

terraform/ec2/main.tf
PaurushGarg: Oh. removed by mistake. Thanks for highlighting. I have put it back.

bryan-aguilar:
I noticed commit `83c7fc1` was pushed that added a lot of changes related to EKS and EC2 test scripts. Can these be extracted into a separate PR that builds on top of this one? Adding this amount of changes on top of an existing PR isn't very helpful for reviewers. Creating PRs that build on top of each other can be good because we can review PRs one at a time. 

If this PR just focuses on adding validation + ecs tests that would make review easier.

terraform/testcases/hostmetrics_receiver/otconfig.tpl
```diff
@@ -0,0 +1,76 @@
...
+processors:
+  resourcedetection/ec2:
+    detectors: [ec2]
+    timeout: 2s
+    attributes: [host.id]
+    override: true
+
+exporters:
+  awsemf:
+    namespace: '${testing_id}'
```
rapphil, line 65: Is it possible to add a prefix `hostmetrics-test` or something like that so that it is possible to still explore the metrics in the cloudwatch console of the integration test account?
E.g.:
```
namespace: 'hostmetrics-test/${testing_id}'
```

terraform/ec2/main.tf
```diff
@@ -41,7 +41,7 @@ module "basic_components" {
 
   cortex_instance_endpoint = var.cortex_instance_endpoint
 
-  sample_app_listen_address_host = aws_instance.sidecar.public_ip
+  sample_app_listen_address_host = one(aws_instance.sidecar[*].public_ip)
```
bryan-aguilar, line 44: Is the usage of `one()` going to break our `Canary` tests since we are [pinned](https://github.com/aws-observability/aws-otel-collector/blob/e4ddd80fffefca1bb417ed07996a508931ec96de/.github/workflows/canary.yml#L112) to version 14? `one` appears to require `v15` or newer. [docs](https://developer.hashicorp.com/terraform/language/functions/one).

terraform/eks/aoc_standalone_daemonset.tf
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  data = {
+    "aoc-config.yml" = module.basic_components.0.otconfig_content
+  }
+}
+
+# deploy aoc and mocked server
+resource "kubernetes_daemonset" "aoc_deployment" {
+  count = var.aoc_base_scenario == "aoc_standalone_daemonset" && replace(var.testcase, "_adot_operator", "") == var.testcase && var.deployment_type != "fargate" && var.aoc_deploy_mode == "daemonset" ? 1 : 0
```
bryan-aguilar, line 45: Will having a daemonset deployment for all EKS test cases to be run synchronously? Can we run multiple test cases in parallel with this config?

PR #1077, opened 2023-02-07
Use cdk from node modules

rapphil:
**Description:** we should use the CDK from the node modules instead of using the global one. This will guarantee reproducibility between development environment and production environment because it will use the version defined in the `package-lock.json`.

### DIFF
--- a/cdk_infra/Makefile
+++ b/cdk_infra/Makefile
@@ -1,3 +1,3 @@
 .PHONY: deploy-infra
 deploy-infra: 
-	cdk deploy --all --require-approval never --concurrency 4
+	npm run cdk deploy -- --all --require-approval never --concurrency 4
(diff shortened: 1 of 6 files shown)


cdk_infra/Makefile
```diff
@@ -1,3 +1,3 @@
 .PHONY: deploy-infra
 deploy-infra: 
-	cdk deploy --all --require-approval never --concurrency 4
```
rapphil, line 3: `--` is to pass parameters to the node script that is being called.

cdk_infra/Makefile
```diff
@@ -1,3 +1,3 @@
...
+	npm run cdk deploy -- --all --require-approval never --concurrency 4
```
rapphil, line 3: `--` is to pass parameters to the node script that is being called.

bryan-aguilar (approved):

PR #1078, opened 2023-02-07
Create MSK Infrastructure

rapphil:
Signed-off-by: Raphael Silva <rapphil@gmail.com>

**Description:** Add permanent MSK clusters.

**Testing:** Deployed the infra locally and run integration tests using it.

**Documentation:** Updated documentation

### DIFF
--- a/cdk_infra/lib/stacks/msk/msk-stack.ts
+++ b/cdk_infra/lib/stacks/msk/msk-stack.ts
@@ -32,9 +32,12 @@ export class MSKClustersStack extends Stack {
 
     // Get the legacy VPC that was created with terraform
     // This lookup has to be done inside a stack
-    const aocVPC = props.aocVPC == null ? aws_ec2.Vpc.fromLookup(this, 'aoc-vpc', {
-      vpcName: 'aoc-vpc'
-    }) : props.aocVPC;
+    const aocVPC =
+      props.aocVPC == null
+        ? aws_ec2.Vpc.fromLookup(this, 'aoc-vpc', {
+            vpcName: 'aoc-vpc'
+          })
+        : props.aocVPC;
 
     // Legacy VPC clusters
     MSK_VERSIONS.forEach(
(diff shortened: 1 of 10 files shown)


cdk_infra/package-lock.json
```diff
@@ -36,8 +36,8 @@
         "jest": "^27.5.1",
         "prettier": "^2.8.1",
         "ts-jest": "^27.1.4",
-        "ts-node": "^10.8.1",
-        "typescript": "~3.9.7",
+        "ts-node": "^10.9.1",
+        "typescript": "~4.9.5",
```
rapphil, line 40: This had to be updated in order to get access to `string.replaceAll`

cdk_infra/lib/stacks/msk/msk-stack.ts
```diff
@@ -0,0 +1,141 @@
...
+          }
+        }
+      }
+    });
+  }
+}
+
+export interface MSKStackProps extends StackProps {
+  eksVPC: IVpc; // VPC used in the EKS Clusters
+  aocVPC: IVpc; // Legacy VPC used in the remaining in the tests
```
bryan-aguilar: Can we instead accept a set of VPCs as input and build clusters based on that?

bryan-aguilar (commented):
Overall not many nits. The stack that is being added represents the set of clusters required to execute kafka based tests in the account. Could we add some comments about the naming requirements of the clusters?

cdk_infra/lib/stacks/msk/msk-stack.ts
rapphil: Instead of making the stack generic, I made the setup for generating a msk cluster for tests generic: i create the `MSKTestCluster` construct. 

The reason why we cannot make the stack generic is because we need to lookup for the aoc VPC that was created in terraform. the lookup can only happen inside a stack, therefore it is not possible to simple pass a set of vpcs to the msk stack. One alternative would be to create a stack just to perform the lookup, but I think that would add too much complexity for little benefit.

bryan-aguilar (approved):

PR #1079, opened 2023-02-08
Add kafka integration tests

rapphil:
**Description:** Add kafka integration tests.

**Testing:** Locally using terraform.


terraform/data_aoc_msk/vars.tf
```diff
@@ -0,0 +1,20 @@
...
+// Used to make sure that we don't have topic collisions. We cannot generate infinite amount of topics
+// but we also don't want tests to collide.
+variable "dedup_topic" {
+    default = ""
+}
+
+// Testcase name as passed to each main module (ec2, ecs or eks). E.g.: -var="testcase=../testcases/statsd"
+variable "testcase" {
+    default = ""
+}
```
bryan-aguilar, line 21: Is this not available already in `common`?

bryan-aguilar: I think I see now, we have to pass it into the `data_aoc_msk` module because this module does not import `common`.

terraform/ec2/main.tf
```diff
@@ -21,6 +21,14 @@ module "common" {
   security_group_name = var.aoc_vpc_security_group
 }
 
+module "aoc_msk_cluster" {
+  source = "../data_aoc_msk"
+
+ cluster_version = var.kafka_version
+ testcase = var.testcase
+ dedup_topic = "ec2${var.testing_type}${var.testing_ami}"
```
bryan-aguilar, line 29: Should we also be using the testing id as part of the dedup topic?

terraform/ecs/main.tf
```diff
@@ -20,6 +20,15 @@ module "common" {
...
 }
 
+
+module "aoc_msk_cluster" {
+  source = "../data_aoc_msk"
+
+ cluster_version = var.kafka_version
+ testcase = var.testcase
+ dedup_topic = "ecs"
+}
```
bryan-aguilar, line 30: This module is imported into every platform module.

Does it belong in `basic_components`? 
Will this module initialize for all test cases? Even non kafka ones? Should we be adding something like
```suggestion
module "aoc_msk_cluster" {
  # this syntax is wrong, don't copy it. It is merely a suggestion to get my point across. 
  count = string.startsWith(kafka, testcasename)
  source = "../data_aoc_msk"

 cluster_version = var.kafka_version
 testcase = var.testcase
 dedup_topic = "ecs"
}
```

bryan-aguilar (commented):
What's the data flow here for these test cases
otlp data gen -> collector -> export with kafka -> MSK Cluster -> Back to collector -> Kafka receiver -> export with (xray/emf) -> validate? 
If so, can we document that somewhere to avoid having to build a mental model each time someone comes across these tests? Maybe even a diagram.

terraform/data_aoc_msk/vars.tf
rapphil: Correct. my idea was to make this module totally self contained. It is basically used to enforce the naming conventions for looking up for clusters.

terraform/ec2/main.tf
rapphil: we cannot use testid because there is a finite number of topics supported in kafka, which are created on first use, but are not deleted automatically, therefore we have to reuse topics across tests. 
I still need to figure out how to make this work with the different branches.

terraform/ecs/main.tf
rapphil: This is is already done inside the `aoc_msk_cluster` module. If the cluster_version variable is empty, it will not fetch any data and it will return an empty map.

terraform/testcases/kafka_otlp_trace_3_2_0/README.md
```diff
@@ -0,0 +1 @@
+For more details of how this test works, please look into docs/kafka-tests.md
```
bryan-aguilar, line 1: Will this doc be included in this PR or a subsequent one?

bryan-aguilar (approved):

terraform/testcases/kafka_otlp_trace_3_2_0/README.md
bryan-aguilar: I'm good with merging to dev without the doc existing. But it should be a blocker for merging to `terraform`.

rapphil: No, it was my miss not to commit this. I added.

PR #1105, opened 2023-03-01
Update expected data files for sample app validation

humivo:
**Description:** Update the expected data files for the trace/metric tests for the new sample apps

**Testing:** Locally tested and it passed with the JS sample app as well as the python sample app

**Documentation:** Spec will also be updated with these changes


vasireddy99 (approved):

bryan-aguilar:
Aren't these metric names supposed to match the metrics names defined in the [spec](https://github.com/aws-observability/aws-otel-community/blob/master/sample-apps/SampleAppSpec.md)?

PR #1109, opened 2023-03-02
[batchTestGenerator] Add ability to exclude tests from certain clusters

bryan-aguilar:
**Description:** Adds the ability to exclude tests cases from executing on specific clusters. 

### DIFF
--- a/tools/batchTestGenerator/internal/config.go
+++ b/tools/batchTestGenerator/internal/config.go
@@ -56,8 +56,18 @@ type ClusterTarget struct {
 }
 
 type Target struct {
-	Name   string `json:"name"`
-	Region string `json:"region"`
+	Name          string   `json:"name"`
+	Region        string   `json:"region"`
+	ExcludedTests []string `json:"excluded_tests"`
+}
+
+func (t *Target) isTestCaseExcluded(value string) bool {
+	for _, x := range t.ExcludedTests {
+		if x == value {
+			return true
+		}
+	}
+	return false
 }
 
 func (r *RunConfig) unmarshalInputFile() error {
@@ -109,6 +119,7 @@ func (r *RunConfig) validateTestCaseInput() error {
 			}
 		}
 	}
+
 	return nil
 }
 
--- a/tools/batchTestGenerator/internal/config.go
+++ b/tools/batchTestGenerator/internal/config.go
@@ -72,14 +72,6 @@ func (t *Target) isTestCaseExcluded(value string) bool {
 
 	_, contains := t.excludedTestMap[value]
 	return contains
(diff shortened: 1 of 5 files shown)


tools/batchTestGenerator/internal/config.go
```diff
@@ -56,8 +56,22 @@ type ClusterTarget struct {
 }
 
 type Target struct {
-	Name   string `json:"name"`
-	Region string `json:"region"`
+	Name            string   `json:"name"`
+	Region          string   `json:"region"`
+	ExcludedTests   []string `json:"excluded_tests"`
+	excludedTestMap map[string]struct{}
```
Aneurysm9:
```suggestion
	excludedTestSet map[string]struct{}
```

You're using it as a set, should probably name it that way.

tools/batchTestGenerator/internal/config.go
```diff
@@ -56,8 +56,22 @@ type ClusterTarget struct {
...
+}
+
+func (t *Target) isTestCaseExcluded(value string) bool {
+	if t.excludedTestMap == nil {
+		t.excludedTestMap = make(map[string]struct{})
+		for _, val := range t.ExcludedTests {
+			t.excludedTestMap[val] = struct{}{}
+		}
+	}
```
Aneurysm9: This could probably be done in an `UnmarshalJSON()` method.

tools/batchTestGenerator/internal/config.go
```diff
@@ -56,8 +56,22 @@ type ClusterTarget struct {
...
-	Name   string `json:"name"`
-	Region string `json:"region"`
+	Name            string   `json:"name"`
+	Region          string   `json:"region"`
+	ExcludedTests   []string `json:"excluded_tests"`
+	excludedTestMap map[string]struct{}
+}
+
+func (t *Target) isTestCaseExcluded(value string) bool {
+	if t.excludedTestMap == nil {
```
rapphil, line 66: if `excludedTestMap == nil` why not return `false` right away instead of populating this map with empty data?

rapphil: I see why now after reading @Aneurysm9 's comment.

tools/batchTestGenerator/internal/config.go
bryan-aguilar: Done!

tools/batchTestGenerator/internal/config.go
bryan-aguilar: Good suggestion, done!

Aneurysm9 (approved):

PR #1115, opened 2023-03-06
Add configurable sort key for test case batching

bryan-aguilar:
**Description:** This PR adds a new enviornment variable option for executing batch tests. This is necessary because the cache will need to be used for more than just Collector tests. Previously the sort key of the DDB Table was `<Collector Version><Testing Framework Version>`. This needs to change because the collector version will not always be the unit under test. This makes the cache usable for testing things like the OpenTelemetry SDKs where the SDK Version could be in the sort key value while the primary key remains the unique test combination. 

The change was made this way so that it is backwards compatible with existing implementations. The collector CI will not need to be updated since it does not set the `DDB_BATCH_CACHE_SK` value. 

Testing was done with the custom script below
```
#!/bin/bash
if [[ -z "${DDB_BATCH_CACHE_SK}" ]]; then
    DDB_SK_PREFIX=$TF_VAR_aoc_version
else
    DDB_SK_PREFIX=$DDB_BATCH_CACHE_SK
fi
test_framework_shortsha=$(git rev-parse --short HEAD)
echo $DDB_SK_PREFIX$test_framework_shortsha
```
Results are below
```
bryaag@147ddac12c15 terraform % export TF_VAR_aoc_version=v0.99.0
bryaag@147ddac12c15 terraform % ./sktest.sh 
v0.99.0b7992972
bryaag@147ddac12c15 terraform % export DDB_BATCH_CACHE_SK=v0.88.0
bryaag@147ddac12c15 terraform % ./sktest.sh                      
v0.88.0b7992972
```

### DIFF
--- a/terraform/checkCacheHit.sh
+++ b/terraform/checkCacheHit.sh
@@ -5,10 +5,21 @@
 # required env vars
 # DDB_TABLE_NAME: dyanmodb table name that will be queries
 # TF_VAR_aoc_version: test image version that was used
+# DDB_BATCH_CACHE_SK(OPTIONAL): If set then the prefix
+# of the sortkey will be set to this value. The default value is the 
+# value of TF_VAR_aoc_version. This is useful if testing
+# something other than the ADOT Collector and you would like to use a different
+# sort key. 
 
 test_framework_shortsha=$(git rev-parse --short HEAD)
 
-CACHE_HIT=$(aws dynamodb get-item --region=us-west-2 --table-name ${DDB_TABLE_NAME} --key {\"TestId\":{\"S\":\"$1$2$3\"}\,\"aoc_version\":{\"S\":\"${TF_VAR_aoc_version}$test_framework_shortsha\"}})
+if [[ -z "${DDB_BATCH_CACHE_SK}" ]]; then
+    DDB_SK_PREFIX=$TF_VAR_aoc_version
+else
+    DDB_SK_PREFIX=$DDB_BATCH_CACHE_SK
+fi
+
+CACHE_HIT=$(aws dynamodb get-item --region=us-west-2 --table-name ${DDB_TABLE_NAME} --key {\"TestId\":{\"S\":\"$1$2$3\"}\,\"aoc_version\":{\"S\":\"$DDB_SK_PREFIX$test_framework_shortsha\"}})
 
 if [ -z "${CACHE_HIT}" ]; then
     echo "Cache miss for $@"
--- a/terraform/checkCacheHit.sh
+++ b/terraform/checkCacheHit.sh
@@ -12,6 +12,9 @@
 # sort key. 
 
 test_framework_shortsha=$(git rev-parse --short HEAD)
+SERVICE="$1"
+TESTCASE=$2
+ADDTL_PARAMS=$3
 
 if [[ -z "${DDB_BATCH_CACHE_SK}" ]]; then
     DDB_SK_PREFIX=$TF_VAR_aoc_version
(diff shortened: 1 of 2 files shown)


terraform/checkCacheHit.sh
```diff
@@ -5,10 +5,21 @@
...
 test_framework_shortsha=$(git rev-parse --short HEAD)
 
-CACHE_HIT=$(aws dynamodb get-item --region=us-west-2 --table-name ${DDB_TABLE_NAME} --key {\"TestId\":{\"S\":\"$1$2$3\"}\,\"aoc_version\":{\"S\":\"${TF_VAR_aoc_version}$test_framework_shortsha\"}})
+if [[ -z "${DDB_BATCH_CACHE_SK}" ]]; then
+    DDB_SK_PREFIX=$TF_VAR_aoc_version
+else
+    DDB_SK_PREFIX=$DDB_BATCH_CACHE_SK
+fi
+
+CACHE_HIT=$(aws dynamodb get-item --region=us-west-2 --table-name ${DDB_TABLE_NAME} --key {\"TestId\":{\"S\":\"$1$2$3\"}\,\"aoc_version\":{\"S\":\"$DDB_SK_PREFIX$test_framework_shortsha\"}})
```
rapphil: since we are touching this, can we assign names to each of the parameter variables `$1`, `$2` and `$3`?

Dynamodb is schemaless, so can we change the `aoc_version` key to something else that is more descriptive?

Also, Is there any side effect besides the cache misses if use add a prefix for everything, including the collector?

This pr is a cache miss anyways, so it is ok to use this opportunity to change everything.

terraform/executeTerraformTest.sh
```diff
@@ -63,7 +74,7 @@ esac
 
 test_framework_shortsha=$(git rev-parse --short HEAD)
 checkCache(){
-    CACHE_HIT=$(aws dynamodb get-item --region=us-west-2 --table-name ${DDB_TABLE_NAME} --key {\"TestId\":{\"S\":\"$1$2$3\"}\,\"aoc_version\":{\"S\":\"${TF_VAR_aoc_version}$test_framework_shortsha\"}})
```
rapphil, line 66: can't we use the other script `checkCacheHit` here instead of duplicating the logic?

bryan-aguilar: Done

terraform/checkCacheHit.sh
bryan-aguilar: Switched to using named variables instead of $1$2$3. As we have discussed changing the `aoc_version` field name is not viable at this time.

terraform/checkCacheHit.sh
```diff
@@ -5,10 +5,24 @@
...
+# DDB_BATCH_CACHE_SK(OPTIONAL): If set then the prefix
+# of the sortkey will be set to this value. The default value is the 
+# value of TF_VAR_aoc_version. This is useful if testing
+# something other than the ADOT Collector and you would like to use a different
+# sort key. 
 
 test_framework_shortsha=$(git rev-parse --short HEAD)
+SERVICE="$1"
+TESTCASE=$2
+ADDTL_PARAMS=$3
```
rapphil, line 17: ❤️

terraform/executeTerraformTest.sh
```diff
@@ -62,37 +74,32 @@ case ${AWS_REGION} in
...
+    case "$SERVICE" in
         EKS*) terraform destroy --auto-approve $opts;
         ;;
     *)
         terraform destroy --auto-approve;
     ;;
     esac
 
-    checkCache $1 $2 $3 
+    ../checkCacheHit.sh $SERVICE $TESTCASE $ADDTL_PARAMS
```
rapphil: I think we can remove this line as you updated the condition to exit the loop so that it does not rely on CACHE_HIT:
```while [ $ATTEMPTS_LEFT -gt 0 ] && ! ../checkCacheHit.sh $SERVICE $TESTCASE $ADDTL_PARAMS; do```

bryan-aguilar: Good point! Nice catch.

rapphil (approved):

PR #1117, opened 2023-03-08
Add resources used by config map providers

rapphil:
**Description:** Add resources used by the config map providers in the integration tests. For now we just have a bucket that will be used to store configurations used during the tests by the following config map providers:
* s3
* http
* https

**Testing:** Tested in my own account

### DIFF
--- a/cdk_infra/lib/app.ts
+++ b/cdk_infra/lib/app.ts
@@ -21,6 +21,6 @@ new MSKClustersStack(app, 'msk-clusters', {
   eksVPC: vpcStack.vpc
 });
 new ConfigMapProvidersStack(app, 'config-map-providers', {
-  suffix: process.env.CONFIG_BUCKET_SUFFIX || ''
+  suffix: process.env.CONFIG_BUCKET_SUFFIX
 });
 deployClusters(app, vpcStack.vpc, envDefault);
--- a/cdk_infra/lib/app.ts
+++ b/cdk_infra/lib/app.ts
@@ -21,6 +21,6 @@ new MSKClustersStack(app, 'msk-clusters', {
   eksVPC: vpcStack.vpc
 });
 new ConfigMapProvidersStack(app, 'config-map-providers', {
-  suffix: process.env.CONFIG_BUCKET_SUFFIX
+  env: envDefault
 });
 deployClusters(app, vpcStack.vpc, envDefault);
--- a/cdk_infra/lib/app.ts
+++ b/cdk_infra/lib/app.ts
@@ -4,6 +4,7 @@ import * as cdk from 'aws-cdk-lib';
 import { deployClusters } from './cluster-deployment';
 import { VPCStack } from './stacks/vpc/vpc-stack';
 import { MSKClustersStack } from './stacks/msk/msk-stack';
+import { ConfigMapProvidersStack } from './stacks/config-map-providers/config-map-providers-stack';
 
 const envDefault = {
   account: process.env.CDK_DEFAULT_ACCOUNT,
@@ -18,5 +19,8 @@ const vpcStack = new VPCStack(app, 'EKSVpc', {
 new MSKClustersStack(app, 'msk-clusters', {
   env: envDefault,
   eksVPC: vpcStack.vpc
+});
+new ConfigMapProvidersStack(app, 'config-map-providers', {
+  suffix: process.env.CONFIG_BUCKET_SUFFIX || ''
 });
 deployClusters(app, vpcStack.vpc, envDefault);
--- a/cdk_infra/lib/constructs/msk/test-cluster.ts
+++ b/cdk_infra/lib/constructs/msk/test-cluster.ts
@@ -4,7 +4,7 @@ import { aws_msk, RemovalPolicy } from 'aws-cdk-lib';
 import { LogGroup, RetentionDays } from 'aws-cdk-lib/aws-logs';
 
 export interface MSKTestClusterProps {
-  readonly namePrefix: String;
+  readonly namePrefix: string;
   readonly kafkaVersion: string;
   readonly vpc: IVpc;
   // Kafka server.properties file https://kafka.apache.org/documentation/#configuration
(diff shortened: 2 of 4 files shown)


cdk_infra/lib/constructs/msk/test-cluster.ts
```diff
@@ -4,7 +4,7 @@ import { aws_msk, RemovalPolicy } from 'aws-cdk-lib';
 import { LogGroup, RetentionDays } from 'aws-cdk-lib/aws-logs';
 
 export interface MSKTestClusterProps {
-  readonly namePrefix: String;
+  readonly namePrefix: string;
```
rapphil, line 7: unrelated to the pr description, but very small fix worth slipping into this one.

cdk_infra/lib/app.ts
```diff
@@ -19,4 +20,7 @@ new MSKClustersStack(app, 'msk-clusters', {
   env: envDefault,
   eksVPC: vpcStack.vpc
 });
+new ConfigMapProvidersStack(app, 'config-map-providers', {
+  suffix: process.env.CONFIG_BUCKET_SUFFIX
```
bryan-aguilar: Relying on an environment variable does not allow for deterministic deployments. Is it possible that we can hard the required suffix in the stack? Or is there a behavior here that we desire for non deterministic naming like this?

cdk_infra/lib/stacks/config-map-providers/config-map-providers-stack.ts
```diff
@@ -0,0 +1,47 @@
...
+export class ConfigMapProvidersStack extends Stack {
+  constructor(
+    scope: Construct,
+    id: string,
+    props: ConfigMapProvidersStackProps
+  ) {
+    super(scope, id, props);
+
+    const suffix = props.suffix || '';
+    const bucketName = `aws-otel-collector-integ-test-configurations${suffix}`;
```
bryan-aguilar:
```suggestion
    const bucketName = `adot-collector-integ-test-configurations${suffix}`;
```

Can we move away from the `aws-otel` nomenclature for all new infrastructure.

cdk_infra/lib/app.ts
bryan-aguilar: Is the purpose of the suffix to allow deployment of this stack in a dev account? If so can we update the naming or add documentation relating to this purpose?

bryan-aguilar: Adventures in going through all the files. I now see the comment in the `stack` file. 

nit: I  still am not a fan of relying on environment variables even for dev and testing purposes. I would prefer if instead of relying on env variables it was hard coded in the app. I would not hold the PR up for it.

rapphil: It is in the comments:
>   // Suffix used to test the stack and avoid collisions with s3 buckets names, that must be globally unique.

rapphil: We cannot get rid of the environment variable because of local tests. I hid the environment variable inside the stack instead of leaking into the app.

rapphil: also, I'm happy to hear alternatives for the case of local tests in a dev account.

cdk_infra/lib/app.ts
```diff
@@ -19,4 +20,7 @@ new MSKClustersStack(app, 'msk-clusters', {
   env: envDefault,
   eksVPC: vpcStack.vpc
 });
+new ConfigMapProvidersStack(app, 'config-map-providers', {
+  env: envDefault
```
bryan-aguilar, line 24: Should we be specifying a region here to deploy the cfn template to? Would an error arise if someone tried to deploy this CFN template in multiple regions?

cdk_infra/lib/app.ts
bryan-aguilar: In scenarios like this, one thing I have done  is to not rely on a bucket name without a UUID. Even in the default, non dev account, scenario. Have we considered attaching the account number as a UUID to the bucket?

cdk_infra/lib/app.ts
rapphil: your question is applicable to all the other resources as well. 

We are relying on region definition from here: https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/.github/workflows/cdk-deploy.yml#L26

bryan-aguilar: I don't think this is true. EKS and MSK are not globally unique in the way that s3 bucket names are.

bryan-aguilar: Perhaps there are some resources inside those stacks that are globally unique, IAM Roles(?), and that would be a fair critique.

cdk_infra/lib/app.ts
rapphil: I guess it makes sense to use the account id as a dedup mechanism since it is is only going to be used for the tests.

the downside is that we will need to fetch the account id in the tests and it might create confusion in case the account used for holding the configurations is not the same as the accounts used to run the tests. I guess this is ok.

cdk_infra/lib/app.ts
rapphil: I see, you are referring to the globally unique names. I think we can add the region name in the bucket name as well.

cdk_infra/lib/app.ts
bryan-aguilar: You can derive the account ID from inside the stack. Could [this](https://github.com/aws/aws-cdk/issues/1754#issuecomment-513542145) possibly be a good solution?

bryan-aguilar: Also, I'm not sure how this "could" break unit tests or how unit tests handle account env values. Can you make an account id in the unit tests?

rapphil: The tests that I meant are the integration tests that run in terraform.

bryan-aguilar: I see

bryan-aguilar (approved):

PR #1118, opened 2023-03-10
Add containerInsight test case for containerd runtime

vasireddy99:
**Description:** 

This PR adds the containerinsight testcase for containerd runtime and will be run only on >eks1.23 clusters. These changes include the slight modifications for containerinsightlogstructuredvalidator to accomodate changes required for this added testcase.

**Testing:** 

Testing is done locally


prometheus2-2.42.0-1.el7.x86_64.rpm
bryan-aguilar, line 1: What is this and why was it committed?

validator/src/main/java/com/amazon/aoc/validators/ContainerInsightStructuredLogValidator.java
```diff
@@ -34,18 +34,39 @@
...
   private static final int MAX_RETRY_COUNT = 15;
   private static final int QUERY_LIMIT = 100;
+  private static List<String> LOG_TYPE_TO_VALIDATE;
 
   @Override
   void init(Context context, FileConfig expectedDataTemplate) throws Exception {
     logGroupName = String.format("/aws/containerinsights/%s/performance",
             context.getCloudWatchContext().getClusterName());
     MustacheHelper mustacheHelper = new MustacheHelper();
+    if (context.getTestcase().contains("containerinsight_eks")) {
```
bryan-aguilar: This will end up being true for all `containerinsight_eks` test won't it? `containerinsight_eks` and `containerinsight_eks_containerd` will both evaluate to true.

bryan-aguilar (changes requested):
Please check EOF new lines. Lets also start by running this on dev branch. We can add this test case and parameters there. You may need to first update dev branch with changes in terraform.

validator/src/main/java/com/amazon/aoc/validators/ContainerInsightStructuredLogValidator.java
rapphil: good catch.
I think we can refactor this a bit so that we don't need to have this branching behavior.

* Lets create a base class `ContainerInsightStructuredLogValidatorBase` that contains everything currently here, except the LOG_TYPE_TO_VALIDATE. This class should have an abstract method `getLogTypeToValidate`, that will return the `List<String>` containing that information.
* Create a class ContainerInsightStructuredLogValidatorDocker that implements:
```
@override
public List<String> getLogTypeToValidate() {
return Arrays.asList(
          "Cluster",
          "ClusterNamespace",
          "ClusterService",
          "Container",
          "ContainerFS",
          "Node",
          "NodeDiskIO",
          "NodeFS",
          "NodeNet",
          "Pod",
          "PodNet"
  );
}
```
* Create a class ContainerInsightStructuredLogValidatorContainerd that implements:
```
@override
public List<String> getLogTypeToValidate() {
return Arrays.asList(
          "Cluster",
          "ClusterNamespace",
          "ClusterService",
          "Container",
          "Node",
          "NodeDiskIO",
          "NodeFS",
          "NodeNet",
          "Pod",
          "PodNet"
  );
}
```
* Update the ValidatorFactory accordingly creating distinct validation types (`containerinsight-eks-docker-logs` and `containerinsight-eks-containerd-logs`.
* Update testcases accordingly to use the new validator.

vasireddy99: I have resolved it with modified case name.

bryan-aguilar: I've sat on this idea a bit more. I don't think the pattern of using the test case name to make decisions is a good idea. I know there are examples elsewhere, in terraform and possibly the validator, but I don't think we should add more functionality that depends on it. It is not apparent when troubleshooting and when not documented properly it can cause a lot of headache. I would prefer that a specific variable gets passed in when necessary to the validator. One that is explicitly named so that we know that we are making a decision based on it.

validator/src/main/java/com/amazon/aoc/validators/ContainerInsightStructuredLogValidatorContainerd.java
```diff
@@ -0,0 +1,94 @@
...
+    return jsonNode.get("Type").asText();
+  }
+
+  @Override
+  protected int getMaxRetryCount() {
+    return MAX_RETRY_COUNT;
+  }
+
+  @Override
+  protected void fetchAndValidateLogs(Instant startTime) throws Exception {
```
bryan-aguilar: Is this duplicate code from  `ContainerInsightStructuredLogValidatorDocker`? Can we store this in an abstract class that `ContainerInsightStructuredLogValidatorDocker` and `ContainerInsightStructuredLogValidatorContainerd` extend?

bryan-aguilar: I think is similar to what @rapphil has requested here https://github.com/aws-observability/aws-otel-test-framework/pull/1118#discussion_r1139093992. We should have a base class for this.

bryan-aguilar (approved):

validator/src/main/resources/validations/eks-container-insight.yml
rapphil, line 1: do we need to change the name of this file as well?

rapphil: `eks-container-insight.yml` -> `eks-container-insight-docker.yml`

rapphil (approved):

PR #1126, opened 2023-03-16
Switch to aws_ecs_cluster resource from ecs-cluster module and require IMDSv2

humivo:
**Description:** The switch to building a aws_ecs_cluster resource replicates what we previous had from the ecs-cluster module except we are able to configure the ec2 launch template to require IMDSv2

**Testing:** Local testing passed for ecs tests in test framework launched through EC2 and FARGATE

### DIFF
--- a/terraform/ecs/main.tf
+++ b/terraform/ecs/main.tf
@@ -106,7 +106,6 @@ resource "aws_launch_template" "launchtemp" {
 
 resource "aws_autoscaling_group" "clusterasg" {
   name = "clusterasg"
-  #availability_zones = ["us-west-2a", "us-west-2b", "us-west-2c"]
   vpc_zone_identifier = module.basic_components.aoc_private_subnet_ids
   desired_capacity   = 1
   max_size           = 10
--- a/terraform/ecs/main.tf
+++ b/terraform/ecs/main.tf
@@ -151,7 +151,7 @@ data "aws_instances" "ecs-instances" {
     values = ["Integ-test-ecs-instance"]
   }
 
-  instance_state_names = ["running", "stopped"]
+  instance_state_names = ["running"]
 
   depends_on = [aws_autoscaling_group.clusterasg]
 }
--- a/terraform/ecs/main.tf
+++ b/terraform/ecs/main.tf
@@ -170,7 +170,7 @@ resource "null_resource" "iam_wait" {
   ]
 
   provisioner "local-exec" {
-    command = "sleep 30"
+    command = "echo iam_wait done"
   }
 }
 
--- a/terraform/ecs/main.tf
+++ b/terraform/ecs/main.tf
@@ -81,9 +81,19 @@ resource "aws_ecs_capacity_provider" "capacityprovider" {
   }
 }
 
+data "aws_ami" "amazon_linux_2" {
+  most_recent = true
+  owners      = ["amazon"]
+
+  filter {
+    name   = "name"
+    values = ["amzn2-ami-ecs-hvm-*-x86_64-ebs"]
+  }
+}
+
 resource "aws_launch_template" "launchtemp" {
   name_prefix   = "launchtemp"
-  image_id      = "ami-0ae546d2dd33d2039"
+  image_id      =  data.aws_ami.amazon_linux_2.image_id
   instance_type = "t2.medium"
   user_data            = "${base64encode("#!/bin/bash\necho ECS_CLUSTER=${module.common.testing_id} >> /etc/ecs/ecs.config")}"
   metadata_options {
--- a/terraform/ecs/main.tf
+++ b/terraform/ecs/main.tf
@@ -227,14 +227,14 @@ resource "aws_ssm_parameter" "otconfig" {
 
 resource "null_resource" "check_patch" {
   depends_on = [
-    data.aws_instances.ecs-instances, aws_instance.collector_efs_ec2]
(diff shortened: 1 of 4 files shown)


terraform/ecs/main.tf
```diff
@@ -58,21 +58,120 @@ provider "aws" {
...
+  depends_on = [
+    aws_iam_role.cluster_instance_role,
+    aws_iam_policy.cluster_instance_policy,
+    aws_iam_policy_attachment.cluster_instance_policy_attachment,
+    aws_iam_instance_profile.cluster,
+  ]
+
+  provisioner "local-exec" {
+    command = "sleep 30"
+  }
```
bryan-aguilar: I understand what you are doing here with this `null_resource` but is the sleep needed? You should be able to depend on `iam_wait` without the sleep. It will only be complete after it's dependencies are resolved. If a provisioner is required I suggest just using something like `echo iam_wait done`.

humivo: I added the sleep there because the previous module we used put a `sleep` there and I was trying to replicate it but I also think an `echo iam_wait done` would do the same and verified it in my local testing.  Changed in my latest commit.

terraform/ecs/main.tf
```diff
@@ -58,21 +58,120 @@ provider "aws" {
...
+  owners      = ["amazon"]
+
+  filter {
+    name   = "name"
+    values = ["amzn2-ami-ecs-hvm-*-x86_64-ebs"]
+  }
+}
+
+resource "aws_launch_template" "launchtemp" {
+  name_prefix   = "launchtemp"
```
rapphil, line 95: do we need to add tags so that ssm knows how to patch this?
https://github.com/aws-observability/aws-otel-test-framework/blob/terraform/terraform/ec2/main.tf#L95

if yes, we also need to wait for the patching to finish before moving on with the tests.

humivo: I've added these tags to the launch template instances in my latest commit

bryan-aguilar: FYI that's not how the SSM patching is triggered in ec2. The `check_patch` resource uses that variable as a flag. https://github.com/aws-observability/aws-otel-test-framework/blob/80f205ee27e1b60c7e110fdc14ddda8385bc1808/terraform/ec2/main.tf#L143

terraform/ecs/variables.tf
```diff
@@ -39,6 +39,13 @@ variable "ecs_taskdef_network_mode" {
   default = "awsvpc"
 }
 
+# if patch is true, we will wait until the instance gets patched after launching the instance,
+# also set a patch tag onto the instance so that the instance get picked by the ssm patching process.
+# and then start the installation of collector.
+variable "patch" {
+  default = true
+}
```
bryan-aguilar, line 47: See comment further above. Patching is not done through tags. We may have to jump through more hoops to do the same type of patching on ecs ec2 instances. I don't consider this a blocker for this PR because it was not being done previously.

bryan-aguilar (approved):

terraform/ecs/variables.tf
humivo: Acknowledged the comment above.  So should I remove the `Patch` tag for now then we can look into adding patching on ecs ec2 instances in the future?

bryan-aguilar: Yes. I would remove the tag because it is not used in the same context as EC2.

terraform/ecs/main.tf
```diff
@@ -141,6 +141,21 @@ resource "aws_autoscaling_group" "clusterasg" {
...
+  instance_tags = {
+    Patch = var.patch
+  }
+
+  filter {
+    name   = "tag:Name"
+    values = ["Integ-test-ecs-instance"]
+  }
+
+  instance_state_names = ["running", "stopped"]
```
bryan-aguilar:
```suggestion
  instance_state_names = ["running", "stopped"]
```
Why do we filter on both `running` and `stopped` rather than just `running`?

humivo: I carried it over from the example.  I think just `running` makes more sense in this case.

bryan-aguilar (approved):

terraform/ecs/main.tf
```diff
@@ -58,21 +58,146 @@ provider "aws" {
...
+
+  roles      = [aws_iam_role.cluster_instance_role.id]
+  policy_arn = aws_iam_policy.cluster_instance_policy.arn
+}
+
+resource "aws_iam_instance_profile" "cluster" {
+  role = aws_iam_role.cluster_instance_role.name
+}
+
+resource "null_resource" "iam_wait" {
```
rapphil, line 190: if nothing is depending on this resource, then there is no need for it right?

terraform/ecs/main.tf
```diff
@@ -100,6 +225,26 @@ resource "aws_ssm_parameter" "otconfig" {
   tier  = "Advanced" // need advanced for a long list of prometheus relabel config
 }
 
+resource "null_resource" "check_patch" {
```
rapphil, line 228: are you adding the proper `depends_on` on other resources so that we don't try to start tests while this resource is finishing?

rapphil: Never mind. I can see that you added.

terraform/ecs/main.tf
rapphil: Never mind. I just spotted.

rapphil (approved):

PR #1130, opened 2023-03-22
feat: Add support to test config map providers

rapphil:
**Description:** Add e2e tests to test config map providers:

* s3
* http
* https

**Testing:** Testing was done manually in each of the supported platforms.

**Documentation:** A new readme file was added describing how tests are performed.


terraform/ecs/variables.tf
```diff
@@ -62,3 +62,8 @@ variable "ecs_extra_apps" {
 variable "kafka_version" {
   default = ""
 }
+
+// Source of the configurations in ECS: env_var, s3, http and https
+variable "configuration_source" {
```
bryan-aguilar, line 67: https://developer.hashicorp.com/terraform/language/values/variables#custom-validation-rules
Should we start using validation rules to enforce that the value falls within the set defined in the comment?

terraform/templates/configmap-providers/cloudwatch_context.json
```diff
@@ -0,0 +1,3 @@
+{
+  "clusterName": "${cluster_name}"
+}
```
bryan-aguilar:
```suggestion
}

```

terraform/testcases/configmap_provider_http/otconfig.tpl
```diff
@@ -0,0 +1,37 @@
...
+        match_type: regexp
+        action: update
+        operations:
+          - action: add_label
+            new_label: "testingId"
+            new_value: "${testing_id}"
+
+exporters:
+  logging:
+    loglevel: debug
```
bryan-aguilar, line 24: Do we NEED this? This will generate a lot of extra log lines in the GHA workflow logs.

rapphil: this is logged inside the environment where the collector is running. we don't push those logs to the GHA workflow, do we?

bryan-aguilar: Good point, thank you for catching that.

Aneurysm9: Isn't there a template value used for this in other config templates?

edit: I appear to be wrong.  Ignore me.

bryan-aguilar: It's used in this same config but the `loglevel` values are not the same for the `logging` exporter. I think the logging exporter uses something like `detailed` and something else instead of `debug`.

terraform/ecs/variables.tf
rapphil: TIL. good idea.

bryan-aguilar (approved):

PR #1147, opened 2023-04-03
Make autoscalinggroup names unique to avoid duplicate asgs being created

humivo:
**Description:** Changed the asg naming to `name_prefix` which will take the prefix and add create a unique asg name afterwards.

**Testing:** Local testing passed.

### DIFF
--- a/terraform/ecs/main.tf
+++ b/terraform/ecs/main.tf
@@ -59,7 +59,7 @@ data "aws_caller_identity" "current" {
 }
 
 resource "aws_ecs_cluster" "ecscluster" {
-  name = module.common.testing_id
+  name = "aoc-testing-${module.common.testing_id}"
 
   depends_on = [
     null_resource.iam_wait
@@ -74,7 +74,7 @@ resource "aws_ecs_cluster_capacity_providers" "clustercapacity" {
 }
 
 resource "aws_ecs_capacity_provider" "capacityprovider" {
-  name = "capacityprovider"
+  name = "capacityprovider-${module.common.testing_id}"
 
   auto_scaling_group_provider {
     auto_scaling_group_arn = aws_autoscaling_group.clusterasg.arn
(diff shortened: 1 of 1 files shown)


terraform/ecs/main.tf
```diff
@@ -126,7 +126,7 @@ resource "aws_launch_template" "launchtemp" {
 }
 
 resource "aws_autoscaling_group" "clusterasg" {
-  name                = "clusterasg"
+  name_prefix         = "clusterasg"
```
bryan-aguilar: Can we use the testing id here or at least tag the auto scaling group with the testing ID?

humivo: Added it and changed the other names

terraform/ecs/main.tf
```diff
@@ -126,11 +126,23 @@ resource "aws_launch_template" "launchtemp" {
 }
 
 resource "aws_autoscaling_group" "clusterasg" {
-  name                = "clusterasg"
+  name_prefix         = "clusterasg-${module.common.testing_id}-"
```
rapphil, line 129: should the name be different? https://github.com/aws-observability/aws-otel-collector/blob/main/tools/workflow/cleaner/autoscaling/clean.go#L29

terraform/ecs/main.tf
```diff
@@ -126,11 +126,23 @@ resource "aws_launch_template" "launchtemp" {
...
   vpc_zone_identifier = module.basic_components.aoc_private_subnet_ids
   desired_capacity    = 1
   max_size            = 10
   min_size            = 1
+  tags = [
+    {
+      "key"                 = "ephermeral"
```
bryan-aguilar:
```suggestion
      "key"                 = "ephemeral"
```

bryan-aguilar (approved):

rapphil (approved):

PR #1171, opened 2023-04-11
Add infra for lambda integration tests

rapphil:
**Description:** Add bucket for integration tests of confmap providers in lambda.

### DIFF
--- a/cdk_infra/lib/app.ts
+++ b/cdk_infra/lib/app.ts
@@ -10,7 +10,7 @@ const envDefault = {
   account: process.env.CDK_DEFAULT_ACCOUNT,
   region: process.env.CDK_DEFAULT_REGION
 };
-
+// Lambda CI runs in us-east-1
 const lambdaDefaults = {
   account: process.env.CDK_DEFAULT_ACCOUNT,
   region: "us-east-1"
--- a/cdk_infra/lib/app.ts
+++ b/cdk_infra/lib/app.ts
@@ -10,6 +10,12 @@ const envDefault = {
   account: process.env.CDK_DEFAULT_ACCOUNT,
   region: process.env.CDK_DEFAULT_REGION
 };
+
+const lambdaDefaults = {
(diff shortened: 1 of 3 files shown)


cdk_infra/lib/app.ts
```diff
@@ -10,6 +10,12 @@ const envDefault = {
   account: process.env.CDK_DEFAULT_ACCOUNT,
   region: process.env.CDK_DEFAULT_REGION
 };
+// Lambda CI runs in us-east-1
+const lambdaDefaults = {
+  account: process.env.CDK_DEFAULT_ACCOUNT,
+  region: "us-east-1"
```
bryan-aguilar: Is the deployment workflow setup to handle this? Will there be issues with `AWS_REGION` env var that gets set? Is the deployment workflow ready for multi region deployments?

cdk_infra/lib/app.ts
```diff
@@ -10,6 +10,12 @@ const envDefault = {
   account: process.env.CDK_DEFAULT_ACCOUNT,
   region: process.env.CDK_DEFAULT_REGION
 };
+// Lambda CI runs in us-east-1
+const lambdaDefaults = {
+  account: process.env.CDK_DEFAULT_ACCOUNT,
+  region: 'us-east-1'
```
bryan-aguilar, line 16: Is the deployment workflow prepared for multi region deployments? Will there be any issue with the `AWS_REGION` env var being set?

rapphil: that is why we need to set the region explicitly here.

rapphil: I think we will need to bootstrap the region though.

bryan-aguilar: Okay good, I was hoping the explicit region fixed that. Agreed on bootstrapping.

bryan-aguilar (approved):

PR #1283, opened 2023-06-21
Eks operator java agent testcase

PaurushGarg:

Adding test-cases (2 nos) to validate metrics and traces emitted by auto-instrumentation injected through the Opentelemetry-Operator for the Spark Sample App (without Java Agent).

Followup PRs:

- [x] Add Clusters and Entry to `testcases.json`: https://github.com/aws-observability/aws-otel-test-framework/pull/1285/files
- [x] Workflow: https://github.com/aws-observability/aws-otel-java-instrumentation/pull/447
- [x] IAM Roles:

Manually tested by deploying through local

- [x] TODO: Publish Spark Sample App (without Java Agent)


terraform/eks/otlp.tf
```diff
@@ -291,6 +293,42 @@ resource "null_resource" "aoc_deployment_adot_operator" {
   depends_on = [module.adot_operator]
 }
 
+data "template_file" "java_auto_instrumentation_config_file" {
```
rapphil: I think you can use the `templatefile` function instead of this data module. https://developer.hashicorp.com/terraform/language/functions/templatefile

Documentation suggest that the function should be used instead of the module. https://registry.terraform.io/providers/hashicorp/template/latest/docs/data-sources/file.html

terraform/testcases/otlp_trace_auto_instrumentation_adot_operator/README.md
```diff
@@ -0,0 +1 @@
+This tests uses java-auto-instrumentation to instrument the spark sample app image (without java agent) using Opentelemetry-Operator. Opentelemetry-Operator here also deploys and manages the ADOT-Collector. Validation and other rest functionality is similar to otlp_trace_adot_operator testcase. This testcase validates traces emitted using java-auto-instrumentation deployed through Opentelemetry-Operator.
```
rapphil:
```suggestion
# Description 

This testcase validates traces emitted using java-auto-instrumentation deployed through Opentelemetry-Operator. This test uses the ADOT java-auto-instrumentation image to instrument a Spark sample-app image (without java-agent baked in) using the OpenTelemetry Operator. The OpenTelemetry Operator also deploys and manages the ADOT collector. The validation and other functionality is similar to otlp_trace_adot_operator testcase. 
```

terraform/eks/otlp/push_mode_samples_inject_auto_instrumentation.tf
```diff
@@ -0,0 +1,96 @@
...
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+# -------------------------------------------------------------------------
+
+##########################################
+# Push mode deployment to inject with auto_instrumentation
+##########################################
+
+# deploy sample app
+resource "kubernetes_deployment" "push_mode_sample_app_deployment_inject_auto_instrumentation" {
```
bryan-aguilar: Is it possible to consolidate this sample app deployment with the other sample app deployment? It seems that not much changes other than some labels and the image being used.

terraform/eks/variables.tf
```diff
@@ -80,3 +80,19 @@ variable "configuration_source" {
...
+
+variable "java_auto_instrumentation_repository" {
+  type    = string
+  default = "public.ecr.aws/aws-observability/adot-autoinstrumentation-java"
+}
+
+# latest tag is not available
+variable "java_auto_instrumentation_tag" {
+  type    = string
+  default = "v1.27.0"
```
bryan-aguilar: Having a default here doesn't seem right. I feel like there could be a scenario where the default is used unintentionally. I think we should not have a default which will lead to test failures instead.

terraform/testcases/otlp_metric_auto_instrumentation_adot_operator/otconfig.tpl
```diff
@@ -0,0 +1,24 @@
...
+      logging:
+        verbosity: detailed
+      awsemf:
+        region: '${region}'
+
+    service:
+      pipelines:
+        metrics:
+          receivers: [otlp]
+          processors: [ ]
```
bryan-aguilar: Why are we not using the batch processor here?

terraform/eks/otlp.tf
PaurushGarg: thanks. updated accordingly.

terraform/eks/otlp/push_mode_samples_inject_auto_instrumentation.tf
PaurushGarg: @bryan-aguilar I tried that initially but the environment variables and annotations required in these two cases are different. 
For auto-instrumentation through Operator, I preferred to inject OTLP specific env variables through the Operator Instrumentation template file. And the test-case fails if we use the `push_mode_samples.tf` instead of using `push_mode_samples_inject_auto_instrumentation`, b/c of the difference in env variables.

terraform/testcases/otlp_metric_auto_instrumentation_adot_operator/otconfig.tpl
PaurushGarg: Added.

terraform/eks/otlp/push_mode_samples_inject_auto_instrumentation.tf
PaurushGarg: @bryan-aguilar I can consolidate the two sample app deployment into one. But then it's not possible to pass the exporter endpoints through the `java_auto_instrumentation.tpl`, because currently the `OTEL_EXPORTER_OTLP_ENDPOINT` env variable gets set in `push_mode_samples.tf` and I couldn't get the Operator to override the env values provided with sample_app deployment.  
I thought the right way for this test, is to provide all the OTLP related env variables through the Instrumentation template and not through sample app deployment. And for this reason created a new deployment for sample app deployment without OTLP env variables. 
Do you want me to consolidate the apps deployment and keep the OTLP env variables set through the sample-app deployment?

terraform/eks/variables.tf
PaurushGarg: Removed the default value.

terraform/testcases/otlp_trace_auto_instrumentation_adot_operator/README.md
PaurushGarg: Updated.

terraform/eks/variables.tf
```diff
@@ -81,6 +81,21 @@ variable "configuration_source" {
...
+  default = false
+}
+
+variable "java_auto_instrumentation_repository" {
+  type    = string
+  default = "public.ecr.aws/aws-observability/adot-autoinstrumentation-java"
+}
+
+# latest tag is not available
+variable "java_auto_instrumentation_tag" {
```
rapphil, line 95: where are you setting the value for this variable?

terraform/eks/adot-operator/java_auto_instrumentation.tpl
```diff
@@ -0,0 +1,15 @@
...
+  name: my-instrumentation
+  namespace: ${AOC_NAMESPACE}
+spec:
+  exporter:
+    endpoint: http://aoc-collector:${GRPC_PORT}
+  env:
+   - name: OTEL_SERVICE_NAME
+     value: ${SERVICE_NAME}
+   - name: OTEL_RESOURCE_ATTRIBUTES
+     value: "service.namespace=${SERVICE_NAMESPACE},service.name=${SERVICE_NAME}"
```
rapphil: setting `OTEL_SERVICE_NAME` and `service.name` in `OTEL_RESOURCE_ATTRIBUTES` is redundant. pick one of them.

terraform/eks/variables.tf
PaurushGarg: @rapphil as per [this](https://github.com/aws-observability/aws-otel-test-framework/pull/1283#discussion_r1242461105) comment from Bryan, I have removed the default value of this variable. So when running this testcase manually the terraform prompts to provide this variable value. But we are passing the value of this variable from [here](https://github.com/aws-observability/aws-otel-java-instrumentation/blob/4c42d3a4dbe418b01d2e1c384bea0f0c4a93367a/.github/workflows/main-build.yml#L255) in mainbuild workflow in java-instrumentation repo.

terraform/eks/adot-operator/java_auto_instrumentation.tpl
PaurushGarg: thanks. Updated. I have removed `service.name `from `OTEL_RESOURCE_ATTRIBUTES` as setting this doesn't set the `OTEL_SERVICE_NAME`. And `OTEL_SERVICE_NAME` is required to be set to ensure the metrics are exported to desired namespace in Cloudwatch.

cdk_infra/lib/config/cluster-config/clusters.yml
```diff
@@ -71,3 +71,23 @@ clusters:
...
+  - name: java-instrumentation-operator-ci-arm64-1-23
+    version: "1.23"
+    launch_type: ec2
+    instance_type: m6g.large
+    cert_manager: true
+  - name: java-instrumentation-operator-ci-arm64-1-26
+    version: "1.26"
+    launch_type: ec2
+    instance_type: m6g.large
+    cert_manager: true
```
bryan-aguilar, line 93: Wasn't this merged in this PR? https://github.com/aws-observability/aws-otel-test-framework/pull/1285/files

terraform/testcases/otlp_trace_auto_instrumentation_adot_operator/parameters.tfvars
```diff
@@ -0,0 +1,7 @@
+validation_config = "spark-otel-trace-validation.yml"
+
+sample_app = "spark"
+
+sample_app_image = "public.ecr.aws/aws-otel-test/aws-otel-java-spark-without-auto-instrumentation-agent:latest"
```
bryan-aguilar, line 5: Was this changed to be a private repo? Does this repo even exist?

terraform/eks/otlp/variables.tf
```diff
@@ -50,3 +50,13 @@ variable "aoc_service" {
...
+
+variable "is_inject_auto_instrumentation" {
+  type = bool
+}
+
+# is_adot_operator is true if the current test case is testing ADOT Operator.
+variable "is_adot_operator" {
+  type    = bool
+  default = false
+}
```
bryan-aguilar, line 62: Is this just a relocation from the `push_mode_sample_app` file?

terraform/eks/otlp/push_mode_samples_inject_auto_instrumentation.tf
bryan-aguilar: My thoughts are that we should do our best to consolidate here to improve readability. Does terraform allow you to optionally define these env variables? Only create them in when `is_inject_auto_instrumentation` is false? 

If we are going to keep them separate deployments then I'd like to see the services moved into a separate file. I think this indicates that they will always be created in an otlp test case where these deployments are optional depending on some condition.

terraform/eks/otlp.tf
```diff
@@ -291,6 +293,34 @@ resource "null_resource" "aoc_deployment_adot_operator" {
...
+      SERVICE_NAMESPACE                    = module.common.otel_service_namespace
+      SERVICE_NAME                         = module.common.otel_service_name
+      JAVA_AUTO_INSTRUMENTATION_REPOSITORY = var.java_auto_instrumentation_repository
+      JAVA_AUTO_INSTRUMENTATION_TAG        = var.java_auto_instrumentation_tag
+  })
+
+  depends_on = [module.adot_operator]
+}
+
+resource "null_resource" "java_auto_instrumentation_adot_operator" {
```
bryan-aguilar: I don't think the pattern of defining a template file and then using a null resource is a best practice here. I know we may do it in other places but I believe there is a better way to do this. Custom resources can be deployed using the `kubernetes_manifest` module. See a writeup from hashicorp [here](https://developer.hashicorp.com/terraform/tutorials/kubernetes/kubernetes-crd-faas#deploy-an-openfaas-function) on how to manage custom resources.

terraform/eks/otlp/push_mode_samples_inject_auto_instrumentation.tf
PaurushGarg: @bryan-aguilar consolidated `push_mode_samples.tf`.

cdk_infra/lib/config/cluster-config/clusters.yml
PaurushGarg: [This](https://github.com/aws-observability/aws-otel-test-framework/pull/1285/files) PR's target was not based of`terraform` branch but `eks_operator_java_agent_testcase` branch. Hence it was merged to branch of this PR. 
Originally that PR contained changes that were dependent on this PR, so I created that PR dependent on this PR.

terraform/eks/otlp/variables.tf
PaurushGarg: yes, I just relocated this variable to `variables.tf` to keep the variables in one file.

terraform/eks/otlp.tf
PaurushGarg: @bryan-aguilar yes this is how it is currently done [here](https://github.com/aws-observability/aws-otel-test-framework/blob/0afc4eb0110ad4e3c1dffcab5b4773bcdbb310f6/terraform/eks/otlp.tf#L284) for adot_operator's `adot_collector_deployment` as well, so I reused this component. I am going to look into the proposed method and comeback.

PaurushGarg: @bryan-aguilar thanks. Updated accordingly.

terraform/testcases/otlp_trace_auto_instrumentation_adot_operator/parameters.tfvars
PaurushGarg: @bryan-aguilar  No, this was kept as public repo (same as `spark` sample-app with agent). This repo doesn't exist yet. But under this [PR](https://github.com/aws-observability/aws-otel-ops/pull/356) we created this repo [here](https://github.com/aws-observability/aws-otel-ops/blob/378b45801e1aaa0e63da1153f5056102991bd6b2/iam-role-setup/lib/config/ecr/611364707713/ecr-config.ts#L58).  This PR is now merged.

terraform/eks/otlp.tf
```diff
@@ -291,6 +293,26 @@ resource "null_resource" "aoc_deployment_adot_operator" {
   depends_on = [module.adot_operator]
 }
 
+resource "kubectl_manifest" "java_auto_instrumentation_deployment" {
```
bryan-aguilar: Where does `kubectl_manifest` module come from? Who is the provider?

Why not use `kubernetes_manifest` in the official `kubernetes` provider? 
https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest

terraform/eks/otlp/push_mode_samples_inject_auto_instrumentation.tf
bryan-aguilar: What was your strategy for resolving your previous concerns? I don't see any changes related to the environment variables provided to the deployment.

PaurushGarg: @bryan-aguilar : I am not providing the OTLP related env variables through the `kind:Instrumentation` template now. By not doing that, I was able to consolidate the workload deployment templates.

bryan-aguilar: So is this a tradeoff? There is no way to do both?

terraform/eks/otlp.tf
PaurushGarg: I was getting `kubernetes_manifest unmarshaling unknown values is not supported` error with the same yaml template with the `kubernetes_manifest`. 
`kubectl_manifest`:  https://registry.terraform.io/providers/cpanato/kubectl/latest/docs/resources/kubectl_manifest

terraform/eks/otlp/push_mode_samples_inject_auto_instrumentation.tf
PaurushGarg: @bryan-aguilar yes its a trade off ( I couldn't find a way to override env variables already provided at workload deployment stage). but also discussed this with @rapphil, he also mentioned that since priority of this test is not to test whether the env variables to the workload can be set through the instrumentation crd template - so we should consolidate the workload deployment template to increase readability.

terraform/eks/otlp.tf
PaurushGarg: Updated accordingly with `kubernetes_manifest`.

bryan-aguilar (approved):

PR #1285, opened 2023-06-21
Adding java_auto_instrumentation testcases to batchtestgenerator

PaurushGarg:

Adds Clusters for: https://github.com/aws-observability/aws-otel-test-framework/pull/1283
This PR is dependent on : https://github.com/aws-observability/aws-otel-test-framework/pull/1283


tools/batchTestGenerator/testcases.json
```diff
@@ -118,6 +118,32 @@
 					"region": "us-west-2"
 				}
 			]
+		},
+		{
+			"type": "EKS_JAVA_INSTRUMENTATION_ADOT_OPERATOR",
```
rapphil: if you add the testcases to this file, they will be run by the collector. we need a new testcase file, that can live in the java repository. 

take a look into how the collector create the testcases: https://github.com/aws-observability/aws-otel-collector/blob/main/.github/workflows/CI.yml#L569

PaurushGarg: @rapphil I noticed that all repos have their own copy of testcases.json (adot-operator, adot-collector) along with test-framework. but I couldn't understand why that is required, when we do have argument to pick only certain services (like its done [here](https://github.com/aws-observability/aws-otel-collector/blob/bdeb7a6a4dba7c3a39813c1b1cee9db31c39fd31/.github/workflows/CI.yml#L51) for collector) then there should be no risk of collector running testcases of other services (like EKS_JAVA_INSTRUMENTATION_ADOT_OPERATOR), right? Wanted to check this with you?

rapphil: that is not a requirement, in fact, we should move the testcase.json file to the adot collector repository and keep the tool in the test framework repository. The inclusion mechanism in the ADOT collector workflow is used for the case we want to run tests in the dev branch.

PaurushGarg: ok. in this case, should I update the `testcases.json` that we have in Collector repo or clone a new one in java-instrumentation repo?

cdk_infra/lib/config/cluster-config/clusters.yml
```diff
@@ -71,3 +71,23 @@ clusters:
     version: "1.26"
     launch_type: ec2
     instance_type: "m5.large"
+  - name: java-instrumentation-operator-ci-amd64-1-22
+    version: "1.22"
+    launch_type: ec2
+    instance_type: m5.large
+    cert_manager: true
```
bryan-aguilar: 1.22 is no longer supported.

tools/batchTestGenerator/internal/testCaseBuilder.go
```diff
@@ -65,7 +65,7 @@ func buildTestCases(runConfig RunConfig) ([]TestCaseInfo, error) {
 						}
 						newTests = append(newTests, newTest)
 					}
-				case "EKS_FARGATE", "EKS_ADOT_OPERATOR", "EKS_ADOT_OPERATOR_ARM64", "EKS", "EKS_ARM64":
+				case "EKS_FARGATE", "EKS_ADOT_OPERATOR", "EKS_ADOT_OPERATOR_ARM64", "EKS", "EKS_ARM64", "EKS_JAVA_INSTRUMENTATION_ADOT_OPERATOR", "EKS_JAVA_INSTRUMENTATION_ADOT_OPERATOR_ARM64":
```
bryan-aguilar: Why do these need to be two new test case types rather than use the existing `EKS_ADOT_OPERATOR` or `EKS_ADOT_OPERATOR_ARM64` types?

tools/batchTestGenerator/testcases.json
bryan-aguilar: I concur, please add a `testcases.json` file in the `java-instrumentation` repo to generate a list of test cases. The collector does not currently do this but that's a deficiency in the collector.

tools/batchTestGenerator/internal/testCaseBuilder.go
PaurushGarg: @bryan-aguilar I noticed that adot-operator workflow is running the `testcases` that are included in the `EKS_ADOT_OPERATOR` or `EKS_ADOT_OPERATOR_ARM64` tags. 
I created the new tag exclusive for java-instrumentation as I wanted to exclude them from all other scopes of adot-Collector and adot-Operator testcases. Reason: To not duplicate the running of same testcases with adot-operator workflow.

bryan-aguilar: The test cases that are being ran is dependent on which tests are defined in the `testcases.json` file. If you implement the advice given to you elsewhere, to give the java-instrumeentation it's own `testcases.json` file you can only define the test cases necessary.

PaurushGarg: @bryan-aguilar. I can change that but forgot to mention one primary reason. 
Tag (like `EKS_ADOT_OPERATOR`) is mapped to a cluster. Since java instrumentation tests are deployed on different cluster, so the new `testcases.json` in `java-instrumentation`, will have different cluster values than the other similar files in other repos. I originally thought this is going to be confusing so thought of creating a new tag specific for `java-instrumentation`. 
Do you think its ok to map the `tag` to `cluster` here differently by reusing the `EKS_ADOT_OPERATOR` tag?

bryan-aguilar: A `testcases.json` file should represent the test case configuration required for that specific repository. We SHOULD have a specific `testcases.json` file for each repository. We do not now, but will in the near future. 

I think it's more confusing to separate these into different platform types. There is no reason that we could not run java instrumentation tests within the `ADOT_OPERATOR` test case scope right? In the end these are all tests of the operator.

tools/batchTestGenerator/testcases.json
PaurushGarg: Removed changes in testcases.json and batchtestgenerator from here.

cdk_infra/lib/config/cluster-config/clusters.yml
PaurushGarg: Updated to 1.23 as min version.

tools/batchTestGenerator/internal/testCaseBuilder.go
PaurushGarg: Removed changes in testcases.json and batchtestgenerator from here.

rapphil (approved):

PR #1302, opened 2023-06-27
Add ability to ignore empty dimension set for EMF Tests

bryan-aguilar:
**Description:** This PR adds the ability for the CW Metric Validator to ignore an additional unexpected empty dimension set metric. Test cases need to provide this flag through their `parameters.tfvars` files to indicate that the validator should should this unexpected dimension set. 

Testing: Added unit test and also tested with a new integration test case that will be added in a followup PR. 

### DIFF
--- a/terraform/eks/variables.tf
+++ b/terraform/eks/variables.tf
@@ -85,5 +85,5 @@ variable "ignore_empty_dim_set" {
   type    = bool
   default = false
 
-  description = "Toggles whether or not the validataor will ignore an empty EMF dimension set"
+  description = "Toggles whether or not the validator will ignore an empty EMF dimension set"
 }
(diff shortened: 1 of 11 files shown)


terraform/eks/variables.tf
```diff
@@ -80,3 +80,10 @@ variable "configuration_source" {
     error_message = "Invalid configuration_source for eks"
   }
 }
+
+variable "ignore_empty_dim_set" {
+  type    = bool
+  default = false
+
+  description = "Toggles whether or not the validataor will ignore an empty EMF dimension set"
```
rapphil:
```suggestion
  description = "Toggles whether or not the validator will ignore an empty EMF dimension set"
```

rapphil (approved):

PR #1303, opened 2023-06-28
Add load balancing validator and unit tests

humivo:
**Description:** Adding a new validator to the test framework to test the load balancing exporter component.  With this new validator, there are also unit tests implemented.

**Testing:** Local testing completed and unit tests successfully run on local.


validator/src/main/java/com/amazon/aoc/exception/ExceptionCode.java
```diff
@@ -34,6 +34,8 @@ public enum ExceptionCode {
   EMPTY_LIST(50007, "list is empty or null"),
   LOG_FORMAT_NOT_MATCHED(50008, "log format not matched"),
   HEALTH_STATUS_NOT_MATCHED(50009, "health_check status not matched"),
+  NOT_ENOUGH_SPANS(50010, "not enough spans in the trace"),
+  COLLECTOR_ID_NOT_MATCHED(50011, "collector id of spans not matched"),
```
bryan-aguilar:
```suggestion
  COLLECTOR_ID_NOT_MATCHED(50011, "span collector-id attribute does not match"),
```

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+  private ICaller caller;
+  private Context context;
+
+  private static final ObjectMapper MAPPER =
+      new ObjectMapper().setPropertyNamingStrategy(PropertyNamingStrategy.SNAKE_CASE);
+
+  // for unit test
+  public void setXRayService(XRayService xrayService) {
+    this.xrayService = xrayService;
+  }
```
bryan-aguilar: Creating a method just for unit testing does not seem right. I think there is probably another way to go about this. Possibly change the Class `init` method to accept an `XRayService` object.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class LoadBalanceValidator implements IValidator {
```
bryan-aguilar: This class shares a lot of functionality as `TraceValidator.java`. That functionality should be abstracted away into where duplicate bits of logic can be shared.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+      throws Exception {
+    this.xrayService = new XRayService(context.getRegion());
+    this.caller = caller;
+    this.context = context;
+  }
+
+  @Override
+  public void validate() throws Exception {
+    // 2 retries for calling the sample app to handle the Lambda case,
+    // where first request might be a cold start and have an additional unexpected subsegment
```
bryan-aguilar: I'm not sure this comment makes sense. What is being retried twice?

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+    int successes = 0;
+    while (successes < 5) {
+      // Call sample app and get locally stored trace
+      String traceId = this.getAppTraceId();
```
bryan-aguilar: I'm not sure if this comment or method name makes sense. The purpose of the method is to call the sample app right? Maybe consider renaming the method. It's still okay to return the traceid but that should be documented in method about its return value.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+      List<String> traceIdList = Collections.singletonList(traceId);
```
bryan-aguilar: Why is this being stored in a list? Isn't this a single trace id?

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+      ;
```
bryan-aguilar: Rogue semicolon?

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+      log.info("value of Sample App traceId: {}", traceId);
```
bryan-aguilar: I'm not sure we need to log this. 

nit: We should be careful about the logging we add in the validator. It can create messy and large log outputs in the github workflow where it is normally called. Later on in this method we log the entire trace which includes the trace id.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+
+      // Retry 5 times to since segments might not be immediately available in X-Ray service
+      RetryHelper.retry(
+          5,
+          Integer.parseInt(GenericConstants.SLEEP_IN_MILLISECONDS.getVal()),
+          true,
+          () -> {
+            // get retrieved trace from x-ray service
+            Map<String, Object> retrievedTrace = this.getRetrievedTrace(traceIdList);
+            log.info("value of retrieved trace map: {}", retrievedTrace);
```
bryan-aguilar: Same comment here about logging to aggressively. I don't think this is needed.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+            Set<String> set =
+                retrievedTrace.keySet().stream()
+                    .filter(s -> s.endsWith("collector-id"))
```
bryan-aguilar:
```suggestion
                    .filter(s -> s.endsWith("collector-id"))
```
Why `endsWith` here? Is it due to how the trace keys are mapped? What's the full name for this key?

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+                    .collect(Collectors.toSet());
+            if (set.size() < 2) {
+              log.error("not enough spans in trace map (need at least 2)");
+              log.info("==========================================");
+              throw new BaseException(ExceptionCode.NOT_ENOUGH_SPANS);
+            }
```
bryan-aguilar: nit: lets extract this into into it's own method. Something like `checkSpanCount`.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+            String targetCollectorId = null;
```
bryan-aguilar: nit: Instead of setting this to `null` you could set it to the first value in your collection. Then iterate from `1 to N`.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+                  log.info("value of retrieved id: {}", retrievedTrace.get(collectorId).toString());
+                  log.info("==========================================");
+                  throw new BaseException(ExceptionCode.COLLECTOR_ID_NOT_MATCHED);
+                }
+              }
+            }
+          });
+
+      successes += 1;
+      log.info("Total number of successful runs: {}", successes);
```
bryan-aguilar: nit: this log line does not seem necessary.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    log.info("validation is passed for path {}", caller.getCallingPath());
+  }
+
+  // this method will hit get trace from x-ray service and get retrieved trace
+  private Map<String, Object> getRetrievedTrace(List<String> traceIdList) throws Exception {
+    List<Trace> retrieveTraceList = xrayService.listTraceByIds(traceIdList);
```
bryan-aguilar: Does XRayService allow us to query by a single trace id instead of a collection of them?

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
```diff
@@ -0,0 +1,170 @@
...
+    if (retrieveTraceList == null || retrieveTraceList.isEmpty()) {
+      throw new BaseException(ExceptionCode.EMPTY_LIST);
+    }
+
+    return this.flattenDocument(retrieveTraceList.get(0).getSegments());
```
bryan-aguilar: Again, the use of a list here does not make sense. We are always returning the first set of segments.

validator/src/main/java/com/amazon/aoc/validators/ValidatorFactory.java
```diff
@@ -47,6 +47,9 @@ public IValidator launchValidator(ValidationConfig validationConfig) throws Exce
         validator = new TraceValidator();
         expectedData = validationConfig.getExpectedTraceTemplate();
         break;
+      case "loadbalancing":
+        validator = new LoadBalanceValidator();
+        break;
```
bryan-aguilar: nit: If we are going to use `loadbalancing` here we should rename the validator to `LoadBalancingValidator`.

validator/src/main/resources/validations/load-balancing-validation.yml
```diff
@@ -0,0 +1,5 @@
+-
+  validationType: "loadbalancing"
+  httpPath: "/gen-mult-spans"
+  httpMethod: "get"
+  callingType: "http"
```
bryan-aguilar: missing EOF new line.

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+  /** test validation when it succeeds */
+  @Test
+  public void testValidationSucceeds() throws Exception {
+    ValidationConfig validationConfig = new ValidationConfig();
+    validationConfig.setCallingType("http");
+    Context context = initContext();
+
+    String mockedTrace =
+        "[{ \"segments\": [{ \"document\": '{ \"metadata\": { \"default\": {\"collector-id\": 3} }, \"subsegments\": [{ \"metadata\": { \"default\": {\"collector-id\": 3} }}, { \"metadata\": { \"default\": {\"collector-id\": 3} }}] }' }] }]";
```
bryan-aguilar: Having to manually escape JSON strings is a painful developer experience. Let's find another way to provide this mocked trace JSON. Possibly a testdata json file or some type of raw string if java supports it.

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+    List<Trace> mockedTraceList =
+        mapper.readValue(
+            mockedTrace.getBytes(StandardCharsets.UTF_8), new TypeReference<List<Trace>>() {});
+
+    runValidation(validationConfig, context, mockedTraceList);
+  }
+
+  /** test validation when fails due to difference collector ids */
```
bryan-aguilar:
```suggestion
  /** test validation when fails due to different collector ids */
```

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
```diff
@@ -0,0 +1,137 @@
...
+        mapper.readValue(
+            mockedTrace.getBytes(StandardCharsets.UTF_8), new TypeReference<List<Trace>>() {});
+
+    BaseException e =
+        assertThrows(
+            BaseException.class, () -> runValidation(validationConfig, context, mockedTraceList));
+    assertEquals(e.getCode(), ExceptionCode.COLLECTOR_ID_NOT_MATCHED.getCode());
+  }
+
+  /** test validation when it fails due to not enough spans for the test */
```
bryan-aguilar: I don't think this comment is necessary. Your test name does a good job of explaining what the test does.

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
```diff
@@ -0,0 +1,137 @@
...
+    // fake vars
+    String namespace = "fakednamespace";
+    String testingId = "fakedTesingId";
+    String region = "us-west-2";
+
+    // faked context
+    Context context = new Context(testingId, region, false, true);
+    context.setMetricNamespace(namespace);
+    context.setCloudWatchContext(new CloudWatchContext());
+    context.getCloudWatchContext().setIgnoreEmptyDimSet(false);
```
bryan-aguilar: Is this actually required to be set for the trace validator tests?

bryan-aguilar (commented):
I left quite a few comments around the usage of a `List` type for retrieving the trace ids from x-ray. I am not quite sure if we are bound to that interface due to previous implementation choices in TraceValidator.java. I think the first step we should do is look to see how that code can be shared between these two classes. Then lets look to see if that list usage is actually required.

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
rapphil: If we could use validator with jdk17, we could use multiline strings and therefore no scaping would be needed.

I did a bit of research and other solutions include using hashmaps to build the json https://www.baeldung.com/java-org-json#1-creating-jsonarray

no good solution here :(

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
rapphil: this is actually a common pattern in Java https://timkranen.medium.com/using-visiblefortesting-to-test-your-private-functions-c9e511fd1c2a but many people don't like.

I think Bryan is right here. You could overload the init method and create a version that accepts a XrayService as parameter.

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
bryan-aguilar: Can we just read in and deserialize from a `.json` file instead?

validator/src/main/java/com/amazon/aoc/exception/ExceptionCode.java
humivo: Changed in my next commit.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: Hmm does this mean we would have to change the CWMetricValidator as well?  In the CWMetricValidator class, it has a [method](https://github.com/aws-observability/aws-otel-test-framework/blob/35a8e23ee4fb81c5d69b023826132266f9639bdd/validator/src/main/java/com/amazon/aoc/validators/CWMetricValidator.java#L50) for unit tests called setCloudWatchService().  Also, the init method cannot be overloaded it is not present in the interface.  It seems like the other validators use the init() method to only set the parameters needed.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: Comment has been removed.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: I changed the method name to getSampleAppResponse() and documented the return value in the method.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: Removed.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: I believe this can be helpful as I remember sometimes during testing, the traceId could be present but the entire trace cannot be found for some reason.  Having this log could show that the sample app was called and a traceId was returned but the xrayService could not retrieve the full trace.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
bryan-aguilar: In java, can't you define multiple of the same method names as long as they have different parameters?

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: This comment could also helpful to see the entire trace returned to, for example, see why the test failed.

bryan-aguilar: Isn't the entire trace logged on failure?

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: Yes, because the key for the trace map that I iterate over would look something like `[0].metadata.default.collector-id`.  Therefore, I am only looking for if the key ends with `collector-id` to find the attribute.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: Extracted into its own method.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: I now set this to the first value in the set and then remove it to iterate through the rest.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: Removed in my commit, but would it not be helpful to see that it succeeded?  Or can we assume the user would know it did by seeing that the retry count did not go down and no errors came up?

validator/src/main/java/com/amazon/aoc/validators/ValidatorFactory.java
humivo: Renamed to `LoadBalancingValidator`

validator/src/main/resources/validations/load-balancing-validation.yml
humivo: Added EOF new line.

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
humivo: Changed now in my commit.

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
humivo: Removed now in my commit.

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
humivo: It was not necessary.  Removed now in my commit.

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
humivo: I saw that with a higher JDK version we would use multiline strings which would be helpful but not sure if we want to do that.  I think reading in and deserializing from a `.json` could be a good alternative but we would have to create a folder and manage those files.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: I have now abstracted a class that this one now extends and overoaded the init() method to change the XrayService.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: I have abstracted duplicate logic to an `XrayValidator`

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: I have found that we store it in a singleton list because this is the [method](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/xray/model/BatchGetTracesRequest.html#:~:text=was%20not%20specified.-,traceIds,-public%20final%C2%A0) required to get the segments.  It takes in a list of traceIds.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: In this specific validator, the entire trace is not logged on failure.  That's why I believe it would be helpful to log the retrieved trace map every retry as the TraceValidator does.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: I have found that we store it in a singleton list because this is the [method](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/xray/model/BatchGetTracesRequest.html#:~:text=was%20not%20specified.-,traceIds,-public%20final%C2%A0) required to get the segments. It takes in a list of traceIds.

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
humivo: Please refer to other comment about the list of traceIds

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
humivo: I've added what it would look like to read from a `.json` file for the input to the first testcase and it still does not look the best.  I found out that even reading in `.json` is not much better because the value of the `document` key needs to be a string and `.json` does not allow for multi-line strings as well.  This means we still have a long escape json string.  I'm not sure if there are any other alternatives to this.  What do we think?

validator/src/test/java/com/amazon/aoc/testdata/LoadBalancingValidatorTraceSuccess.json
bryan-aguilar, line 1: Please format this with a json prettier.

validator/src/test/java/com/amazon/aoc/validators/LoadBalancingValidatorTest.java
```diff
@@ -47,26 +48,35 @@ public void testValidationSucceeds() throws Exception {
...
+  /** test validation when it fails due to different collector ids */
   @Test
   public void testValidationFailedExpectedMatchingId() throws Exception {
     ValidationConfig validationConfig = new ValidationConfig();
     validationConfig.setCallingType("http");
     Context context = initContext();
 
     String mockedTrace =
-        "[{ \"segments\": [{ \"document\": '{ \"metadata\": { \"default\": {\"collector-id\": 3} }, \"subsegments\": [{ \"metadata\": { \"default\": {\"collector-id\": 2} }}, { \"metadata\": { \"default\": {\"collector-id\": 3} }}] }' }] }]";
+        "[{ \"segments\": [{ \"document\": '{ \"metadata\": { \"default\": {\"collector-id\": 2} }, \"subsegments\": [{ \"metadata\": { \"default\": {\"collector-id\": 3} }}, { \"metadata\": { \"default\": {\"collector-id\": 3} }}] }' }] }]";
```
bryan-aguilar: Shouldn't this be stored in it's own file also?

validator/src/test/java/com/amazon/aoc/validators/LoadBalancingValidatorTest.java
```diff
@@ -47,26 +48,35 @@ public void testValidationSucceeds() throws Exception {
     validationConfig.setCallingType("http");
     Context context = initContext();
 
-    String mockedTrace =
-        "[{ \"segments\": [{ \"document\": '{ \"metadata\": { \"default\": {\"collector-id\": 3} }, \"subsegments\": [{ \"metadata\": { \"default\": {\"collector-id\": 3} }}, { \"metadata\": { \"default\": {\"collector-id\": 3} }}] }' }] }]";
+    ObjectMapper mapper = new ObjectMapper();
+    JsonNode jsonNode =
+        mapper.readValue(
+            new File(
+                "./src/test/java/com/amazon/aoc/testdata/LoadBalancingValidatorTraceSuccess.json"),
```
bryan-aguilar: Can we use `getClass().getResource()` for loading this file? We do similar file loads in other parts of the validator. https://github.com/aws-observability/aws-otel-test-framework/blob/be90b42b6fbe423a59c1d3fc7df208701ff7d70a/validator/src/main/java/com/amazon/aoc/fileconfigs/PredefinedExpectedTemplate.java#L86

validator/src/test/java/com/amazon/aoc/validators/LoadBalancingValidatorTest.java
```diff
@@ -47,26 +48,35 @@ public void testValidationSucceeds() throws Exception {
...
+            JsonNode.class);
+    String mockedTrace = mapper.writeValueAsString(jsonNode);
 
-    ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+    // String mockedTrace = "[{ \"segments\": [{ \"document\": '{ \"metadata\": { \"default\":
+    // {\"collector-id\": 3} }, \"subsegments\": [{ \"metadata\": { \"default\": {\"collector-id\":
+    // 3} }}, { \"metadata\": { \"default\": {\"collector-id\": 3} }}] }' }] }]";
```
bryan-aguilar: Remove dead code

validator/src/main/java/com/amazon/aoc/validators/XrayValidator.java
```diff
@@ -64,63 +54,11 @@ public void init(
...
-      log.info("Total number of successful runs: {}", successes);
-    }
-
-    log.info("validation is passed for path {}", caller.getCallingPath());
+    this.expectedTrace = expectedTrace;
   }
 
   // this method will hit get trace from x-ray service and get retrieved trace
-  private Map<String, Object> getRetrievedTrace(List<String> traceIdList) throws Exception {
+  public Map<String, Object> getRetrievedTrace(List<String> traceIdList) throws Exception {
```
bryan-aguilar: This whole method throws me off a bit. It is called `getRetrievedTrace` (singular) but takes in a list. It then proceeds to use the xrayService to query for a list of traces by id. Then it just blindly grabs all the segments from the first trace.

Shouldn't this method take in a single traceId? 

Is `xrayService.listTraceByIds()` used somewhere else where a List<Trace> data structure is actually needed?

bryan-aguilar: I think the weird ergonomics of this method force us to write not great code in the classes that extend this abstract class. IMO, it would be cleaner to have two methods `getRetrivedTrace(traceId)` and a `getRetrievedTraces(List<String> traceIdList).

bryan-aguilar: Nit: I don't love the word `retrieved` here. Personally, it took me some time to get use to it when diving into the Validator. The validator is basically a big testing application. It does assertions against actual versus expected trace ids and data fields. I think `actual` would be more suiting....but that's just like my opinion man.

validator/src/main/java/com/amazon/aoc/validators/LoadBalancingValidator.java
```diff
@@ -0,0 +1,111 @@
...
+            Set<String> set =
+                retrievedTrace.keySet().stream()
+                    .filter(s -> s.endsWith("collector-id"))
+                    .collect(Collectors.toSet());
+            if (!checkSpanCount(set)) {
+              throw new BaseException(ExceptionCode.NOT_ENOUGH_SPANS);
+            }
+            String targetCollectorId =
+                retrievedTrace.get(set.stream().findFirst().get()).toString();
+            set.remove(set.stream().findFirst().get());
```
bryan-aguilar: Nit: You don't actually need to remove this do you?

validator/src/main/java/com/amazon/aoc/validators/LoadBalancingValidator.java
```diff
@@ -0,0 +1,111 @@
...
+      // Retry 5 times to since segments might not be immediately available in X-Ray service
+      RetryHelper.retry(
+          5,
+          Integer.parseInt(GenericConstants.SLEEP_IN_MILLISECONDS.getVal()),
+          true,
+          () -> {
+            // get retrieved trace from x-ray service
+            Map<String, Object> retrievedTrace = this.getRetrievedTrace(traceIdList);
+            log.info("value of retrieved trace map: {}", retrievedTrace);
+            Set<String> set =
```
bryan-aguilar: Why a set for this?

bryan-aguilar: Can we choose a better and more descriptive name for this?

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
bryan-aguilar: Can you add this as a comment?

validator/src/main/java/com/amazon/aoc/validators/LoadBalanceValidator.java
bryan-aguilar: If it passed then it had to have 5 successes. That should be known.

validator/src/test/java/com/amazon/aoc/validators/LoadBalanceValidatorTest.java
bryan-aguilar: We already use mustache files heavily in the validator. Can we reuse and extend the mustache rendering functionality for expected testcase files?

bryan-aguilar: The only difference is that we wouldn't need to inject any data into the mustache files. Unless you find a need to inject data during rendering.

humivo: I now switched to rendering mustache files and it works.

validator/src/test/java/com/amazon/aoc/validators/LoadBalancingValidatorTest.java
humivo: Changed in my newest commit.

validator/src/test/java/com/amazon/aoc/validators/LoadBalancingValidatorTest.java
humivo: I've opted to use `File().getAbsolutePath` still as getClass().getResource(path) returns a URL object but we are looking to pass in a String path to `LocalPathExpectedTemplate` for rendering the .mustache file.

validator/src/test/java/com/amazon/aoc/validators/LoadBalancingValidatorTest.java
humivo: Changed in my newest commit.

validator/src/main/java/com/amazon/aoc/validators/XrayValidator.java
humivo: I believe it was a list of traceIds due to this [method](https://docs.aws.amazon.com/xray/latest/api/API_BatchGetTraces.html) which is the only way to get segments from a trace and only takes in a list of traceIds.  For readability, I have added a method to our XrayService to get the segments from a single traceId and so I can add another method to our XrayValidator called `getActualTrace()` which is less confusing and prevents us from being constricted to providing a list input.

validator/src/main/java/com/amazon/aoc/validators/LoadBalancingValidator.java
humivo: I have changed this to a list and no longer need to remove it.

validator/src/main/java/com/amazon/aoc/validators/LoadBalancingValidator.java
humivo: I have changed this to a list now.

humivo: I have changed the name to `collectorIdList`

bryan-aguilar (approved):

PR #1322, opened 2023-07-14
Add ability to fetch expected values from k8s pods

bryan-aguilar:
**Description:** This PR adds functionality which allows the validator to fetch metadata values from a pod. This will allow assertions to be made against pod metadata. 

The validator uses the same kubecfg file that the terraform test uses to authenticate to the cluster. The kubecontext factory expects only one sample-app pod to be running in the namespace. 


validator/src/main/java/com/amazon/aoc/App.java
```diff
@@ -104,6 +106,21 @@ public class App implements Callable<Integer> {
       defaultValue = "true")
   private boolean isRollup;
 
+  @CommandLine.Option(
+      names = {"--kubeCfgFilePath"},
```
rapphil: super nit: all the rest of the parameters don't use camel case. can we keep the consistency?

validator/src/main/java/com/amazon/aoc/models/kubernetes/KubernetesContextFactory.java
```diff
@@ -0,0 +1,59 @@
...
+
+  public KubernetesContextFactory(
+      String kubeConfigFilePath, String deploymentName, String namespace) throws Exception {
+    this.kubeConfigFilePath = kubeConfigFilePath;
+    this.deploymentName = deploymentName;
+    this.namespace = namespace;
+    this.kubernetesService = new KubernetesService(this.kubeConfigFilePath);
+  }
+
+  public KubernetesContextFactory(
```
rapphil: nit: since the test class is in the[ same package ](https://github.com/aws-observability/aws-otel-test-framework/pull/1322/files#diff-5f739c96e0ac9d4a16c6a1bc6fceb2b4f2589bbaa730864035f2cc42bec791ebR1)as this one and this constructor is only intended to be used for tests, you can remove the `public` visibility modifier so that this constructor can only be visible in the tests.

validator/src/test/java/com/amazon/aoc/models/kubernetes/KubernetesContextFactoryTest.java
```diff
@@ -0,0 +1,47 @@
...
+                    .uid(mockUid))
+            .spec(new V1PodSpec().nodeName(mockNodeName));
+    KubernetesService mockKubernetesService = mock(KubernetesService.class);
+    when(mockKubernetesService.getSampleAppPod(mockDeploymentName, mockNamespace))
+        .thenReturn(mockPod);
+    KubernetesContextFactory mockFactory =
+        new KubernetesContextFactory(
+            "./kubecfg", mockDeploymentName, mockNamespace, mockKubernetesService);
+
+    MockedStatic<Kubectl> staticMock = mockStatic(Kubectl.class);
```
rapphil: do you still need this?

rapphil (approved):

validator/src/test/java/com/amazon/aoc/models/kubernetes/KubernetesContextFactoryTest.java
bryan-aguilar: no, that was a miss from a previous implementation.

PR #1353, opened 2023-07-26
[dev] Add additional safety checks for creating KubernetesContext

bryan-aguilar:
**Description:** This PR adds additional logic for when the validator attempts to build the kubernetesContext object. The object should only be created if a kubecfg file path is given.

The test framework validator docker compose was also adjusted. The volume source path must always be given but the flag value given to the validator requires a conditional check. 

**Testing:** Ran ec2 otlp_metric and eks otlp_metric_k8sattr tests.

### DIFF
--- a/validator/src/test/java/com/amazon/aoc/models/kubernetes/KubernetesContextFactoryTest.java
+++ b/validator/src/test/java/com/amazon/aoc/models/kubernetes/KubernetesContextFactoryTest.java
@@ -1,6 +1,7 @@
 package com.amazon.aoc.models.kubernetes;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 import static org.mockito.Mockito.*;
 
 import com.amazon.aoc.services.KubernetesService;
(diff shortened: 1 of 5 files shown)


validator/src/test/java/com/amazon/aoc/models/kubernetes/KubernetesContextFactoryTest.java
```diff
@@ -11,6 +12,13 @@
 
 public class KubernetesContextFactoryTest {
 
+  @Test
+  public void testCreateContextEmptyStrings() throws Exception {
+    KubernetesContextFactory mockFactory = new KubernetesContextFactory("", "", "");
+    KubernetesContext actualKubernetesContext = mockFactory.create();
+    assertNull(actualKubernetesContext);
+  }
```
rapphil, line 20: If you have this test here, why are you not instantiating the factory so that `create` returns `null` [here](https://github.com/aws-observability/aws-otel-test-framework/pull/1353/files#diff-d564e22ddf81aaa52b4b0ab7244be15864537545edb3538499eba3f27bebf03dR169).

bryan-aguilar: Good point, I wasn't sure if this was the correct behavior when I was creating this PR. I have moved all the empty checks into `create()`.

rapphil (approved):

PR #1356, opened 2023-07-27
Add configurable sleep duration for load balancing tests

bryan-aguilar:
**Description:** While building locally I noticed that the validator was taking 2+ minutes to complete the load balancing validator tests. This is due to the sleep duration constant that was used for its retry behavior. I have made changes to the validator to allow it to be configured in the `init` method used by the tests. This speeds up validator build time dramatically which will translate to faster collector CI run times.  

I have also changed a method signature which would never throw an exception. 

### DIFF
--- a/validator/src/main/java/com/amazon/aoc/validators/LoadBalancingValidator.java
+++ b/validator/src/main/java/com/amazon/aoc/validators/LoadBalancingValidator.java
@@ -102,8 +102,8 @@ public class LoadBalancingValidator extends XrayValidator {
     return sampleAppResponse.getTraceId();
   }
 
-  private boolean checkSpanCount(List<String> spanSet) {
-    if (spanSet.size() < 2) {
+  private boolean checkSpanCount(List<String> spanList) {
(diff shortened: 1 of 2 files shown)


validator/src/main/java/com/amazon/aoc/validators/LoadBalancingValidator.java
```diff
@@ -97,7 +102,7 @@ private String getSampleAppResponse() throws Exception {
     return sampleAppResponse.getTraceId();
   }
 
-  private boolean checkSpanCount(List<String> spanSet) throws Exception {
+  private boolean checkSpanCount(List<String> spanSet) {
```
humivo: Now that I changed the collection of spans from a set to a list, do you mind changing this variable name to something else like `spanList`?

bryan-aguilar: done

humivo (approved):

PR #1369, opened 2023-08-02
[chore] Add ignore for package com.github.johnrengelman.shadow

vasireddy99:
**Description:** 
Add ignore for package com.github.johnrengelman.shadow during the dependabot run to avoid major version updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -71,7 +71,9 @@ updates:
         update-types:
           - "version-update:semver-major"
           - "version-update:semver-minor"
-
+      - dependency-name: "com.github.johnrengelman.shadow"
+          update-types:
+            - "version-update:semver-minor"
 
   - package-ecosystem: "gradle"
     # build.gradle stored in `/load-generator` directory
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -71,7 +71,9 @@ updates:
         update-types:
           - "version-update:semver-major"
           - "version-update:semver-minor"
-
+      - dependency-name: "com.github.johnrengelman.shadow"
+          update-types:
```
bryan-aguilar: The name is a bit confusing but `update-types` is a list of update types to ignore. So we would want `version-update:semver-major` here. See dependabot docs [here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore).

bryan-aguilar: Also, this tabbing does not look correct.

.github/dependabot.yml
```diff
@@ -71,7 +71,9 @@ updates:
...
+          - "version-update:semver-major"
```
bryan-aguilar, line 76: does this need to be indented or is it already? I can't tell

vasireddy99: Oh yes yes, it is indented in my IDE

image

vasireddy99: Oh yes yes, it is indented in my IDE. double checked and recommited

image

bryan-aguilar (approved):

PR #1425, opened 2023-10-03
Use yum to install docker instead of amazon-linux-extra

Kausik-A:
**Description:** 

- Use `yum` to install `docker` instead of `amazon-linux-extra` as Amazon linux 2023 does not have `amazon-linux-extra` included anymore. This was causing the installation failure of docker
- Change symbolic link from `libcrypt.so` to `libcrypt.so.1` to fix this workflow [failure](https://github.com/aws-observability/aws-otel-collector/actions/runs/6328429022/job/17188255940#step:9:820)
`Error loading Python lib '/tmp/_MEIaR70C0/libpython3.7m.so.1.0': dlopen: libcrypt.so.1: cannot open shared object file: No such file or directory`

### DIFF
--- a/terraform/ec2/main.tf
+++ b/terraform/ec2/main.tf
@@ -405,7 +405,7 @@ resource "null_resource" "setup_sample_app_and_mock_server" {
   provisioner "remote-exec" {
     inline = [
       "sudo yum update -y",
-      "sudo amazon-linux-extras install docker -y",
+      "sudo yum install -y docker",
       "sudo service docker start",
       "sudo usermod -a -G docker ec2-user",
       "sudo curl -L 'https://github.com/docker/compose/releases/download/1.27.4/docker-compose-Linux-x86_64' -o /usr/local/bin/docker-compose",
--- a/terraform/ec2/main.tf
+++ b/terraform/ec2/main.tf
@@ -405,6 +405,7 @@ resource "null_resource" "setup_sample_app_and_mock_server" {
   provisioner "remote-exec" {
     inline = [
       "sudo yum update -y",
+      "sudo yum install -y libxcrypt-compat",
       "sudo yum install -y docker",
       "sudo service docker start",
       "sudo usermod -a -G docker ec2-user",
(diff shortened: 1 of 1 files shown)


terraform/ec2/main.tf
```diff
@@ -405,7 +405,7 @@ resource "null_resource" "setup_sample_app_and_mock_server" {
   provisioner "remote-exec" {
     inline = [
       "sudo yum update -y",
-      "sudo amazon-linux-extras install docker -y",
+      "sudo yum install -y docker",
```
rapphil, line 409: will this work for any AMI?

Kausik-A: The sample app and mock server is only launched on AL2023 instance. Previously it was on AL2022. So this should be fine

Kausik-A:
Added another commit to fix `libcrypt` issue. Updated the issue description with more info.

terraform/ec2/main.tf
bryan-aguilar: Are these issues only present for the sample and mocked server host? Why not just use AL2 for sample app and mocked server then?

bryan-aguilar: We only care about testing the collector on AL3.

Kausik-A: True but AL2 will reach End of Life on 2025-06-30 so we have to upgrade then to AL3. So wanted to tackle 2 birds with 1 stone.

terraform/ec2/main.tf
```diff
@@ -405,9 +405,10 @@ resource "null_resource" "setup_sample_app_and_mock_server" {
...
       "sudo service docker start",
       "sudo usermod -a -G docker ec2-user",
+      "sudo ln -s /usr/lib/libcrypt.so /usr/lib/libcrypt.so.1",
```
vasireddy99, line 412: Should we [install](https://github.com/lando/lando/issues/3326#issuecomment-1625339749) instead of symbolic link?

vasireddy99 (approved):
referring to packages [removed in Amazon linux 2023](https://docs.aws.amazon.com/linux/al2023/release-notes/removed-AL2023.2-AL2.html)

terraform/ec2/main.tf
Kausik-A: I marked it as a second solution in case any issues come up for symbolic linking.

Kausik-A: Update: The referred solution is better so changed it.

PR #1426, opened 2023-10-04
Add Integration Tests for the Log components

vasireddy99:
**Description:**
This PR adds the below test cases to test the components CWLogs Exporter, Filelog Receiver.
   - log4j
   - syslog rfc5424
   - json

**Testing:** 

Local Testing


adot-testbed/app/src/test/resources/configurations/config-json.yaml
```diff
@@ -0,0 +1,21 @@
...
+    start_at: beginning
+    operators:
+      - type: json_parser
+
+exporters:
+  logging:
+    verbosity: detailed
+  awscloudwatchlogs:
+    log_group_name: "/test/logs"
+    log_stream_name: "logstream-json"
```
rapphil: can you update all the configurations to set the retention period to a small value like 7 days?

bryan-aguilar: We should also use a more descriptive log group and log stream name to make troubleshooting easier.

bryan-aguilar: Consider adding a unique identifier for the test run for each log stream.

adot-testbed/app/src/test/resources/configurations/config-json.yaml
```diff
@@ -0,0 +1,21 @@
...
+    raw_log: true
```
bryan-aguilar: Why `raw_log` set to true?

adot-testbed/app/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
...
+
+application {
+    // Define the main class for the application.
+    mainClass.set("adot.testbed.App")
+}
+
+tasks.named<Test>("test") {
+    // Use JUnit Platform for unit tests.
+    useJUnitPlatform()
+    systemProperty("adot.testbed.localcreds", System.getProperty("adot.testbed.localcreds"))
```
bryan-aguilar: What does this do?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.stream.IntStream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@Testcontainers(disabledWithoutDocker = true)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class LogsTests {
+    private static final String LOCAL_CREDENTIALS = System.getProperty("adot.testbed.localcreds");
```
bryan-aguilar: What are the local credentials for?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -0,0 +1,145 @@
...
+    private static final String TEST_IMAGE = System.getenv("TEST_IMAGE") != null && !System.getenv("TEST_IMAGE").isEmpty()
+        ? System.getenv("TEST_IMAGE")
+        : "public.ecr.aws/aws-otel-test/adot-collector-integration-test:latest";
```
bryan-aguilar: Why do we default to the latest from integration test? If test_image I think we could default to the latest production image instead.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -0,0 +1,145 @@
...
+    private final Logger collectorLogger = LoggerFactory.getLogger("collector");
+
+    private GenericContainer<?> collector;
+
+    LogsTests() throws Exception {
+    }
+
+    private GenericContainer<?> createAndStartCollector(String configFilePath, String logFilePath) throws IOException {
+        var collector = new GenericContainer<>(TEST_IMAGE)
+            .withExposedPorts(4317)
```
bryan-aguilar: Why does `4317` need to be exposed here?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -0,0 +1,145 @@
...
+            .withLogConsumer(new Slf4jLogConsumer(collectorLogger))
+            .waitingFor(Wait.forLogMessage(".*Everything is ready. Begin running and processing data.*", 1))
+            .withCommand("--config", "/etc/collector/config.yaml", "--feature-gates=+adot.filelog.receiver,+adot.awscloudwatchlogs.exporter,+adot.file_storage.extension");
+
+        //Mount the log file for the file log receiver to parse
+        collector.withCopyFileToContainer(MountableFile.forClasspathResource(logFilePath), logFilePath );
+        if (LOCAL_CREDENTIALS != null && !LOCAL_CREDENTIALS.isEmpty()) {
+            collector.withCopyFileToContainer(MountableFile.forHostPath(LOCAL_CREDENTIALS), "/root/.aws/");
+        } else {
+            collector.withEnv(System.getenv());
```
bryan-aguilar: I do not like this fallback. There is a chance secrets may be mounted in the environment and I would not like them present in the container if they do not have to be.

bryan-aguilar: Only mount the environment variables you actually need.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -0,0 +1,145 @@
...
+            .retryIfExceptionOfType(org.opentest4j.AssertionFailedError.class)
+            .withWaitStrategy(WaitStrategies.fixedWait(10, TimeUnit.SECONDS))
+            .withStopStrategy(StopStrategies.stopAfterAttempt(5))
+            .build()
+            .call(() -> {
+                var now = Instant.now();
+                var start = now.minus(Duration.ofMinutes(2));
+                var end = now.plus(Duration.ofMinutes(2));
+                var response = cwClient.getLogEvents(GetLogEventsRequest.builder().logGroupName("/test/logs")
+                    .logStreamName(logName)
```
bryan-aguilar: There is a very tight coupling in the validation between the log file name and the stream it expects it in. We should reduce this coupling or clearly document it. 

For example: `The validation logic expects logs to be in the same log stream as the file name`. 

I think this is bad though. What happens if two instances of these test are ran at the same time. Both will write to the same log stream and log group. This could cause transient failures or even worse, false successes. Each test case should write to a log stream name that is unique to it's test case.

bryan-aguilar (dismissed):
We need to find a way to catch build time failures of this new test bed at PR time. Are build time failures of this test bed caught already by existing PR workflows?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: acutally logName there meant to be log stream name. Sorry for the confusion changed the usage of variable name.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: Seems to me exposing port for the collector just as part of POC, for any sample app to send otlp. but I think we can remove it if we want to.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: Yeah, the idea is to use the test image from the CI ones if not specified, as we usually default to test image in the rest of the test framework repo. It is not necessary to be the integ test image though, even for rest of the framework.

adot-testbed/app/src/test/resources/configurations/config-json.yaml
vasireddy99: Agreed, added unique identifier

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: We can use the system property to provide credentials from the local terminal, the command would be something like `./gradlew -Dadot.testbed.localcreds="/Users/user/.aws" test --rerun-tasks --info`.  [at this point](https://github.com/aws-observability/aws-otel-test-framework/pull/1426/files#diff-b6b213810cf9bd4029972a83c70ccf32f71e4f4eb1e32cc5b6b90dd5bcf7961aR77) it would mount similar to `~/.aws:/root/.aws`.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: We would need the credentials to mount into the container for use in the collector. The idea here is if the system property was not set, then collector will get creds from the [local env like here](https://github.com/aws-observability/aws-otel-test-framework/pull/1426/files#diff-c6a9ad1ffe59a3efeb6ff938e2ce60dfe900c717f4cd85c1eab3f6a71a530eedR7-R10). Yes I will look further to only get required env - (AWS_REGION, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN) if that is the suggestion.

adot-testbed/app/src/test/resources/configurations/config-json.yaml
vasireddy99: In Terms of validation purpose. It makes it easy as it would just export as is that is parsed from log file and also have one log entry per line .

.github/workflows/pr-build.yml
```diff
@@ -68,8 +68,23 @@ jobs:
...
+      - name: Checkout repo
+        uses: actions/checkout@v4
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: '17'
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+      - name: adot-testbed build
```
rapphil, line 88: you need to setup the credentials otherwise the testbed will not be able to access cw logs.

adot-testbed/app/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
+/*
+ * This file was generated by the Gradle 'init' task.
+ *
+ * This generated file contains a sample Java application project to get you started.
+ * For more details on building Java & JVM projects, please refer to https://docs.gradle.org/8.2/userguide/building_java_projects.html in the Gradle documentation.
+ */
+
+plugins {
+    // Apply the application plugin to add support for building a CLI application in Java.
+    application
```
rapphil: can you switch to type library so that we don't need to define an application section with a main class?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -0,0 +1,150 @@
...
+    private GenericContainer<?> createAndStartCollector(String configFilePath, String logFilePath, String uniqueLogStreamName) throws IOException {
+
+        // Create an environment variable map
+        Map<String, String> envVariables = new HashMap<>();
+        envVariables.put("LOG_STREAM_NAME", uniqueLogStreamName);
+        //Set credentials
+        envVariables.put("AWS_REGION", System.getenv("AWS_REGION"));
+        envVariables.put("AWS_ACCESS_KEY_ID", System.getenv("AWS_ACCESS_KEY_ID"));
+        envVariables.put("AWS_SECRET_ACCESS_KEY", System.getenv("AWS_SECRET_ACCESS_KEY"));
+        envVariables.put("AWS_SESSION_TOKEN", System.getenv("AWS_SESSION_TOKEN"));
```
rapphil: session token is optional. so you should only inject this variable if it is not null.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
rapphil: yes, lets remove it if it is not being used.

rapphil (changes requested):
marking as blocked because of the lack of credentials setup in the workflow.

.github/workflows/pr-build.yml
```diff
@@ -68,8 +68,23 @@ jobs:
...
+        working-directory: adot-testbed/
+        run: gradle build -x test
```
rapphil, line 90: is the intent to run the testbed tests here?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
rapphil: the default image should be the prod image if we want to run the tests in this repository.

.github/workflows/pr-build.yml
vasireddy99: In this PR build I am excluding the test part. I think we wont be needing creds for the build part. just to catch for any build failures

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: Agreed

.github/workflows/pr-build.yml
rapphil: makes sense.

.github/workflows/pr-build.yml
rapphil: resolving as we will not run the test

.github/workflows/pr-build.yml
bryan-aguilar: This is an interesting idea though. Should we add a workflow in this repository to run the tests against the latest version of the collector? It would act as an integration test. I wish we could do it at PR time but I wouldn't love introducing credentials into the PR workflow.

vasireddy99:
> We need to find a way to catch build time failures of this new test bed at PR time. Are build time failures of this test bed caught already by existing PR workflows?

Added PR build check for the build purpose

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: Agreed

adot-testbed/app/build.gradle.kts
vasireddy99: Agreed

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: Yes, removed the port for now

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
rapphil: you forgot to remove line [64](https://github.com/aws-observability/aws-otel-test-framework/pull/1426/files#diff-b6b213810cf9bd4029972a83c70ccf32f71e4f4eb1e32cc5b6b90dd5bcf7961aR64)  since you are optionally injecting the `AWS_SESSION_TOKEN` on line 66.

rapphil (approved):

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -0,0 +1,152 @@
...
+        : "public.ecr.aws/aws-observability/aws-otel-collector:latest";
+    private final Logger collectorLogger = LoggerFactory.getLogger("collector");
+    private static final String uniqueID = UUID.randomUUID().toString();
+
+    private GenericContainer<?> collector;
+
+    LogsTests() throws Exception {
+    }
+
+    private GenericContainer<?> createAndStartCollector(String configFilePath, String logFilePath, String uniqueLogStreamName) throws IOException {
```
bryan-aguilar: Consider renaming `uniqueLogStreamName`. Log streams must be unique anyways as a [requirement](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_CreateLogStream.html). This also doesn't provide any context on why you would want to use unique log streams for these tests. `createAndStartCollector` is and IMO should be generic. Log stream uniqueness for the test case is a constraint for our tests to execute properly.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -0,0 +1,152 @@
...
+    private static final String TEST_IMAGE = System.getenv("TEST_IMAGE") != null && !System.getenv("TEST_IMAGE").isEmpty()
+        ? System.getenv("TEST_IMAGE")
+        : "public.ecr.aws/aws-observability/aws-otel-collector:latest";
+    private final Logger collectorLogger = LoggerFactory.getLogger("collector");
+    private static final String uniqueID = UUID.randomUUID().toString();
+
+    private GenericContainer<?> collector;
+
+    LogsTests() throws Exception {
+    }
```
bryan-aguilar: Is this required? I am not Java expert, apologies.

adot-testbed/app/src/test/resources/configurations/config-log4j.yaml
```diff
@@ -0,0 +1,19 @@
+receivers:
+  filelog:
+    include: [/logs/log4j.log]
+    start_at: beginning
+
+exporters:
+  logging:
+    verbosity: detailed
```
bryan-aguilar: Please remove this logging exporter. It will just create noise in the CI. Viewing the CW Logs for this should be easy enough where we don't have to rely on the logging exporter.

vasireddy99: Yes, more of this existence is to help in local testing to observe whether the logs exported or not. Removed from pipeline.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: Renamed the usage

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: It is not required, yes. removed it .

rapphil (approved):

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -0,0 +1,169 @@
...
+                            return jsonNode.get("body").asText();
+                        } catch (Exception e) {
+                            return null;
+                        }
+                    })
+                    .filter(Objects::nonNull)
+                    .collect(Collectors.toSet());
+
+                //Validate body field in JSON-messageToValidate with actual log line from the log file.
+                assertThat(messageToValidate.containsAll(lines)).isTrue();
```
bryan-aguilar, line 164: Do we need to do an inverse assertion also? 
```suggestion
                assertThat(messageToValidate.containsAll(lines)).isTrue();
                assertThat(lines.containsAll(messageToValidate)).isTrue();
```

bryan-aguilar: We could also validate that their size is the same. This would ensure that no extra logs were went. We expect 1:1 correct?

bryan-aguilar (approved):

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: yes that's correct it can also be like - assertThat(receivedMessages).containsExactlyInAnyOrderElementsOf(lines); added assertion.

PR #1427, opened 2023-10-06
Add Tests for Storage Extension and File Rotation

vasireddy99:
**Description:** 
PR to add Tests for Storage Extension Restart scenarios and File Rotation

Thread.sleep is used to give enough time for file operations, collector(stop/start).

### DIFF
--- a/adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
+++ b/adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
@@ -100,7 +100,7 @@ class LogsTests {
 
         List<String> logFilePaths = new ArrayList<>();
         logFilePaths.add("/logs/RFC5424.log");
-        validateLogs(logStreamName , logFilePaths, true);
+        validateLogs(logStreamName , logFilePaths);
         collector.stop();
     }
 
@@ -111,7 +111,7 @@ class LogsTests {
 
         List<String> logFilePaths = new ArrayList<>();
         logFilePaths.add("/logs/log4j.log");
-        validateLogs(logStreamName , logFilePaths, true);
+        validateLogs(logStreamName , logFilePaths);
         collector.stop();
     }
 
@@ -122,7 +122,7 @@ class LogsTests {
 
         List<String> logFilePaths = new ArrayList<>();
         logFilePaths.add("/logs/testingJSON.log");
-        validateLogs(logStreamName , logFilePaths, true);
+        validateLogs(logStreamName , logFilePaths);
         collector.stop();
     }
 
@@ -147,7 +147,7 @@ class LogsTests {
         String expectedLogPath = logDirectory.toString();
         logFilePaths.add(expectedLogPath + "/storageExtension.log");
 
-        validateLogs(logStreamName , logFilePaths, false);
+        validateLogs(logStreamName , logFilePaths);
 
         collector.stop();
 
@@ -166,7 +166,7 @@ class LogsTests {
        //restarting the collector
         collector.start();
 
-        validateLogs(logStreamName , logFilePaths, false);
+        validateLogs(logStreamName , logFilePaths);
         collector.stop();
     }
 
@@ -208,19 +208,19 @@ class LogsTests {
         logFilePaths.add(expectedLogPath + "/testlogA-1234.log");
         logFilePaths.add(expectedLogPath + "/testlogA.log");
 
-        validateLogs(logStreamName, logFilePaths, false);
+        validateLogs(logStreamName, logFilePaths);
 
         collector.stop();
     }
 
 
-    void validateLogs(String testLogStreamName, List<String> logFilePaths, boolean areResourceFiles) throws Exception {
+    void validateLogs(String testLogStreamName, List<String> logFilePaths) throws Exception {
         var lines = new HashSet<String>();
 
         for (String logFilePath : logFilePaths) {
             InputStream inputStream;
             //Check whether the filePath is from resource folder.
-            if (areResourceFiles) {
+            if (getClass().getResource(logFilePath) != null) {
                 inputStream = getClass().getResourceAsStream(logFilePath);
             } else {
                 inputStream = new FileInputStream(logFilePath);
--- a/adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
+++ b/adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
@@ -16,7 +16,14 @@ import java.nio.file.Files;
 import java.net.URISyntaxException;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.*;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Objects;
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
@@ -50,10 +57,10 @@ class LogsTests {
         : "public.ecr.aws/aws-observability/aws-otel-collector:latest";
     private final Logger collectorLogger = LoggerFactory.getLogger("collector");
     private static final String uniqueID = UUID.randomUUID().toString();
-    private final Path logDirectory = Files.createTempDirectory("filerotationlogs");
+    private final Path logDirectory = Files.createTempDirectory("tempLogs");
     private GenericContainer<?> collector;
 
-    private GenericContainer<?> createAndStartCollector(String configFilePath, String logFilePath, String logStreamName) throws IOException {
+    private GenericContainer<?> createAndStartCollector(String configFilePath, String logStreamName) throws IOException {
 
         // Create an environment variable map
         Map<String, String> envVariables = new HashMap<>();
@@ -75,19 +82,13 @@ class LogsTests {
             .withEnv(envVariables)
             .withClasspathResourceMapping("/logs", "/logs", BindMode.READ_WRITE)
 
-            .withPrivilegedMode(true)
+//            .withPrivilegedMode(true)
             .withCommand("--config", "/etc/collector/config.yaml", "--feature-gates=+adot.filelog.receiver,+adot.awscloudwatchlogs.exporter,+adot.file_storage.extension");
 
-        //Mount the log file for the file log receiver to parse
-        collector.withCopyFileToContainer(MountableFile.forClasspathResource(logFilePath), logFilePath);
-        collector.withFileSystemBind(logDirectory.toString(),"/filerotation", BindMode.READ_WRITE);
-//        collector.withFileSystemBind(tempFile.getAbsolutePath(), "/filerotation", BindMode.READ_WRITE);
-//        collector.withFileSystemBind(tempFileB.getAbsolutePath(), "/logs/filerotationlogs/logsB.log", BindMode.READ_WRITE);
-//        collector.withCopyFileToContainer(MountableFile.forClasspathResource("/logs/"), "/logs/",  );
-//        collector.withCommand("sh", "-c", "chmod -R a+rw " + "/logs");
+       //Mount the Temp directory
+        collector.withFileSystemBind(logDirectory.toString(),"/tempLogs", BindMode.READ_WRITE);
 
         collector.start();
-//        collector.waitingFor(Wait.forHealthcheck());
         return collector;
     }
 
@@ -96,7 +97,7 @@ class LogsTests {
     @Test
     void testSyslog() throws Exception {
         String logStreamName = "rfcsyslog-logstream-" + uniqueID;
-        collector = createAndStartCollector("/configurations/config-rfcsyslog.yaml", "/logs/RFC5424.log", logStreamName);
+        collector = createAndStartCollector("/configurations/config-rfcsyslog.yaml", logStreamName);
 
         List<String> logFilePaths = new ArrayList<>();
         logFilePaths.add("/logs/RFC5424.log");
@@ -107,7 +108,7 @@ class LogsTests {
     @Test
     void testLog4j() throws Exception {
         String logStreamName = "log4j-logstream-" + uniqueID;
-        collector = createAndStartCollector("/configurations/config-log4j.yaml", "/logs/log4j.log", logStreamName);
+        collector = createAndStartCollector("/configurations/config-log4j.yaml", logStreamName);
 
         List<String> logFilePaths = new ArrayList<>();
         logFilePaths.add("/logs/log4j.log");
@@ -118,7 +119,7 @@ class LogsTests {
     @Test
     void testJson() throws Exception {
         String logStreamName = "json-logstream-" + uniqueID;
-        collector = createAndStartCollector("/configurations/config-json.yaml", "/logs/testingJSON.log", logStreamName);
+        collector = createAndStartCollector("/configurations/config-json.yaml", logStreamName);
 
         List<String> logFilePaths = new ArrayList<>();
         logFilePaths.add("/logs/testingJSON.log");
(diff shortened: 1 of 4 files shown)


adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -65,61 +73,167 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
-    void validateLogs(String testLogStreamName, String logFilePath) throws Exception {
-        var file = new File(logFilePath);
+    @Test
+    void testCollectorRestartStorageExtension() throws Exception {
+        String logStreamName = "storageExtension-logstream-" + uniqueID;
+        collector = createAndStartCollector("/configurations/config-storageExtension.yaml", logStreamName);
+        File tempFile = new File(logDirectory.toString(), "storageExtension.log");
+        Thread.sleep(5000);
+
+        FileWriter fileWriter = new FileWriter(tempFile);
```
rapphil: nit: with [PrintWriter](https://docs.oracle.com/javase/8/docs/api/java/io/PrintWriter.html) you will be able to use the `println` method and will not need to append line break `\n`.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -65,61 +73,167 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
+        try {
+            fileWriter.write("First Message, collector is running" + "\n");
+            fileWriter.write("Second Message, collector is running" + "\n");
+            fileWriter.write("Third Message,  collector is running" + "\n");
+        } catch (IOException e) {
```
rapphil: do you need this since your method has a `throws Exception`? to simplify the code, I think you don't need to catch this exception here.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -65,61 +73,167 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
+
+        List<String> logFilePaths = new ArrayList<>();
+        String expectedLogPath = logDirectory.toString();
+        logFilePaths.add(expectedLogPath + "/storageExtension.log");
+
+        validateLogs(logStreamName , logFilePaths, false);
+
+        collector.stop();
+
+        Thread.sleep(5000);
```
rapphil: why do you need this sleep? `collector.stop()` will block until the container is fully stopped and `validateLogs` will guarantee that you got the content that is expected.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -65,61 +73,167 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
+        validateLogs(logStreamName , logFilePaths, false);
+        collector.stop();
+    }
+
+    @Test
+    void testFileRotation() throws Exception {
+        String logStreamName = "fileRotation-logstream-" + uniqueID;
+        collector = createAndStartCollector("/configurations/config-fileRotation.yaml", logStreamName);
+
+        Thread.sleep(5000);
```
rapphil, line 185: why this sleep? isent' `createAndStartCollector` going to wait for the collector to be fully started?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -65,61 +73,167 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
+        // Create and write data to File A
+        File tempFile = new File(logDirectory.toString(), "testlogA.log");
+        try (FileWriter fileWriter = new FileWriter(tempFile)) {
+            fileWriter.write("Message in File A" + "\n");
+            fileWriter.flush();
+            fileWriter.close();
+        } catch (IOException e) {
+            throw new RuntimeException("Error writing to File A.", e);
+        }
+        Thread.sleep(5000);
```
rapphil: you van validate the log entries so far here so that you don't need to add a sleep. that will block in a safe way before you move with the rest of the tests.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -65,61 +73,167 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
+        try (FileWriter newfileWriter = new FileWriter(tempFileB)) {
+            newfileWriter.write("Message in renamed file - line 1" + "\n");
+            newfileWriter.write("Message in renamed file - line 2" + "\n");
+            newfileWriter.write("Message in renamed file - line 3" + "\n");
+            newfileWriter.flush();
+            newfileWriter.close();
+        } catch (IOException e) {
+            throw new RuntimeException("Error writing to File B.", e);
+        }
+        Thread.sleep(5000);
```
rapphil: again, I don't see a reason for sleeping here

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -65,61 +73,167 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
 
-        try (InputStream inputStream = getClass().getResourceAsStream(logFilePath);
-             BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream))) {
-            String line;
-            while ((line = reader.readLine()) != null) {
-                lines.add(line);
+        for (String logFilePath : logFilePaths) {
+            InputStream inputStream;
+            //Check whether the filePath is from resource folder.
+            if (areResourceFiles) {
```
rapphil: https://luzkan.github.io/smells/flag-argument

rapphil: why don't you always pass as parameter to this method a list of InputStreams?

adot-testbed/app/src/test/resources/configurations/config-fileRotation.yaml
```diff
@@ -0,0 +1,18 @@
+receivers:
+  filelog:
+    include: [/tempLogs/testlogA.log]
+    start_at: beginning
+
+exporters:
+  logging:
+    verbosity: detailed
```
rapphil: you are not using this exporter. should you delete it?

adot-testbed/app/src/test/resources/configurations/config-fileRotation.yaml
```diff
@@ -0,0 +1,18 @@
+receivers:
+  filelog:
+    include: [/tempLogs/testlogA.log]
+    start_at: beginning
```
rapphil: why start_at: beginning?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: Yeah, the collector is waiting till start but not until the file log receiver is watching. Because in this test case here the log file to parse is created after the collector is started.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: I agree that validation can be done, but the thread here is making sure to give enough time for the file operations to be done. So even if we do the validation without the thread, since the log file is not generated it would lead FileNotFoundException.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: This thread can be removed, i was playing very safe i think in my local testing. Removed this one

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: Agreed, added a check within the method, to let it decide

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
vasireddy99: True, it is removed

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -49,10 +57,10 @@ class LogsTests {
         : "public.ecr.aws/aws-observability/aws-otel-collector:latest";
     private final Logger collectorLogger = LoggerFactory.getLogger("collector");
     private static final String uniqueID = UUID.randomUUID().toString();
-
+    private Path logDirectory;
     private GenericContainer<?> collector;
```
bryan-aguilar, line 61: How does test parallelism work with the framework we are using? Does having these in the Class scope have a potential to cause conflicts if test cases are ran in parallel?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -65,61 +73,164 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
+        validateLogs(logStreamName , logFilePaths);
+        collector.stop();
+    }
+
+    @Test
+    void testCollectorRestartStorageExtension() throws Exception {
+        String logStreamName = "storageExtension-logstream-" + uniqueID;
+        collector = createAndStartCollector("/configurations/config-storageExtension.yaml", logStreamName);
+        File tempFile = new File(logDirectory.toString(), "storageExtension.log");
+        Thread.sleep(5000);
```
bryan-aguilar, line 141: Why sleep here?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -65,61 +73,164 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
 
-        validateLogs(logStreamName , "/logs/testingJSON.log");
+        collector.stop();
+
+        // write to the file when collector is stopped
+        try {
+            fileWriter.write("First Message after collector is stopped" + "\n");
+            fileWriter.write("Second Message after the collector is stopped" + "\n");
+            fileWriter.write("Third Message after the collector is stopped" + "\n");
+            fileWriter.flush();
```
bryan-aguilar: Why is flush called inside the try block here but outside of it on line 144?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -65,61 +73,164 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
+        } catch (IOException e) {
+            throw new RuntimeException("Error writing to File A.", e);
+        }
+        fileWriter.flush();
+
+        List<String> logFilePaths = new ArrayList<>();
+        String expectedLogPath = logDirectory.toString();
+        logFilePaths.add(expectedLogPath + "/storageExtension.log");
+
+        validateLogs(logStreamName , logFilePaths);
```
bryan-aguilar: Why do we need to validate logs twice? Could we run through the `start -> write -> stop -> write -> start -> stop` cycle then validate once?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
bryan-aguilar: Is writing to the file a blocking operation? Why do you need to wait? What file operations are you referring to?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -160,6 +271,13 @@ void validateLogs(String testLogStreamName, String logFilePath) throws Exception
                     .filter(Objects::nonNull)
                     .collect(Collectors.toSet());
 
+                //Print
+                System.out.println("Actual Logs - Log lines From Cloudwatch");
+                messageToValidate.forEach(System.out::println);
+
+                System.out.println("Expected logs- Log lines from log file");
+                lines.forEach(System.out::println);
```
bryan-aguilar: Let's be more thoughtful about what we are printing to std out. We should only need to print if there is a failure. And in the case of the failure it should be descriptive to tell us what was expected and what was actually given.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
rapphil: ok, I got it for this one. it is because you are not using read from start. Makes send.

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
rapphil: what operations needs to be done?

adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -65,61 +73,164 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
 
-        try (InputStream inputStream = getClass().getResourceAsStream(logFilePath);
-             BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream))) {
-            String line;
-            while ((line = reader.readLine()) != null) {
-                lines.add(line);
+        for (String logFilePath : logFilePaths) {
+            InputStream inputStream;
+            //Check whether the filePath is from resource folder.
+            if (getClass().getResource(logFilePath) != null) {
```
rapphil: this is still the boolean flag behaviour disguised.....

my suggestion was to change the function signature to use

```    void validateLogs(String testLogStreamName, List<InputStream> inputStreams) throws Exception {```

Another option is to create two functions, one that should only be used for resources paths and another that is used only for local paths.

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java (hunk 13 above)
Comment (rapphil) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java]: validate is behaving like a sleep here. we need to make sure that logs were read before stopping the collector.

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java (hunk 1 above)
Comment (vasireddy99) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java]: I agree, modified

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java (hunk 2 above)
Comment (vasireddy99) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java]: Simplified as suggested.

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java (hunk 15 above)
Comment (vasireddy99) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java]: Oh, now I see the suggestion. Modified.

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java (hunk 14 above)
Comment (vasireddy99) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java]: Sure, I have added it for easy lookout on whats' validated. if we should only need to print during failure then the assertions will log the actual/expected when in failure, removed the print statements

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java (hunk 12 above)
Comment (vasireddy99) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java]: Actually no specific reason, modified .

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java (hunk 5 above)
Comment (vasireddy99) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java]: File operations i meant here are to create and write data to File A etc. it takes considerable time for creating and writing to file.
So without the thread.sleep the collector is not able to parse log line, since it seems the file testlogA.log didn't have the line written to it yet.

```
        java.lang.AssertionError: 
        Expecting actual:
          ["Message in renamed file - line 1",
            "Message in renamed file - line 2",
            "Message in renamed file - line 3"]
        to contain exactly in any order:
          ["Message in File A",
            "Message in renamed file - line 1",
            "Message in renamed file - line 2",
            "Message in renamed file - line 3"]
        but could not find the following elements:
          ["Message in File A"]
```

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java (hunk 10 above)
Comment (vasireddy99) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java:61]: I have tried running with `./gradlew test --rerun-tasks --info --parallel` to have optimal number of threads to use no conlicts were seen. We are also using specific path/file within the directory so parallel wouldn't be a problem IMO

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java (hunk 11 above)
Comment (vasireddy99) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java:141]: The sleep is similar as [here](https://github.com/aws-observability/aws-otel-test-framework/pull/1427#discussion_r1356420926), its just to wait for the collector to start watching the file.

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java (hunk 5 above)
Comment (rapphil) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java]: that is why I suggested that you add a `validateLogs` instead of a `Thread.sleep` to validate only "Message in File A". It will block until the lines are sent to cwlogs. This will guarantee that it is safe to continue with the tests.

Comment (vasireddy99) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java]: Committed the changes.

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java (hunk 13 above)
Comment (vasireddy99) [reply adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java]: >we need to make sure that logs were read before stopping the collector.
I tend to agree with that. After changes now it is doing - `start -> write -> validate->stop -> write -> start -> stop` and then it validates again.

File: adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java [hunk 16]
```diff
@@ -65,61 +73,175 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
...
+        File tempFile = new File(logDirectory.toString(), "storageExtension.log");
+        Thread.sleep(5000);
+
+        PrintWriter printWriter = new PrintWriter(tempFile);
+        printWriter.println("First Message, collector is running");
+        printWriter.println("Second Message, collector is running");
+        printWriter.println("Third Message, collector is running");
+        printWriter.flush();
+
+        Thread.sleep(5000);
```
rapphil: I bet this could be removed and the tests would still succeed.

rapphil (approved):

PR #1468, opened 2023-10-25
Run test container as user root

vasireddy99:
**Description:**
Run user as root for test container


adot-testbed/app/src/test/java/software/amazon/adot/testbed/LogsTests.java
```diff
@@ -96,7 +97,7 @@ private GenericContainer<?> createAndStartCollector(String configFilePath, Strin
     void testSyslog() throws Exception {
         String logStreamName = "rfcsyslog-logstream-" + uniqueID;
         collector = createAndStartCollector("/configurations/config-rfcsyslog.yaml", logStreamName);
-
+        Thread.sleep(50000);
```
bryan-aguilar: Why is this sleep required? And why is to so long? If this is really required then I would expect a comment describing its necessity.

vasireddy99: It is really not required, please ignore, removed it

PaurushGarg (approved):

PR #1471, opened 2023-10-26
Introduce AL2023 to CI tests

Kausik-A:
**Description:** 

Introduce AL2023 to CI tests. These changes were already bought to the dev branch and tested on the ADOt CI.

https://github.com/aws-observability/aws-otel-test-framework/pull/1413

### DIFF
--- a/terraform/ec2/main.tf
+++ b/terraform/ec2/main.tf
@@ -405,7 +405,7 @@ resource "null_resource" "setup_sample_app_and_mock_server" {
   provisioner "remote-exec" {
     inline = [
       "sudo yum update -y",
-      "sudo amazon-linux-extra install docker -y",
+      "sudo amazon-linux-extras install docker -y",
       "sudo service docker start",
       "sudo usermod -a -G docker ec2-user",
       "sudo curl -L 'https://github.com/docker/compose/releases/download/1.27.4/docker-compose-Linux-x86_64' -o /usr/local/bin/docker-compose",
--- a/terraform/ec2/main.tf
+++ b/terraform/ec2/main.tf
@@ -405,9 +405,11 @@ resource "null_resource" "setup_sample_app_and_mock_server" {
   provisioner "remote-exec" {
     inline = [
       "sudo yum update -y",
-      "sudo amazon-linux-extras install docker -y",
+      "sudo yum install -y libxcrypt-compat",
+      "sudo yum install -y docker",
       "sudo service docker start",
       "sudo usermod -a -G docker ec2-user",
+      "sudo ln -s /usr/lib/libcrypt.so /usr/lib/libcrypt.so.1",
       "sudo curl -L 'https://github.com/docker/compose/releases/download/1.27.4/docker-compose-Linux-x86_64' -o /usr/local/bin/docker-compose",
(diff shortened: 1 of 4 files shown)


terraform/ec2/main.tf
```diff
@@ -405,9 +405,11 @@ resource "null_resource" "setup_sample_app_and_mock_server" {
   provisioner "remote-exec" {
     inline = [
       "sudo yum update -y",
-      "sudo amazon-linux-extras install docker -y",
+      "sudo yum install -y libxcrypt-compat",
+      "sudo yum install -y docker",
       "sudo service docker start",
       "sudo usermod -a -G docker ec2-user",
+      "sudo ln -s /usr/lib/libcrypt.so /usr/lib/libcrypt.so.1",
```
bryan-aguilar: Do we need this anymore since we are not updating sample app to al2023?

terraform/ec2/variables.tf
```diff
@@ -21,7 +21,7 @@ variable "package_s3_bucket" {
 }
 
 variable "testing_ami" {
-  default = "amazonlinux2"
+  default = "amazonlinux3"
 }
```
bryan-aguilar: What is this default changing? For the sample app or the collector? Either way, I don't think we should change the default.

terraform/ec2/main.tf
Kausik-A: We dont need them but it does not change any behavior on Al2022. Skipping from using `amazon-linux-extras` will make sure the future engineer not face the same issue when they try to upgrade the sample app instances to AL2023

bryan-aguilar: This didn't even work though right? There were additional issues on top? I would prefer to just remove this if it's not necessary. You can leave it as a comment if you'd like.

terraform/ec2/variables.tf
Kausik-A: The collector. But yeah, I agree. I changed default back to AL2022

bryan-aguilar:
This is missing required changes to batch test generator. https://github.com/aws-observability/aws-otel-test-framework/blob/c869895c95b5d52b9089bff1596ca49d1bd9bda5/tools/batchTestGenerator/internal/testCaseBuilder.go#L8C2-L8C2

terraform/ec2/main.tf
Kausik-A: They worked but there were additional issues relating to docker login. Okay, I have reverted the changes.

terraform/ec2/main.tf
```diff
@@ -405,7 +405,7 @@ resource "null_resource" "setup_sample_app_and_mock_server" {
   provisioner "remote-exec" {
     inline = [
       "sudo yum update -y",
-      "sudo amazon-linux-extras install docker -y",
+      "sudo amazon-linux-extra install docker -y",
```
bryan-aguilar: So the original command was wrong? Or are they different?

Kausik-A: oops, fixed it

bryan-aguilar (approved):

vasireddy99 (approved):

PR #1472, opened 2023-10-26
Add OTLP Logs Testcase

Danielzolty:
**Description:** 
Add OTLP Logs Testcase


load-generator/bin/main/log4j2.xml
vasireddy99, line 1: We can remove this file.

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
```diff
@@ -0,0 +1,113 @@
...
+
+  protected void validateJsonSchema(String logEventMsg) throws Exception {
+    JsonNode logEventNode = mapper.readTree(logEventMsg);
+    if (schema != null) {
+      ProcessingReport report = schema.validate(JsonLoader.fromString(logEventNode.toString()));
+      if (report.isSuccess()) {
+        log.info("Report was a success");
+      } else {
+        // This will probably generate a lot of extra logs
+        // may want to log this to a different level in the future.
```
vasireddy99: We may not need this comment here i think

terraform/testcases/otlp_logs/otconfig.tpl
```diff
@@ -0,0 +1,30 @@
...
+        endpoint: 0.0.0.0:${grpc_port}
+
+processors:
+  batch:
+
+exporters:
+  logging:
+    verbosity: detailed
+  awscloudwatchlogs:
+    log_group_name: "otlp-receiver"
```
vasireddy99:
```suggestion
    log_group_name: "/aws/ecs/otlp/${testingId}/logs"
```

log group can be something like

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
```diff
@@ -0,0 +1,113 @@
...
+  @Override
+  public void init(
+      Context context,
+      ValidationConfig validationConfig,
+      ICaller caller,
+      FileConfig expectedDataTemplate)
+      throws Exception {
+    this.context = context;
+    cloudWatchService = new CloudWatchService(context.getRegion());
+    logGroupName = String.format("otlp-receiver", context.getCloudWatchContext().getClusterName());
```
vasireddy99:
```suggestion
    logGroupName = String.format(LOGGROUPPATH, context.getTestingId());
```

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
```diff
@@ -0,0 +1,113 @@
...
+import com.github.fge.jsonschema.util.JsonLoader;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.*;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class CWLogValidator implements IValidator {
+
+  protected String logStreamName = "otlp-logs";
```
vasireddy99, line 29:
```suggestion
  protected String logStreamName = "otlp-logs";
  
  private static final String LOGGROUPPATH = "/aws/ecs/otlp/%s/logs";
```

terraform/testcases/otlp_logs/otconfig.tpl
Danielzolty: Then we need to change it in CWLogValidator as well. How would I get the testingId there?

vasireddy99: Suggested those changes

terraform/templates/defaults/ecs_taskdef.tpl
```diff
@@ -42,16 +42,28 @@
...
-        "name": "ZIPKIN_RECEIVER_ENDPOINT",
-        "value": "127.0.0.1:${http_port}"
+            "name": "ZIPKIN_RECEIVER_ENDPOINT",
+            "value": "127.0.0.1:${http_port}"
         },
         {
-        "name": "OTEL_METRICS_EXPORTER",
-        "value": "otlp"
+            "name": "OTEL_METRICS_EXPORTER",
+            "value": "none"
```
bryan-aguilar: Why have you disabled the metrics exporter?

terraform/validation/main.tf
```diff
@@ -73,7 +73,7 @@ resource "null_resource" "validator" {
   provisioner "local-exec" {
     command = <<-EOT
       docker-compose -f ${local.docker_compose_path} down
-      docker-compose -f ${local.docker_compose_path} build
+      docker-compose -f ${local.docker_compose_path} build --no-cache
```
bryan-aguilar:
```suggestion
      docker-compose -f ${local.docker_compose_path} build
```

This was only needed for local builds. We can use the cache normally.

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
```diff
@@ -0,0 +1,114 @@
...
+
+  private static final String LOGGROUPPATH = "/aws/ecs/otlp/%s/logs";
```
bryan-aguilar, line 31: nit: This method of building the log group name introduces a strong coupling between the terraform test and the validator. It requires the config to use this exact log group name format. Instead it could be better if the validator is provided the log group and stream name at launch.

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
```diff
@@ -0,0 +1,114 @@
...
+    MustacheHelper mustacheHelper = new MustacheHelper();
+    String templateInput = mustacheHelper.render(expectedDataTemplate, context);
+    JsonNode jsonNode = JsonLoader.fromString(templateInput);
+    JsonSchemaFactory jsonSchemaFactory =
+        JsonSchemaFactory.newBuilder()
+            .setReportProvider(new ListReportProvider(LogLevel.INFO, LogLevel.FATAL))
+            .freeze();
+    JsonSchema jsonSchema = jsonSchemaFactory.getJsonSchema(jsonNode);
+    this.schema = jsonSchema;
+    caller.callSampleApp();
```
bryan-aguilar: Why do we call the sample app during init? Should we be calling the sample app each retry attempt instead?

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
```diff
@@ -0,0 +1,114 @@
...
+  protected void fetchAndValidateLogs(Instant startTime) throws Exception {
+    List<OutputLogEvent> logEvents =
+        cloudWatchService.getLogs(
+            logGroupName, logStreamName, startTime.toEpochMilli(), QUERY_LIMIT);
+    if (logEvents.isEmpty()) {
+      throw new BaseException(
+          ExceptionCode.LOG_FORMAT_NOT_MATCHED,
+          String.format(
+              "[StructuredLogValidator] no logs found under log stream %s" + " in log group %s",
+              logStreamName, logGroupName));
```
bryan-aguilar, line 97: Don't use `String` class for logging. Use methods such a `log.info()` or `log.error()`.

validator/src/main/resources/expected-data-template/otlpExpectedLog.mustache
```diff
@@ -0,0 +1,10 @@
+{
+    "body": "Executing outgoing-http-call",
+    "severity_number": "{{severity_number}}",
+    "severity_text": "INFO",
+    "trace_id": "{{trace_id}}",
+    "span_id": "{{span_id}}",
```
bryan-aguilar: Where do these values get populated from? Are they present in the context?

terraform/templates/defaults/ecs_taskdef.tpl
Danielzolty: I was getting a log error stating that I should add this. I'll try rerunning without and see what happens.

bryan-aguilar: This would probably break other test cases that may require metrics to be emitted. I would not over analyze the logs in the sample app.

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
Danielzolty, line 31: As in I should add it as an environment variable or have it as a variable established at runtime? Also, looking at other validators, it seems they establish the log group name in a similar fashion. For example for ContainerInsightECSPrometheus Validator, it is established by
`logGroupName =
        String.format(
            "/aws/ecs/containerinsights/%s/%s",
            context.getCloudWatchContext().getClusterName(), "prometheus");`
Does this not have a coupling issue between the tf test and validator?

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
Danielzolty: Looking at other validators, it seems it is being called in the validate() method, but not in the retry helper(also, I don't think it should be in a retry because calling the sample app uses retry helper on its own right, which would be redundant). I'll move the callSampleApp() to validate() method.

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
Danielzolty: Just to clarify, you are saying to use log.error() in place of throwing exception?

validator/src/main/resources/expected-data-template/otlpExpectedLog.mustache
Danielzolty: I believe they are being populated from the actual log message. The goal wasn't to test that the values matched up, rather, to confirm that there were these fields present in the actual log message.

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
bryan-aguilar: Yes, there is a coupling there also. You don't need to address this comment now.

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
bryan-aguilar: Ahhh, ignore this comment. I missed that this was being thrown in an exception. my apologies.

validator/src/main/java/com/amazon/aoc/validators/CWLogValidator.java
```diff
@@ -0,0 +1,114 @@
...
+  protected void validateJsonSchema(String logEventMsg) throws Exception {
+    JsonNode logEventNode = mapper.readTree(logEventMsg);
+    if (schema != null) {
+      ProcessingReport report = schema.validate(JsonLoader.fromString(logEventNode.toString()));
+      if (report.isSuccess()) {
+        log.info("Report was a success");
+      } else {
+        log.info("[StructuredLogValidator] failed to validate schema \n");
+        log.info(report.toString() + "\n");
+      }
```
bryan-aguilar: How will the test case know to fail if it reaches this logical block? Nothing is returned back to the calling function to indicate a failure.

It would be much better if you added some unit tests for this new validator class where we could mock out some scenarios.

terraform/templates/defaults/ecs_taskdef.tpl
Danielzolty: Tests passed, but got this error message in cloudwatch:
`[otel.javaagent 2023-10-31 16:59:10:056 +0000] [OkHttp http://127.0.0.1:4317/...] ERROR io.opentelemetry.exporter.internal.grpc.GrpcExporter - Failed to export metrics. Server responded with UNIMPLEMENTED. This usually means that your collector is not configured with an otlp receiver in the "pipelines" section of the configuration. If export is not desired and you are using OpenTelemetry autoconfiguration or the javaagent, disable export by setting OTEL_METRICS_EXPORTER=none. Full error message: unknown service opentelemetry.proto.collector.metrics.v1.MetricsService`
So, what do you think? Keep it or get rid of it?

bryan-aguilar: Just leave it. You are adjusting the default template for all ecs test cases. Other test cases may require that, there is a large blast radius in that template.

bryan-aguilar: That is an error from the sample app because the collector does not start an OTLP metrics receiver in your test case

validator/src/test/java/com/amazon/aoc/validators/CWLogValidatorTest.java
```diff
@@ -0,0 +1,145 @@
...
+            + "    \"resource\": {\n"
+            + "        \"service.name\": \"aws-otel-integ-test\"}}";
+
+    OutputLogEvent outputLogEvent = new OutputLogEvent();
+    outputLogEvent.setMessage(message);
+    List<OutputLogEvent> eventList = new ArrayList<>();
+    eventList.add(outputLogEvent);
+    CWLogValidator validator = runValidation(validationConfig, context, eventList);
+    ProcessingReport report = validator.getProcessingReport();
+    assertFalse(report.isSuccess());
```
bryan-aguilar: You don't need this. Your entire validator factory should be tested when you call `runValidation()`. 

See [here](https://github.com/aws-observability/aws-otel-test-framework/blob/efb8515bd27d154787067c8a016fd8ddafbdfd7d/validator/src/test/java/com/amazon/aoc/validators/CWMetricValidatorTest.java#L59) on how the unit test should behave on good data..aka the validation passes. 
See [here](https://github.com/aws-observability/aws-otel-test-framework/blob/efb8515bd27d154787067c8a016fd8ddafbdfd7d/validator/src/test/java/com/amazon/aoc/validators/CWMetricValidatorTest.java#L121) on how the unit test should behavior when it handles bad data...aka the validation fails.

validator/src/test/java/com/amazon/aoc/validators/CWLogValidatorTest.java
```diff
@@ -0,0 +1,151 @@
...
+    outputLogEvent.setMessage(message);
+    List<OutputLogEvent> eventList = new ArrayList<>();
+    eventList.add(outputLogEvent);
+    CWLogValidator validator = runValidation(validationConfig, context, eventList);
+    ProcessingReport report = validator.getProcessingReport();
+    assertTrue(report.isSuccess());
+  }
+
+  @Test
+  public void testFailedLogMessage() throws Exception {
```
bryan-aguilar: Can we use a more descriptive test case name? Why do we expect this log to fail?

validator/src/test/java/com/amazon/aoc/validators/CWLogValidatorTest.java
```diff
@@ -0,0 +1,151 @@
...
+            + "    \"resource\": {\n"
+            + "        \"service.name\": \"aws-otel-integ-test\"}}";
+
+    OutputLogEvent outputLogEvent = new OutputLogEvent();
+    outputLogEvent.setMessage(message);
+    List<OutputLogEvent> eventList = new ArrayList<>();
+    eventList.add(outputLogEvent);
+    CWLogValidator validator = runValidation(validationConfig, context, eventList);
+    ProcessingReport report = validator.getProcessingReport();
+    assertTrue(report.isSuccess());
```
bryan-aguilar:
```suggestion
```
These assertions should not be required. No exception being thrown indicates the test case is passing right?

validator/src/test/java/com/amazon/aoc/validators/CWLogValidatorTest.java
bryan-aguilar: How long does this test case take to run? Does it continue to cycle through all of it's retry attempts before it fails?  If it does, we need a way to initialize a CW Log validator where we can pass in the retry attempts and/or duration of retry wait time. 

Basically, these unit tests should run fast.

validator/src/main/java/com/amazon/aoc/exception/ExceptionCode.java
```diff
@@ -37,6 +37,7 @@ public enum ExceptionCode {
   NOT_ENOUGH_SPANS(50010, "not enough spans in the trace"),
   COLLECTOR_ID_NOT_MATCHED(50011, "span collector-id attribute does not match"),
   NULL_VAR(50012, "variable is null"),
+  EXPECTED_LOG_NOT_FOUND(50013, "Expected Log Not Found"),
```
bryan-aguilar:
```suggestion
  EXPECTED_LOG_NOT_FOUND(50013, "expected log not found"),
```

validator/src/test/java/com/amazon/aoc/validators/CWLogValidatorTest.java
Danielzolty: No, the retry number is set to 1, so it takes all the tests milliseconds to run

bryan-aguilar: Ahh, I see. 👍

bryan-aguilar (approved):

PR #1497, opened 2023-11-15
Adding aws_sdk_2 trace prefix

PaurushGarg:

ADOT Java Agent now prefixes like [this](https://github.com/aws-observability/aws-apm-java-instrumentation/blob/2f4cfd531449a59cd256028b7d5bb5dcec1ae93a/awsagentprovider/src/main/java/software/amazon/opentelemetry/javaagent/providers/AwsMetricAttributeGenerator.java#L197) for aws-sdk-2 scoped traces.
which is causing [this](https://github.com/aws-observability/aws-otel-java-instrumentation/actions/runs/6873043238/job/18692875229) failure.
springboot-otel-trace-metric-validation.yml is not used anywhere else other than for testing springboot app from java agent repo, which is using aws_sdk_2. Made changes accordingly.


vasireddy99:
Do we have to change the usage in this [template](https://github.com/aws-observability/aws-otel-test-framework/blob/f7aa645faac6e0dd3eb5850aa13e804a1c4451a5/validator/src/main/resources/validations/springboot-otel-trace-validation.yml#L12) as well. If so, is it affecting other tests as well

PR #1627, opened 2024-05-10
Set collector image and tag in helm chart

bryan-aguilar:
**Description:** Operator helm chart now requires collectorImage.repository to be set. This PR ties it to the `aoc_image_repo` and `aoc_version` that is widely used in terraform framework

https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1154

### DIFF
--- a/terraform/common.tf
+++ b/terraform/common.tf
@@ -31,7 +31,7 @@ variable "sample_app_image" {
 }
 
 variable "aoc_image_repo" {
-  default = "public.ecr.aws/aws-observability/aws-otel-collector"
+  default = "public.ecr.aws/aws-otel-test/adot-collector-integration-test"
 }
 
 variable "aoc_version" {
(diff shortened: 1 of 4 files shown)


terraform/common.tf
```diff
@@ -31,7 +31,7 @@ variable "sample_app_image" {
 }
 
 variable "aoc_image_repo" {
-  default = "public.ecr.aws/aws-otel-test/adot-collector-integration-test"
+  default = "public.ecr.aws/aws-observability/aws-otel-collector"
```
vasireddy99: Can we keep the image to - public.ecr.aws/aws-otel-test/adot-collector-integration-test integ test image itself as in by default to pick the latest test image. Not a strong reason though, just a preference

bryan-aguilar: done

Aneurysm9 (approved):

PR #1645, opened 2024-06-14
Allow test cases to specify arguments for the collector

Aneurysm9:
This is used to disable feature gates that effect the removal of deprecated components, allowing them to continue to be tested until they are fully removed.

### DIFF
--- a/terraform/testcases/datadog_exporter_metric_mock/parameters.tfvars
+++ b/terraform/testcases/datadog_exporter_metric_mock/parameters.tfvars
@@ -4,3 +4,5 @@ soaking_data_mode = "metric"
 sample_app = "spark"
 
 sample_app_image = "public.ecr.aws/aws-otel-test/aws-otel-java-spark:latest"
+
+otconfig_args = ["--feature-gates=-adot.exporter.datadogexporter.deprecation"]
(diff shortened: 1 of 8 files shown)


HudsonHumphries (approved):

terraform/testcases/datadog_exporter_metric_mock/parameters.tfvars
```diff
@@ -4,3 +4,5 @@ soaking_data_mode = "metric"
 sample_app = "spark"
 
 sample_app_image = "public.ecr.aws/aws-otel-test/aws-otel-java-spark:latest"
+
+otconfig_args = ["--feature-gates=-adot.exporter.datadogexporter.deprecation"]
```
vasireddy99, line 8: We might not need the flag passed for datadog, as we have the datadag exporter in stagealpha and enabled by default.

Aneurysm9: Yes, but doing it now ensures that when the flag changes there won't be a chance it is forgotten here.

vasireddy99: yeah Agreed, makes sense

PR #1829, opened 2026-05-06
fix: use ecs detector instead of eks for Fargate resourcedetection

mitali-salvi:
The `eks` detector in the `resourcedetection` processor fatally crashes on EKS Fargate in OTel Collector Contrib v0.143.0+ because IMDS is unavailable and `K8S_NODE_NAME` is not set. The collector never starts, producing zero metrics and zero logs — causing both `eks_containerinsights_fargate_docker` and `eks_containerinsights_fargate_docker_metric` tests to fail.

The only resource attribute needed by this pipeline is `ClusterName`, which is already provided by the `env` detector via `OTEL_RESOURCE_ATTRIBUTES` set in the StatefulSet.

Fix: `detectors: [env, eks]` → `detectors: [env]`

Also reverts incorrect Pod.json schema and stateful_set feature gate changes from PR #1827.

Ref: https://github.com/aws-observability/aws-otel-collector/issues/3161

### DIFF
--- a/terraform/eks/container-insights-agent/config_map_fargate.yml
+++ b/terraform/eks/container-insights-agent/config_map_fargate.yml
@@ -402,9 +402,11 @@ data:
             action: update
             new_name: container_$$1
 
-      # add cluster name from env variable and EKS metadata
+      # add cluster name from env variable (OTEL_RESOURCE_ATTRIBUTES)
+      # Note: the eks detector fatally fails on Fargate because IMDS
+      # is unavailable and K8S_NODE_NAME is not set.
       resourcedetection:
-        detectors: [env, eks]
+        detectors: [env]
 
       batch:
         timeout: 15s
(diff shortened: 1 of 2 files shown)


terraform/eks/container-insights-agent/config_map_fargate.yml
```diff
@@ -402,9 +402,11 @@ data:
             action: update
             new_name: container_$$1
 
-      # add cluster name from env variable and EKS metadata
+      # add cluster name from env variable (OTEL_RESOURCE_ATTRIBUTES)
+      # Note: the eks detector fatally fails on Fargate because IMDS
+      # is unavailable and K8S_NODE_NAME is not set.
```
jefchien, line 407: We don't need to verify that the rest of the resource attributes work on EKS Fargate?

jefchien (approved):

